Issue #3, opened 2023-03-21
github workflows to run tests


guynir42:
Should probably also add a demo test for now.

PR #5, opened 2023-03-21
Automated Workflows

guynir42:
Added a requirements file, a github workflow that runs automated tests (including a postgres database) and added a demo test.


guynir42:
addresses #3

PR #10, opened 2023-03-22
Pulling in my config system

rknop:
The config system I've used in lensgrinder and curveball.  Allows us to config everything (e.g. database locations, file storage locations, etc.) with yaml files.  If you look at what I did in lensgrinder, I make extensive use of the includes so that stuff that's the same for different instances (e.g. remote services) go in one file.


tests/test_config.py
```diff
@@ -0,0 +1,141 @@
...
+        assert cfg.value('mainlist1') == ['main1', 'main2', 'main3', 'aug1', 'aug2']
+        assert cfg.value('mainlist2') == ['override1', 'override2']
+
+        assert cfg.value('mainscalar1') == 'aug1'
+        assert cfg.value('mainscalar2') == 'override2'
+        assert cfg.value('mainscalar3') == 'override2'
+
+        assert cfg.value('mainnull') is None
+        with pytest.raises( ValueError ):
+            print( f"config value notdefined is {cfg.value('notdefined')}" )
```
guynir42, line 64:
```suggestion
        with pytest.raises( ValueError ) as e:
            cfg.value('notdefined')
        assert 'Error getting field' in str(e)
```

guynir42:
```suggestion
        with pytest.raises( ValueError, match='Error getting field' ):
            cfg.value('notdefined')
```

guynir42: The second one is even better :)

tests/test_config.py
```diff
@@ -0,0 +1,141 @@
...
+        with pytest.raises( TypeError ):
+            cfg.set_value( 'settest.list.notanumber', 'kitten', appendlists=True )
+        with pytest.raises( TypeError ):
+            cfg.set_value( 'settest.0', 'puppy' )
+        with pytest.raises( TypeError ):
+            cfg.set_value( 'settest.0.subset', 'bunny' )
+        with pytest.raises( TypeError ):
+            cfg.set_value( 'settest.dict.0', 'iguana' )
+        with pytest.raises( TypeError ):
+            cfg.set_value( 'settest.dict.2.something', 'tarantula' )
```
guynir42, line 101: Maybe add a `match=...` clause to all of the `raises` calls?

util/config.py
```diff
@@ -0,0 +1,432 @@
...
+                filename = f"{newdir}{filename[len(olddir):]}"
+                break
+        return filename
+
+    @staticmethod
+    def clone( configfile=None, reread=False, logger=logging.getLogger("main"), dirmap={} ):
+        """Returns a config object.
+
+        Will call "get" on the passed configfile, but will *not* return
+        the singleton object for that config file.  RAther, will make a
```
guynir42, line 162:
```suggestion
        the singleton object for that config file.  Rather, will make a
```

util/config.py
```diff
@@ -0,0 +1,432 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
```
guynir42, line 3: Do we need these?

util/config.py
```diff
@@ -0,0 +1,432 @@
...
+    _default_default = os.getenv("SEECHANGE_CONFIG", None)
+
+    _default = None
+    _configs = {}
+
+    @staticmethod
+    def init( configfile=None, logger=logging.getLogger("main"), dirmap={} ):
+        """Initialize configuration globally for process."""
+
+        Config.get( configfile, logger=logger, dirmap=dirmap )
```
guynir42, line 103: Will this call set the default? It looks like it doesn't...

util/config.py
```diff
@@ -0,0 +1,432 @@
...
+        default config object.  If there is no default config object
+        (because neither Config.get() nor Config.init() have been called
+        previously), an exception will be raised.
+
+        """
+        if configfile is None:
+            if Config._default is None:
+                if Config._default_default is None:
+                    raise RuntimeError( f'No default config defined yet; run Config.init(configfile)' )
+                Config._default = Config._default_default
```
guynir42, line 134: What does `_default_default` actually has?

util/config.py
```diff
@@ -0,0 +1,432 @@
...
+        curfield = fields[0]
+        try:
+            ifield = int(curfield)
+        except ValueError:
+            ifield = None
+        return fields, isleaf, curfield, ifield
+
+    @staticmethod
+    def merge_trees( left, right, augment=False ):
+        """Internal usage, do not call."""
```
guynir42, line 407: I have come to appreciate that even internal functions should have a proper docstring. Down the line it could help make things clearer when trying to debug something. 

I would prefer if we use "numpy" style for functions:
```
"""
Function description...

Parameters
--------------
par1: type
  description
par2: type
  description

Returns
----------
return1: type
  description
```

util/config.py
```diff
@@ -0,0 +1,432 @@
...
+            return struct
+
+    def set_value( self, field, value, structpass=None, appendlists=False ):
+        """Set a value in the singleton for the current session.
+
+        Does not save to disk.  Follows the standard rules docuemnted in
+        "augment" and "override"; if appendlists is True, uses
+        "augment", else "override".  Will create the whole hierarchy if
+        necessary.
+
```
guynir42, line 323: Description for the other input parameters could be useful here.

util/config.py
```diff
@@ -0,0 +1,432 @@
...
+            if Config._default is None:
+                Config._default = configfile
+
+        if setdefault:
+            Config._default = configfile
+                
+        return Config._configs[configfile]
+
+    @staticmethod
+    def dirmap( filename, dirmap ):
```
guynir42, line 150: Maybe explain how to use this in a docstring?

guynir42 (commented):
I have some comments but the code in general looks good. I think this would be really useful.

tests/test.yaml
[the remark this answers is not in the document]
guynir42: I would also split up the tests themselves into subfolders. Right now I think `models`, `pipeline` and `utils` may be a good start.

tests/test_config.py
```diff
@@ -0,0 +1,141 @@
...
+if not str(_rundir.parent) in sys.path:
+    sys.path.insert(0, str(_rundir.parent ) )
+from util import config
+
+# A note about pytest: Things aren't completely sandboxed.  When I call
+# config.Config.get(), it selts Config._default, and that carries over
+# from one test to the next even if the call wasn't in a fixture with
+# class scope.  (The tests below are ordered with this in mind.)
+
+class TestConfig:
```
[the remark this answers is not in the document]
guynir42: You can see an example in my open PR.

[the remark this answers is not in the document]
rknop: I have to admit that I prefer grouping together all of the tests for a class in a single test class.

But, more importantly : does it really matter?  Within one test file, yeah, we should have consistency for readability, but I don't think they all have to adhere to exactly the same way of doing it as long as they're all readable.  One thing I'd like to avoid in these reviews is reformatting for the sake of reformatting.  We aren't going to have the same style on everything, and that's OK, as long as we aren't doing things that are going to make other things harder later.

tests/test_config.py
```diff
@@ -0,0 +1,141 @@
...
+    def test_no_default( self ):
+        assert config.Config._default == None
+
+    @pytest.mark.skipif( os.getenv("SEECHANGE_CONFIG") is None, reason="Set SEECHANGE_CONFIG to test this" )
+    def test_default_default( self ):
+        cfg = config.Config.get()
+        assert cfg._path == pathlib.Path( os.getenv("CURVEBALL_CONFIG") )
```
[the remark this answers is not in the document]
rknop: Yes... is there a way I can edit it right here while reviewing?

tests/test_config.py
```diff
@@ -0,0 +1,141 @@
+import sys
+import os
+import pathlib
+import pytest
+
+_rundir = pathlib.Path(__file__).parent
+if not str(_rundir.parent) in sys.path:
+    sys.path.insert(0, str(_rundir.parent ) )
```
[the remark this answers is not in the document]
rknop: Yeah, we should do it that way.  We should fidget enough once to make it work, and then hopefully we won't have to fidget any more.

util/config.py
rknop: Yeah, I agree that internal functions should have a docstring.  I was just lazy.  (But, I also think it's worth keeping the "internal function" warning in the docstring.  Occasionally it's an advantage of Python that it doesn't have private methods like C++ or Java, but there is value in posting a warning that you shouldn't use something externally.)

tests/test_config.py
```diff
@@ -0,0 +1,141 @@
+import sys
+import os
+import pathlib
+import pytest
+
+_rundir = pathlib.Path(__file__).parent
```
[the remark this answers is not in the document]
rknop: Yeah, that's a good idea.  That ought to be in config rather than db, though, but wherever it is, there's going to be a chicken and egg problem, though your next comment is really the right way to do it.

tests/test.yaml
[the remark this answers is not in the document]
rknop: I would agree with this, but this comment is marked "outdated"  -- do you know why?

tests/test_config.py
```diff
@@ -0,0 +1,141 @@
...
+        cfg = config.Config.get()
+        assert cfg._path == pathlib.Path( os.getenv("CURVEBALL_CONFIG") )
+
+    def test_set_default( self ):
+        cfg = config.Config.get( _rundir / 'test.yaml', setdefault=True )
+        assert config.Config._default == f'{ (_rundir / "test.yaml").resolve() }'
+
+    @pytest.fixture(scope='class')
+    def cfg( self ):
+        return config.Config.get( _rundir / 'test.yaml', setdefault=True )
```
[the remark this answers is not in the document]
rknop: This may be part of a larger style conversation, but:

I don't think it's worth having a whole separate file for three lines of code.  I could see the argument that the fixtures should be at the top.  And, if there are several fixtures, then, yes, I can see the reason to have them in a separate file.  But a quick thing that a file needs that is only three lines long, doing it for the sake of consistency isn't really worth it, and if anything chops the code up enough to make it harder to read later.   I don't see this fixture being used anywhere else (since the test yaml file is something designed just for this test, and won't be useful elsewhere).

Re: the larger conversation, I am going to object strenuously to rigid style rules.  To my mind, the very most important part of PEP-8 is the first section after the Introduction, and I think the writers of PEP-8 were smart in making that the first thing.  (There are some other parts of PEP-8 I would diagree with, but I think it's more important in that case just not to bother getting into arguments about it.)  I know some people think you should use tools like "black" on your codebase, but I hate that.  (Black in particular destroys so much readability of the code that it hinders maintainability.   As I inherited Lensgrinder from Danny, at one point early on he ran black on the entire codebase.  I spent a lot of time thereafter removing the ugliness that that one change had imposed.)

tests/test_config.py
rknop: I guess I don't understand what you're thinking here.

util/config.py
rknop: No.  Vestigal from stuff I used to always do that I eventually decided I didn't need to do.

tests/test_config.py
rknop: OH.

I was being stupid.  I thought you  meant the python match statement; I see what you mean now.

tests/test_config.py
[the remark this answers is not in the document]
guynir42: Functionally it is the same. But I think it looks a lot more professional if we decide on one standard and conform to that. 
As for this test class thing, I don't see why the extra code and additional indentation level is needed. If you feel strongly about it or if you think it has some added value, I'd just change my tests to the same format as well.

tests/test_config.py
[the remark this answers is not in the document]
guynir42: When there's a suggestion you can just accept it as another commit right from the web-page. In this case I am not sure if that's possible. 
But I usually make all the requested changes locally and push another commit.

tests/test_config.py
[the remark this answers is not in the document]
guynir42: This test doesn't have a lot of (reusable) fixtures, but down the line we will have many. 
The question is this: do we want to keep the "relevant for this test only" fixtures in the same file as the test, and only have the "shared" fixtures in `conftest.py` or should we just put all of them in one place so it is less work to find them? 

I don't mind either way, but we will certainly have a `conftest.py` file.

util/config.py
guynir42: Yeah, another option is to change the function name to start with an underscore.

util/config.py
guynir42: I just can't find the place where `_default_default` actually gets set.

rknop:
So, I screwed up, and committed to the main branch of my fork, which automatically  merged all *those* changes into this PR, even though it's a different set of things.  Going to change how I manage branches on my fork, but for now, it is what it is.

The new changes move the config tests to a "util" subdirectory.  More significantly, it adds a docker-compose.yaml for running the tests, and dockerfiles for a q3c-enabled postgres and for running the application.  (Right now, yes, a list of pip packages to install accomplishes what we need, but that won't be true forever.  Also, since production will be running in the dockerfile, it  makes sense to test it there.)  The docker building in the tests don't work quite the way I want (it won't pull from a docker registry, so it rebuilds the docker file every time, which is gratuitous), but it works.

tests/util/test.yaml
guynir42, line 1: Why not `tests/data`? We will need to have a folder like that sooner or later? And a config file feels more like "data" than "utils" (which we could have in there as well).

tests/util/testpreload2.yaml
guynir42, line 1: I would even put all of these configs inside `tests/data/configs` or something like that.

guynir42 (commented):
I have mostly small comments.

guynir42 (approved):

PR #16, opened 2023-03-27
Parameters object for pipeline classes

guynir42:
Key features:

When adding a parameter using `add_par` it is given allowed types, a description, a default value, and a flag if it is critical.
This lets you "lock" a Parameters object such that it won't accept wrong parameter names or types.
This lets you easily print out the state of an object and the description for each parameter.
All the parameters of a pipeline object are stored inside `.pars` instead of being mixed in with the data attributes.
You can ask the Parameters object for a dictionary of the parameters and their current values.
Initializing the Parameters object can be done by giving `kwargs` to the constructor.

Ideally each pipeline object will have a sub-class of Parameters, with a set of specific parameter names.

Can specify a list of parameters that are shared by all sub classes and copied from one Parameters object to the next. Example: the verbose flag can be passed down from a top-level object's Parameters to all the child object's Parameters.

Example: you have a SubtractionPipeline object that gets images and makes subtractions. It probably has a bunch of data attributes for holding on to the images and PSFs. But it would also have a `.pars` attribute which is a SubtractionPars object that inherits from Parameters. Then you define some specific parameters like `subtraction_type` on it, which means you can't accidentally use `subtraction_tipe` and then assign to an attribute that is misspelled. You'd also have `verbose` inherited from the superclass.
and when it is time to get the parameter dictionary that is saved in the Provenance row, you get it by asking for the critical parameters as a dict.

The down side to all of this is that it requires a little bit more typing when setting/getting parameters and it forces you to define all the parameters inside a helper class, instead of in the pipeline object's __init__(). But the separation between parameters and data is really useful. Also the ability to keep track of "critical" parameters would end up helping a lot.

### DIFF
--- a/pipeline/parameters.py
+++ b/pipeline/parameters.py
@@ -1,0 +1,652 @@
+import copy
+
+
+# parameters that are propagated from one Parameters object
+# to the next when adding embedded objects.
+# If the embedded object's Parameters doesn't have any of these
+# then that key is just skipped
+
+class Parameters:
+    """
+    Keep track of parameters for any of the pipeline classes.
+    You can access the parameters as attributes, but also
+    as dictionary items (using "in" and "pars[key]").
+
+    Methods
+    -------
+    - add_par() to add a new parameter (mostly in the __init__).
+    - add_alias() to add an alias for a parameter.
+    - overwrite() the parameters from a dictionary.
+    - augment() takes parameters from a dictionary and updates (instead of overriding dict/set parameters).
+    - to_dict() converts the non-private parameters to a dictionary.
+    - copy() makes a deep copy of this object.
+    - compare() checks if the parameters of two objects are the same.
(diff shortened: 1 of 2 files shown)


guynir42:
The idea is to have each of the pipeline objects ingest one part of the config. 
Example: you have a "Subtraction" object that does all the references management and subtraction methods. 
You make a ParsSubtraction object for it, so sub.pars contains all the tunable parameters. 
Then instead of doing config['subtraction.some_parameter'] you'd do self.pars.some_parameter. 
This seems superficially the same, but you'd get:
- hard coded defaults in case you don't want them in the config. 
- verification of the name and type of the parameter.
- quick access using shorthands, etc. 
- getting all the critical parameters into a dictionary, including changes the user made in an interactive session. 
- printout all available parameters and their description interactively. 
- separation of global config into local scope of each specific class.

rknop (approved):

PR #19, opened 2023-03-29
Pipeline top-level design sketch

guynir42:
This is really preliminary, and more of a starting point for a discussion. 

The requirements for this, in my mind, were:
- Must work from start to finish on a single exposure ID and CCD ID, and not have to reload things that were already calculated. 
- Must work in smaller pieces: e.g., you should be able to run image subtraction on its own, assuming the needed data is on the DB. 
- Keep database transactions short, and allow a `with` statement for each creation of a session. 
- Make the signature for the `run()` function the same for all classes. 
- Make sure the provenance is added in the correct way without user intervention (the user must make sure that CodeVersions are correct!). 

The models are just empty right now, but we can fill them out in separate PRs, along with the actual code that does things. 

Let me know what you think!


models/provenance.py
```diff
@@ -6,6 +6,8 @@
 from sqlalchemy.orm import relationship
 from sqlalchemy.dialects.postgresql import JSONB
 
+from pipeline.utils import get_git_hash
```
rknop, line 9: This is causing an error in the tests ; looks like a circular import error.  We should fix that before merging ths.

guynir42: I was afraid I would have circular imports. Let me take care of that first.

pipeline/utils.py
```diff
@@ -0,0 +1,56 @@
+import git
+from collections import defaultdict
+
+import sqlalchemy as sa
+
+
+
+def get_image_cache(image_cache=None):
+    if image_cache is None:
+        image_cache = defaultdict(dict)
```
rknop, line 10: What's the benefit of using collections.defaultdict instead of just {} here?

But, also, I think these get_*_cache objects should be combined together into a Context class, as I've mentioned above.

pipeline/astrometry.py
```diff
@@ -0,0 +1,121 @@
...
+        self.pars = ParsAstrometry()
+        self.pars.update(kwargs)
+
+    def run(self, exposure_id, ccd_id, image_cache=None, prov_cache=None, session=None):
+        """
+        Extract sources and use their positions to calculate the astrometric solution.
+
+        Parameters
+        ----------
+        exposure_id : int
```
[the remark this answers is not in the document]
guynir42: I really would like to keep the exposure_id/ccd_id throughout the whole pipeline. The same exp/ccd combination does not uniquely identify which image is picked. This is true for astrometry and for all other pipeline stages. 
The way I think about it is that if you must specify which provenance you are using in the upstream, and only then you can pinpoint the right image from the exp/ccd ids. This can happen in one of the following ways:
1) you gave the image in the image cache, so it is already loaded. 
2) you gave the provenance in the provenance cache, so it can be used to load a specific image (this is the least likely to happen, practically). 
3) you just load the "most recent" provenance and that identifies which image is associated with exp/ccd ids in the database.

pipeline/astrometry.py
```diff
@@ -0,0 +1,121 @@
...
+            The CCD number in the camera's numbering scheme.
+        image_cache : dict
+            A dictionary of images that have already been loaded.
+            Includes a dictionary of 'exposures' keyed by exposure_id.
+            Also includes dictionaries for 'images', 'subtractions',
+            'references' and 'cutouts'.
+            Each one is keyed on a tuple of (exposure_id, ccd_id).
+            If the required image is not found in the cache, it will be loaded
+            from the database and then from disk.
+        prov_cache: dict
```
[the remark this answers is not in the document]
guynir42: That may be a good idea. Let me think about that.

pipeline/subtractor.py
```diff
@@ -0,0 +1,122 @@
...
+    def _get_upstream_process_names(self):
+        return ['preprocess', 'astrometry', 'calibration']
+
+
+class Subtractor:
+    def __init__(self, **kwargs):
+        self.pars = ParsSubtractor()
+        self.pars.update(kwargs)
+
+        # TODO: add a reference cache here.
```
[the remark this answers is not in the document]
guynir42: I thought about this, but I would like to separate the constructor from the data flow. 
The constructor is where you set up the `Parameters` object. Then when the pipeline is ready, you give it data. 
This lets you loop on an object without having to re-construct it every time (which could be expensive if the object also caches stuff like reference images). 

I think the choice of reference image should probably be done using provenance choices, but I can also see someone just making a reference locally and passing it in through the `image_cache` (or `context`) to use that explicitly. 
In either case you must have a provenance to the reference so it gets saved in the subtracted image's provenance's upstream.

pipeline/utils.py
guynir42: The default dict will never say "this key does not exist" so you can just query it using `get` and if it is not there it will be initialized as an empty dict. It's just convenient. 
We can probably do `context` instead.

pipeline/astrometry.py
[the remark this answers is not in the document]
rknop: But how are you going to handle images that don't have an associated exposure?  For instance, references download from a survey, or stacks?

pipeline/subtractor.py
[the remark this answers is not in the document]
rknop: Why would you have to reconstruct it every time if you passed the context in with the constructor?  Would you want to wipe out the image cache between subsequent calls?

For that matter -- why are we calling it more than once in a go through the pipeline?

pipeline/astrometry.py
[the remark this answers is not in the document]
guynir42: I think that would be handled by the same object but using a different function, or maybe just add a `coadd_id` that when you use it, it overrides the exposure_id/ccd_id. 
The reason I went with the exp/ccd ids is that it lets the user easily pick what data to process, without needing to know the ids of the intermediate products.

pipeline/subtractor.py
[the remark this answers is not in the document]
guynir42: I was thinking something like:
```
pipeline = Pipeline(arguments, ...)

for i in range(...):
  for j in range(...):
    pipeline.run(i, j)
```

So that the same object can be re-used in a loop. If you want, you can initialize it once per image, but you don't have to.

pipeline/astrometry.py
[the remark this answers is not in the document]
rknop: I agree that you want the user to be able to pass exposure name and chip number.  But, that can be handled by the code without the architecture of calling every function being tied to exposure name and chip number.

As I understand it, the real goal is to be able to track when the same image starts the pipeline a second time.  This means that a given exposure and CCD will have to have a unique signature, but it doesn't *have* to be (exposure id, ccd id).  We could define an "ImageSource" class or some such where each object has a unique id.  Then, we write as many methods as we need for that ImageSource based on where images come from -- it might be exposure plus ccd, it might be a downloaded survey, it might be a coadd with a specific coadd provnance signature, etc.   Indeed, there can be multiple subclasses of ImageSource to cover the various possibilities (though there will need to be a single method that knows all of them, so other code has a single place to call and doesn't have to try all of them).  All of those methods take what's needed, and return the unique id.  That is the thing that the rest of the methods sling around, and they can go back to that class to get more information as necessary.

When the code is run, the user can specify exposure name and chip, and the code calls the right method in this class to get the unique id.  Later, part of the code identifies wants a reference from a downloaded survey, so it calls whatever code identifies that, and then it can use *that* unique id to pass to astrometry, photometry, etc.

This is more flexible, because it doesn't bake in a specific way of identifying an image to every other routine we call later.  It's also cleaner, because most of the subsequent methods probably don't care about exposure and chip, they just need a way of identifying the image.  A unique handle for that is going to be a better way to do it than the specific data for one type (albeit the most common) of way of originating an image.

pipeline/subtractor.py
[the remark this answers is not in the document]
rknop: I would make it the standard to reinitialize once per image.  Reason: you're talking about caching all of the images in memory.  If you keep that cache around, it's going to keep growing every time you start a new image.  To avoid that becoming a problem, you're going to have to write cache management code, and that's not worth the effort.  It's much easier to just say when you initialize a new subtractor, you specify exactly what image it's for.

The overhead in creating these things, and reading these files from disk, is going to be utterly tiny compared to the time it takes to actually process them.  (The overhead of getting them over the network may well not be, but that's why we have local storage.)

Right now, the way I run the code, I start a whole new executable every time I'm starting a new image.  Again, the run time is minutes, so a few seconds of startup time is not a significant hit.

It will simplify how we deal with things if we specify the context / image etc. in the constructor, and not have to worry about cache bloat and such.

guynir42:
@rknop this is still not ready to merge, but the overall design, I think, is clear. 
There's a DataStore object that is used to cache data products that have already been produced. It is also smart enough to fetch the products from DB if needed. 
Also, you can initialize it with exposure/CCD ids or with an Image id. 
The models are all empty, and the actual code for doing the processing is not here at all, but the data flow is more or less done. 
Also, I am still thinking on how to get the references. 
Let me know what you think in general. I will write some tests to make sure I don't have typos and logical bugs after we agree on the general concept.

pipeline/Measurer.py
```diff
@@ -0,0 +1,88 @@
...
+            'photometry_method',
+            'aperture',
+            str,
+            'Type of photometry used. Possible values are "psf" and "aperture". '
+        )
+        self.aperture_radius = self.add_par(
+            'aperture_radius',
+            3.0,
+            [float, list],
+            'Radius of the aperture in pixels. Can give a list of values. '
```
rknop, line 25: We don't want aperture radius in pixels, we want it in units of FWHM.  Seeing varies enough from one image to the next that a constant-pixel aperture radius rarely makes sense.  (It might for the "large" aperture used to figure out aperture corrections.)

For a sky-noise dominated Gaussian-shaped PSF, the optimum aperture radius is 0.6731 FWHM.

pipeline/astrometry.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class ParsAstrometry(Parameters):
+    def __init__(self, **kwargs):
+        super().__init__()
+        self.cross_match_catalog = self.add_par(
+            'cross_match_catalog',
+            'Gaia',
+            str,
+            'Which catalog should be used for cross matching for astrometry. '
```
rknop, line 15: This is fine for here, but I wonder if we might need more information than just a string here.  E.g., now we'd use Gaia DR2.  I don't know if there's a Gaia DR3 planned, but if there were, we might switch to it.

For what goes here, it should just be a string, but that string should then be a key to something somewhere else that actually defines the catalog.  (I.e., there will be catalog grabbing code that uses this string to figure out what it actually gets.)

(So nothing needs to change here, just thinking ahead.)

pipeline/pipeline.py
```diff
@@ -0,0 +1,112 @@
...
+
+            self.add_par('example_pipeline_parameter', 1, int, 'an example pipeline parameter')
+
+            self._enforce_no_new_attrs = True  # lock against new parameters
+
+
+class Pipeline:
+    def __init__(self, **kwargs):
+        # top level parameters
+        self.pars = ParsPipeline(config.get('pipeline', {}))
```
rknop, line 33: I don't think you mean config.get here, but config.value

config.get() method you feed the config filename to in order to get the entire config hierarcy.  I believe here what you're asking for is the "pipeline" subhierarchy of the config object.

Also, config.value doesn't take a default.  We could change that if you want, or we could just mandate that there must be a pipeline (etc.) section in the config.  I'd lean towards the latter, because there are going to be some parameters in some cases that need to be filled in for the pipline to be able to do its work.  WE can always provide a default config with standard and/or dummy values for everything.

rknop: In fact, it'd be config.Config.get().value('pipeline').  (The "struct" argument to value() should generally not be used; it replaces the entire internal structure with that dictionary.  It's used internally by the config system, but using it from the outside will always be error-prone.)

pipeline/data_store.py
```diff
@@ -0,0 +1,773 @@
...
+            return args[0]
+        else:
+            return DataStore(*args, **kwargs)
+
+    def __init__(self, *args, **kwargs):
+        """
+        See the parse_args method for details on how to initialize this object.
+        """
+        # these are data products that can be cached in the store
+        self.exposure = None  # single image, entire focal plane
```
rknop, line 63: It makes sense to cache the db object, but it does not make sense to cache the pixel values for the exposure.  Usually, the only thing you do with an exposure is strip an image out of it, and each run of the pipeline is going to be working on just one image out of the exposure.  No need to carry around all that extra data.

pipeline/data_store.py
```diff
@@ -0,0 +1,773 @@
...
+        if len(args) == 0:
+            raise ValueError('Must provide at least one argument to DataStore constructor.')
+
+        if len(args) == 1 and isinstance(args[0], DataStore):
+            # if the only argument is a DataStore, copy it
+            self.__dict__ = args[0].__dict__.copy()
+            return
+
+        # catch any sessions given to the args list
+        sessions = [arg for arg in args if isinstance(arg, (sa.orm.session.Session, SmartSession))]
```
rknop, line 116: This interface could potentially lead to trouble.  I can only see two ways of it being used, both of which lead to trouble.
(1) the datastore is initialized with a SmartSession inside a with block.  Once that with block ends, the SmartSession is no longer valid, but we'll probably be keeping the datastore around.
(2) the datastore istelf initializes the session, or is passed one outside a with block, and now we've got a session that's held open as long as the datastore exist, which goes against the "don't hold open sessions" issue we talked about at length before.

I think that datastore shouldn't keep a session inside itself.

pipeline/data_store.py
```diff
@@ -0,0 +1,773 @@
...
+                if not isinstance(new_provs, list):
+                    new_provs = [new_provs]
+
+                for prov in new_provs:
+                    if not isinstance(prov, Provenance):
+                        raise ValueError(f'Provenance must be a Provenance object, got {type(prov)}')
+                    self.upstream_provs.append(prov)
+
+            # check for sessions
+            if key in ['session']:
```
rknop, line 162: Same comment as above

pipeline/data_store.py
```diff
@@ -0,0 +1,773 @@
...
+        """Get a string with the relevant inputs. """
+
+        if self.im_id is not None:
+            return f'im_id={self.im_id}'
+        elif self.exp_id is not None and self.ccd_id is not None:
+            return f'exp_id={self.exp_id}, ccd_id={self.ccd_id}'
+        else:
+            raise ValueError('Could not get inputs for DataStore.')
+
+    def get_provenance(self, process, pars_dict, upstream_provs=None, session=None):
```
rknop, line 233: So, yeah, this is the way I'd do it.  Pass a session to each method, which it uses as the argument to SmartSesssion.  Don't save it in the DataStore self.

pipeline/data_store.py
```diff
@@ -0,0 +1,773 @@
...
+    'astrometry': 'wcs',
+    'calibration': 'zp',
+    'subtraction': 'sub_image',
+    'detection': 'detections',
+    'cutting': 'cutouts',
+    'measurement': 'measurements',
+}
+
+
+class DataStore:
```
rknop, line 40: A question about how this is going to be used.

Do you visualize a single DataStore object for each run through the pipeline?

If so, the structure if it may not be flexible enough.  For instance, we're always going to have at least two images, the science image, and the reference image.  Is the idea that you don't cache the reference image in the datastore?  If so, where would the lazy loading of the reference data be handled?   I can also think of a reason why we might want to read up to four catalogs during the processing.  There's the catalog extracted from the search image, the catalog extracted from the reference image, the catalog used for matching for astrometry, and the catalog used for matching for photometry.  Where is the lazy-loading handled for the last three catalogs?

And I doubt I've thought of everything right here; there may well come other reasons why we have to load ancillary images or catalogs.  For instance, if we decide to get more sophisticated in dealing with color terms, then we might have to deal with mutpiple images at once.  While I don't think we should worry about that as we're first building the pipeline (as that adds all kinds of complication), having a datastore structure set up to be able to handle something like that would be nice (so that we don't have to rewrite everything that uses the datastore later as we realize we need to add something like that). 

We could start adding individual things for each thing like this we think of.  However, in some cases the code is the same, e.g. the catalog for the search and reference images are both produced by our code and will be loaded in the same (or almost the same) way.  A more general datastore for catalogs would make sense then, which could then be keyed here rather than defining as object attributes all of the ones we end up needing in any branch of the pipeline.

pipeline/data_store.py
```diff
@@ -0,0 +1,773 @@
...
+        if self.exposure is None:
+            if self.exp_id is None:
+                raise ValueError('Cannot get raw exposure without exp_id!')
+
+            with SmartSession(session) as session:
+                self.exposure = session.scalars(sa.select(Exposure).where(Exposure.id == self.exp_id)).first()
+
+        return self.exposure
+
+    def get_image(self, provenance=None, session=None):
```
rknop, line 354: Unless I missed it, you don't have the code here to load the actual pixel values, just the database structure.

Is the plan eventually to have code in here that secures the actual image file in local storage, and loads the actual pixel values into memory?  (The way I do it in lensgrinder, that wouldn't happen here, but the image structure itself has the information necessary for another routine to find the fits file on local storage and read it, but I know that you wanted to just keep the image in memory all the time.)

(This matters because of my earlier comment about not storing the pixel values for the entire exposure.)

pipeline/Measurer.py
guynir42: I only put these in as example parameters... clearly we need to decide exactly what goes where. 
I think there may be cases where a variable aperture is less stable but that should also be another choice. 
For example, you may have an "aperture_units" parameter that defaults to "fwhm" but can also take "arcsec" or "pixel".

pipeline/astrometry.py
guynir42: Gaia DR3 is already out since last year... 

Again, we will want to add more parameters to all of these classes, but that will happen naturally when we actually add the code.

pipeline/pipeline.py
guynir42: I just assumed `config` is a dictionary with some configuration (I don't think I even merged your config PR into this branch yet). I will have to plug it in with the right interface later.

pipeline/data_store.py
guynir42: I suppose when we add the code that cuts out the specific CCD image, it will also issue a command to the Exposure object to clear it's internal memory.

pipeline/data_store.py
guynir42: The reason for this design (which is not so clean), is that initialization goes through the DataStore, both for the individual objects (like Subtractor) or for the top level Pipeline object. I want to allow an external caller to provide a session if they choose, just like we give that option to all other function calls. 
The only way I could figure out how to do that was by parsing the session argument along with the other inputs. 
If you give a session externally, it remains in the DataStore. I'm not sure that's a problem. Lets go through examples:
```
  ds = pipeline.run(exp_id, ccd_id, session=session)
  # session is in the ds, but the block is done here anyway
```

If you want to use individual objects you could conceivably do something weird like this:

```
  ds = subtractor.run(im_id, session=session)  # parse the arguments into the ds
  ds = detector.run(ds)  # keep using the same open session
  # session in the ds, but it is closed afterwards

# outside the block there's still a reference to the session in ds:
ds = measurer(ds)  # accidentally use the same session
```

The new SQLA sessions can be safely closed and when accessed again they just start up a new connection using "autobegin" but if you want, we can disable that so that if someone tries to open a session that was closed by accident, it would raise an exception.

pipeline/data_store.py
guynir42: As I say above, this is exactly what I wanted, but the arguments are passed in through the data store, there's no way to avoid putting that session in the ds.

pipeline/data_store.py
guynir42: Yes, the datastore is used to parse a few inputs (like exp/ccd id) and then can lazy load and cache all the information you need for the analysis. 
There's a `ref_image` and `sub_image` in the datastore. But of course you may need more data to be added. I don't future proof my code (that's impossible) so if you need to add things, you'd just add them later. 

Also, because I am using "getter" functions, that have the ability to lazy load from DB, we can easily expand some of the functionality without altering the interface (for example, if we end up with more catalogs). 

Right now I have "sources" which is the catalog from the science image (I am avoiding the word catalog because I think that is a good name for the cross-matching catalogs like Gaia, instead I'm using source list). I also have "detections" which is the exact same thing only on the subtracted image. 

I don't know why we'd need a separate source list for astrometry and photometry? 

The reference image needs a list of sources so we can compare to the new one and swarp them into the same reference, I guess? 

If so, that source list should come with the ref image, not re-produced by us each time.

pipeline/data_store.py
guynir42: The way this will work is that the Exposure and Image classes would have, besides the SQLA interface to their respective rows in the DB, also some code to lazy load the data. 

I thought about something like `self._data` and `self.data` and when you ask for `self.data` it checks if `self._data` is populated and if not it fetches it from disk. 
There would be a "clear data" method that the user can opt in to calling when memory conservation is an issue.

pipeline/data_store.py
rknop: We shouldn't read the entire exposure into memory in the first place.  For 60 2048x4096 chips, that's 2GB of memory, or 4GB if read as doubles.  If then 60 processes are all reading that, we're going to get into trouble on lots of machines.  But, also, even compressed, the exposure file is half a gig; that's a lot of gratuitous I/O.

We should only read the data for the chip that we're actually processing off of the disk, not the whole thing.

guynir42: OK I can figure that out when we actually build the `Exposure` class.

pipeline/data_store.py
rknop: I'd rather disable autobegin, because that's the kind of thing that will lead to database connections exploding without us realizing that it's happening.    Using that feature completely subverts the whole practice of controlling database connections using "with" blocks.

Indeed, I'd much rather not save the database connection in the object's self at all.  The only reason to do so is if we're going to use auto-re-opening things, which I think we shouldn't use.  Given that, it doesn't make any sense to save the session in self, because it will never be usable.

pipeline/data_store.py
rknop: I don't understand why the session has to be in the datastore.  Indeed, as I said above, I think a design that requires that goes against the plan to only open sessions when they're needed.

pipeline/data_store.py
rknop: The list I was talking about for astrometry and photometry are the catalogs with the calibrated information we're matching to.

rknop: A practical issue we're going to have to confront: we can't have too many processes connecting to external services at the same time, or we're going to start getting failures.

For example, the gaia servers.  Right now with lensgrinder, when I'm running 60 chips, a single process downloads the astrometric data for all of them all at once.  That's a single connection to the gaia server.  External servers are going to start kicking us off if we make too many simultaneous connections, which will happen if each process runs independently and tries to get the astrometric data at the same time.

I do similar things for photometric data, and for downloading survey sources for building references.

One solution is local caching (which I also do) -- save the astrometric data for a field, and we never have to go to gaia again for it.  However, that doesn't help when we're doing a new field.

I don't know if this can be solved if we want to deal with each chip as a completely independent process, as the current design suggests.  Lensgrinder does have some exposure-level processes for things like this.

pipeline/data_store.py
guynir42: OK I've added a workaround. 
The session is no longer saved in the data store. 
I suggest we turn off the session auto-begin only if we see there are too many open sessions.

guynir42:
@rknop I've decided to postpone adding tests to this. 
This is bad practice but in order to make the tests work I'm going to need to add models, and that would make this PR even bigger. 
I think we should give it one more serious review and then merge and I'll start adding models and tests in a different PR. 

Please remember that lots of details are not filled out yet because I've tried to keep this simple. We'll add the details for each processing step separately in other PRs.

rknop (approved):

PR #23, opened 2023-03-30
Move database creation and migration to alembic, some associated changes.

rknop:
This builds on top of the "Pulling in my config system" change.


guynir42:
@rknop I think this will be easier for me to review after we merge the previous PR.

guynir42:
@rknop I have merged the previous PR and now we have some conflicts. Pull and merge upstream/main and then it should be much easier to so what is actually new and needs to be reviewed.

rknop:
This should be fully merged with the main branch

README.md
```diff
@@ -89,3 +89,20 @@ In that list, you should see something that has name `devshell_seechange-postgre
...
+  SEECHANGE_CONFIG=<configfile> alembic upgrade head
+```
+
+After editing any schema, you have to create new database migrations to apply them.  Do this by running something like:
+```
+  SEECHANGE_CONFIG=<configfile> alembic revision --autogenerate -m "<put a short comment here>"
+```
+The comment will go in the filename, so it should really be short.  Look out for any warnings, and review the created migration file before applying it (with `alembic upgrade head`).
+
+*NOTE:* Alembic doesn't properly handle q3c indices when autogenerating migrations. If you add any columns or tables that need to use a q3c index, you will need either to edit the migration by hand, or create a new migration (by running a`lembic revision` without `--autogenerate`) and edit it by hand.
```
guynir42, line 108:
```suggestion
*NOTE:* Alembic doesn't properly handle q3c indices when autogenerating migrations. If you add any columns or tables that need to use a q3c index, you will need either to edit the migration by hand, or create a new migration (by running `alembic revision` without `--autogenerate`) and edit it by hand.
```

alembic.ini
```diff
@@ -0,0 +1,105 @@
...
+
+[alembic]
+# path to migration scripts
+script_location = alembic
+
+# template used to generate migration file names; The default value is %%(rev)s_%%(slug)s
+# Uncomment the line below if you want the files to be prepended with date and time
+# see https://alembic.sqlalchemy.org/en/latest/tutorial.html#editing-the-ini-file
+# for all available tokens
+# file_template = %%(year)d_%%(month).2d_%%(day).2d_%%(hour).2d%%(minute).2d-%%(rev)s_%%(slug)s
```
guynir42, line 11: Uncommenting this line is not a bad idea. With skyportal, I always had a really hard time finding the correct version file, so maybe having the files ordered by date could be more convenient.

alembic/env.py
```diff
@@ -0,0 +1,72 @@
...
+            connection=session.connection(),
+            literal_binds=True,
+            dialect_opts={"paramstyle": "named"},
+        )
+
+        with context.begin_transaction():
+            context.run_migrations()
+            session.commit()
+
+def run_migrations_online() -> None:
```
guynir42, line 53: What is the difference between online and offline?

models/provenance.py
```diff
@@ -198,10 +198,10 @@ def update_hash(self):
...
 
 
-
-CodeHash.metadata.create_all(engine)
-CodeVersion.metadata.create_all(engine)
-Provenance.metadata.create_all(engine)
+# Delete these; database will be created 
+# CodeHash.metadata.create_all(engine)
+# CodeVersion.metadata.create_all(engine)
+# Provenance.metadata.create_all(engine)
```
guynir42, line 204: Yeah you can remove these lines.

models/base.py
```diff
@@ -64,11 +57,12 @@ def SmartSession(input_session=None):
...
 
+    cfg = config.Config.get()
     allowed_dirs = [
-        DATA_ROOT,
-        os.path.join(CODE_ROOT, "results"),
-        os.path.join(CODE_ROOT, "catalogs"),
-        DATA_TEMP,
+        cfg.value('path.data_root'),
+        os.path.join(_CODE_ROOT, "results"),
+        os.path.join(_CODE_ROOT, "catalogs"),
```
guynir42, line 64: Why the underscore in `_CODE_ROOT`?

tests/docker-compose.yaml
guynir42, line 1: I'm a little confused by the number of config files. This is run by github actions when it runs the tests?

tests/util/test_config.py
```diff
@@ -14,6 +14,10 @@
 # class scope.  (The tests below are ordered with this in mind.)
 
 class TestConfig:
+    # Nominally, this tests should pass even if SEECHANGE_CONFIG is not None.  However, because there
+    # are autouse fixtures in ../conftest.py, one of the side effects is that the config default
+    # gets set.
+    @pytest.mark.skipif( os.getenv("SEECHANGE_CONFIG") is not None, reason="Clear SEECHANGE_CONFIG to test this" )
```
guynir42, line 20: Is this actually going to get run in the test suite?

models/base.py
```diff
@@ -8,47 +8,40 @@
...
     # open a new session and close it when outer scope is done
     if input_session is None:
-        with Session() as session:
+        if _Session is None:
+            cfg = config.Config.get()
+            url = ( f'{cfg.value("db.engine")}://{cfg.value("db.user")}:{cfg.value("db.password")}'
+                    f'@{cfg.value("db.host")}:{cfg.value("db.port")}/{cfg.value("db.database")}' )
+            engine = sa.create_engine( url, future=True, poolclass=sa.pool.NullPool )
+            
+            _Session = sessionmaker(bind=engine, expire_on_commit=True)
```
guynir42, line 43: Why not just initialize these two things at the beginning of the file?

guynir42 (commented):
Looks good! I have mostly questions but also a few suggestions.

alembic/env.py
rknop: ...I don't know.  This is template code that comes out of alembic.  I just modified it to do what we need.

models/base.py
rknop: I'm thinking of it as an internal variable.

But, also: I don't like these two lines.  We shouldn't have a default data directory that's anywhere in the code directory.  I've fought with this in other packages, that want to read and write where the code lives, and it makes it really tough to use in some circumstances.  (In particular: you may well be running the code out of a place that's not writeable.)  Default data directories should all come out of the config file, and should not be hardcoded inside the code to write into the code directory.

What do you visualize as the reason for the CODE_ROOT variable?

tests/docker-compose.yaml
rknop: I don't know what you mean.  Which config files are you talking about?

tests/docker-compose.yaml is the docker-compose file you use to run the tests.  You can do it on your local machien too; see the README.md file for more informtion.  github actions also uses it.

tests/util/test_config.py
rknop: Not as it exists right now, because of the auto-run frameworks in conftest.py

models/base.py
rknop: So that you can set a different config file at runtime.

If you initialize them in the file, then you have to use a hardcoded or environment-variable session location (or config file).

If you do it this way, then you can call Config.init() with a different config file, and then the results of that config will be picked up here when the session is created.

models/base.py
guynir42: Actually I can see why this would be a good practice, to keep the code read-only. 
I have occasionally ran into cases where I needed the code folder, but maybe if we decide up-front that we never write to it it may not even be necessary to have that folder accessible.

models/base.py
guynir42: yeah that's pretty good.

guynir42 (approved):
OK this looks good to me, but feel free to make any final modifications before you merge.

PR #28, opened 2023-04-24
Exposure model

guynir42:
This PR adds the basic functionality of the Exposure model, although a lot of the details of how to actually load images from files remains to be added separately. 

I need this mostly so I have a bare-minimum data model to play with in tests.


guynir42:
@rknop Ignore all the commits except the last two. I forgot to checkout "main" before starting work...

models/exposure.py
```diff
@@ -1,7 +1,238 @@
...
+class Exposure(Base, SpatiallyIndexed):
 
-class Exposure(Base):
     __tablename__ = "exposures"
 
+    header = sa.Column(JSONB, nullable=False, default={}, doc="Header of the raw exposure. ")
+
+    filename = sa.Column(sa.Text, nullable=False, index=True, unique=True, doc="Filename for raw exposure. ")
+
+    mjd = sa.Column(
```
rknop, line 50: Something we either have to think about, or have to decide not to think about, is the time standard we're going to use....   That is, TAI vs. UTC vs. etc. See, for instance : https://www.ipses.com/eng/in-depth-analysis/standard-of-time-definition/

They're all different by less than a minute, so my inclination is to decide not to think about this.  Just say "hey, whatever the header says the MJD is, that we're gonna believe", and as long as we're only looking at headers from the same telescope, it will all be consistent.

models/exposure.py
```diff
@@ -1,7 +1,238 @@
...
+        nullable=False,
+        index=True,
+        doc="Modified Julian date of the exposure. (MJD=JD-2400000.5)"
+    )
+
+    exp_time = sa.Column(sa.Float, nullable=False, index=True, doc="Exposure time in seconds")
+
+    filter = sa.Column(sa.Text, nullable=False, index=True, doc="Filter name")
+
+    instrument = sa.Column(sa.Text, nullable=False, index=True, doc="Instrument name")
```
rknop, line 61: I noticed in CCDData that you pass something called an instrument_object.

Is the idea that somewhere we'll have an Instrument class, and then we'll make a subclass for each actual instrument, and then this field is going to be an index into a mapping of name->subclass?

rknop: A reason to have an instrument table instead of just classes is so that things like num_ccds, width, height can go into that table instead of being repeated every row of the exposures table.

models/exposure.py
```diff
@@ -1,7 +1,238 @@
...
+
+    gallat = sa.Column(sa.Double, index=True, doc="Galactic latitude of the target. ")
+
+    gallon = sa.Column(sa.Double, index=False, doc="Galactic longitude of the target. ")
+
+    ecllat = sa.Column(sa.Double, index=True, doc="Ecliptic latitude of the target. ")
+
+    ecllon = sa.Column(sa.Double, index=False, doc="Ecliptic longitude of the target. ")
+
+    def __init__(self, *args, **kwargs):
```
rknop, line 83: I don't think I've used an explicit __init__ with an SQLAlchemy class before.  I'm a little confused here, though.

I would expect this __init__ to be called every time an object is instantiated.  If so, then it doesn't make sense to have lines like `self.num_ccds = self.get_instrument_object().get_num_ccds()`, as it's replacing a field that was read from the database with something calcualted.  If __init__ was only called when the database record was first created, then this would make sense, but that's at odds with general python __init__.

rknop: OK, I read up a bit, and indeed __init__ is not called when the object is pulled out of the database.  Which is very weird to me, I have to admit.

models/exposure.py
```diff
@@ -1,7 +1,238 @@
...
+    @orm.reconstructor
+    def init_on_load(self):
+        self._data = None
+        self.instrument_object = None
+
+    def parse_instrument_name(self):
+        if self.filename is None:
+            return
+
+        for name, class_ in inspect.getmembers(models.instruments, inspect.isclass):
```
rknop, line 129: This is making an assumption about filenames that I don't think we can make -- specifically that there's going to be a unique instrument name in the filename that we can use as a key into an instrument table.  I can envision two failure modes: first, the filename being too generic and not having an instrument name.  Second, the instrument name being a string that shows up in other instruments' filenames.

Filenames of FITS files are not part of a FITS standard or anything like that, so writing code that makes assumptions about those filenames is hazardous.

We need more sophisticated instrument detection.  I would argue for two different things.  First, the ability to just specify it manually : pass in the instrument (name, or database id if we go with a database table of instruments) when instantiating an Exposure.  Second, as a fallback, have a generic "is this exposure from my instrument" method that each instrument class has to implement.  The latter will be slow, as each instrument will go and poke at the file to decide if that file belongs to it, which is why I also want the option to manually specify the instrument.

models/exposure.py
```diff
@@ -1,7 +1,238 @@
...
+
+    def __str__(self):
+        return self.__repr__()
+
+    def save(self):
+        pass  # TODO: implement this! do we need this?
+
+    def load(self, ccd_ids=None):
+        if ccd_ids is None:
+            ccd_ids = list(range(self.num_ccds))
```
rknop, line 146: While we can choose to define our internal ccd_ids to be indexed starting from 0, it might be convenient if we made that indexing an instrument-specific thing.  (That way, we don't have to maintain our own map of ccd id to ccd identifier.)  Different instruments are going to have different ways of specifying their CCDs.  E.g., ztf has a ccd indentifier that's like a coordinate position, rather than just an index.  Perhaps instead of list(range(self.num_ccds)), we could cal an instrument's all_ccds() method or some such.

Another thing: it might be worth having a ccd table, with a foreign key that points back to the instrument table.  This could store things like offsets of the corners (and center) of the CCD relative to the overall exposure coordinates.  (This could also just be encoded into the class, not in a table.  One advantage of having it in a table instead of just in code is that you might be able to do database searches on those parameters.)

models/exposure.py
```diff
@@ -1,7 +1,238 @@
...
+
+    def parse_header_keys(self):
+        """
+        Parse the relevant columns: mjd, project, target,
+        num_ccds, width, height, exp_time, filter, telescope
+        from self.header and into the column attributes.
+        """
+
+        for k, v in self.header:
+            norm_k = normalize_header_key(k)
```
rknop, line 213: This can't be in a general pipeline.utils -- it needs to be something that's in the instrument class.

If we try to write one set of code that works for all instruments, we will hate ourselves later.  Much cleaner just to have the defaults in an instrument superclass, and then have the right answer for a given instrument (if it's not the default) in the instrument subclass.

Look at : https://gitlab.com/robknop/curveball/-/blob/master/bin/exposuresource.py  Look at the properties starting at `ra_degrees_keyword`.  (I override those properties in subclasses of the ExposureSource class as necessary.)

models/exposure.py
guynir42: That's a good point. At some point we will need to match the instrument row to the instrument class (where the code to do things lives). 
If we just assume each Exposure's instrument name is matched to a hard-coded instrument class, then we can just load an object of that class (or the pertinent info like width/height) whenever the Exposure is loaded. But it doesn't enforce that each instrument name is indeed matched to an instrument class. 
On the other hand, if we have an instrument table we should still allow the user to input the instrument as a string and match that internally using a foreign key. 
I think I can come up with something like that.

models/exposure.py
guynir42: This just makes it easier to match keywords. I don't think there's anything instrument specific about it. Right now it just removes spaces and underscores and puts it in upper case. Think of it as a more generic `.upper()` function, to be used for matching the keys. 
The list of keywords that get matched could be a different story. I guess we should allow each instrument object to define column names for those fields.

models/provenance.py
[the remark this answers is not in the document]
rknop: Oh, found it in utils.get_latest_provnance()

models/instrument.py
```diff
@@ -0,0 +1,610 @@
...
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.name = 'DECam'
+        self.telescope = 'Blanco'
+        self.aperture = 4.0
+        self.focal_ratio = 2.7
+        self.square_degree_fov = 3.0
+        self.pixel_scale = 0.2637
+        self.read_time = 20.0
+        self.read_noise = 7.0
```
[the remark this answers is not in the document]
guynir42: OK so parsing header keys is moved to the Instrument class. You can use the base class parsing or override/enhance the functions that do that. 
There are two categories of keywords to look for: with/without default. E.g., "mjd" has no default, and if the header doesn't have that value the Exposure object will be left with `mjd=None` (we can check that value later on). 
The other category will use the instrument defaults if the key is missing in the file. E.g., the "gain" will just use the instrument's default value if not given in the header.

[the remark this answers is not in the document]
guynir42: The only question is: do we want to save the "parsed" header values in a separate dictionary on each Exposure? Do we even want to save the entire fits header into the DB? maybe we should only save the "important" keywords, after parsing them into common names (e.g., `exp_time` instead of `EXPTIME`) and leave the full header only on the file on disk?

tests/models/test_exposure.py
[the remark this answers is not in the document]
guynir42: You can scope a fixture to "module" scope that will be executed once per test file.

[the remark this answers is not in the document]
guynir42: I like being able to inherit a big pile of tests from a base class. But that seems like a fairly rare use case, and we should implement that when needed. 
For normal operations I don't really see the justification for the extra boilerplate, but it is fairly easy to add it later.

models/instrument.py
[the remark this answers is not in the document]
rknop: In previous pipelines, I've saved the entire header (as a JSONB) to the database, but that does seem a little excessive.  The only reason really to do that is if you'd want to pull out a table of values that includes some header keywords, and I *have* done that, but rarely.

If we're just going to save the important keywords, then we already have database fields for those things, so I don't think there's a value in saving that additionally in a header field.

tests/models/test_exposure.py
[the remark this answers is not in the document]
rknop: So you're cool with a heterogeneous style?  (That is, some test files define functions, some define a class.)  I'm fine with that, but previously you said you didn't like it.

default_config.yaml
```diff
@@ -1,7 +1,7 @@
 path:
-  seechange_data: null
+  data_root: null
   data_temp: null
-
+  server_data: null
```
rknop, line 4: See comment in README re: storing server location.

Rather than having any archive server information in the config at all, we should just have relative file paths.  data_root says where to find it on the local system, but on the archive, the system we put up to connect to the archive should be responsible for figuring out any absolute paths.  (It may also end up being more complicated than that, as we might archive old images to HPSS or some such.) 

There should be some kind of overall "system" or "collection" variable, so that the same server can handle multiple different sets of images.  E.g., with curveball, I have a "dev" overall collection where I muck about.  There's also a "emily" collection where Emily does her dev work.  Then there's a "ztf" collection where Autumn and I are building lightcurves from ztf, etc., and that's a production collection.  All of these connect to the same server for storing them in the archive on NERSC; server figures out where the root directory is based on the collection it gets, and then we have relative filepaths after that.

pipeline/utils.py
```diff
@@ -51,9 +65,87 @@ def get_latest_provenance(process_name, session=None):
...
+        Choose one of the output formats:
+        'datetime', 'Time', 'float', 'str'.
+
+    Returns
+    -------
+    datetime, Time, float or str
+    """
+    if dateobs is None:
+        dateobs = Time.now()
+    elif isinstance(dateobs, (int, float)):
```
rknop, line 89: We could also put in an if dateobs > 2400000 , and if so, interpret it as a JD, otherwise interpret it as an MJD.

(This codes in a Y2K-like bug that we'll hit in about 6000 years.)

README.md
```diff
@@ -136,3 +140,242 @@ After editing any schema, you have to create new database migrations to apply th
...
+the exposure time, and the filter used (or filter array). 
+
+The Exposure object will contain the filename and path to the underyling raw data files. 
+Ideally these data files are not modified at all from when they were generated at the telescope. 
+Additional processing steps, such as bias/flat should be done on the next processing level
+(i.e., when generating Image objects) so the details of the processing can be tracked.
+The Exposure's `filename` column will contain information on the name and path of the data
+files relative to some data root folder. The root folder can be changed using the config
+(e.g., when moving a mount point) and can contain multiple root folders for storing
+data on a server and locally. The `filename` is always relative to that, so the database
```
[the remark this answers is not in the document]
guynir42: Filepath may be better...

README.md
```diff
@@ -136,3 +140,242 @@ After editing any schema, you have to create new database migrations to apply th
...
+entry remains valid even if the data is moved.
+For multiple-section instruments, there would be multiple data arrays saved to disk. 
+These can be saved in one file (e.g., using a multi-extension FITS file) or in multiple files.
+In that case, the `filename_extensions` will contain an array of strings, 
+each containing the last part of the filename that completes the partial filename in `filename`.
+These can be normal file extensions:
+
+```
+filename = 'Camera_Project_2021-01-01T00:00:00.000.fits'
+filename_extensions = ['.bias', '.flat']
```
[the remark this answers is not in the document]
guynir42: I guess this is a bad example because I used bias/flat. Those would also have the same `filename_extensions` array, but I will give a better example.

README.md
```diff
@@ -136,3 +140,242 @@ After editing any schema, you have to create new database migrations to apply th
...
+   or `return range(10)` for a 10-CCD instrument with integer section IDs.
+   A more general case could be `return ['A', 'B', 'C']`, which highlights 
+   the fact the section IDs can be strings, not only integers. 
+ - `check_section_id`: verify the input section ID is of the correct type and in range. 
+ - `_make_new_section`: make a new section with hard coded properties. 
+   If any of the properties are identical across all sections, leave them as `None`. 
+   If the properties are different but known in advance, this method will be used
+   to fill them up for each section ID, using a lookup table or data file. 
+ - `get_section_offsets`: the geometric layout of the instrument's sections.  
+   if each section does not define an `offset_x` and `offset_y`, these values
```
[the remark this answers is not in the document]
guynir42: If you have the size and offset in x and y, that generally covers it?

pipeline/utils.py
guynir42: I like that!

README.md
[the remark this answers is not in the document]
rknop: Yeah, if there's a size, and we define offset to the center (or anywhere, really, but let's do the center), then you have what you need.

rknop (approved):

PR #34, opened 2023-05-31
Image model

guynir42:
The Image model will serve as the place where the pre-processed data is saved. 
After doing bias/dark/flat calibration we will save the resulting images as files associated with "image" table rows. 
Each Image corresponds to data from a single section (unlike the Exposure which covers the entire focal plane, with multiple sections). 
Images can also be generated from other images, e.g., when making a stack/coadd, a subtraction, a super-flat, and so on. 
Single-image objects will link to a single Exposure. Multi-image objects will link to a list of other Image objects.


models/image.py
```diff
@@ -1,10 +1,133 @@
...
+    )
+
+    source_images = orm.relationship(
+        'Image',
+        cascade='save-update, merge, refresh-expire, expunge',
+        doc=(
+            "Images used to produce a multi-image object "
+            "(e.g., an images stack, reference, difference, super-flat, etc)."
+        )
+    )
```
rknop, line 39: Doesn't this orm.relationship create a 1:1 as written?  It needs to be a 1:many.

models/image.py
```diff
@@ -1,10 +1,133 @@
...
+        sa.Boolean,
+        nullable=False,
+        default=False,
+        index=True,
+        doc=(
+            "Whether this image is a multi-image object. "
+            "Multi-image objects will have a null exposure, "
+            "and a list of Image objects from which it was derived. "
+        )
+    )
```
rknop, line 51: Is this needed?  We could just check that exposure is NULL.

Or are we anticipating images that will have a NULL exposure and also not be multi-image?

What I'm worried about here is redundant information (which then can create issues for keeping the database consistent); if there are no source_images, then we know it's not a mulit-image, so we don't need a flag.

models/image.py
```diff
@@ -1,10 +1,133 @@
...
+
+    combine_method = sa.Column(
+        sa.Text,
+        nullable=True,
+        index=True,
+        doc=(
+            "Type of combination used to produce this multi-image object. "
+            "One of: coadd, subtraction. "
+        )
+    )
```
rknop, line 61: This will need a lot more detail.  We could use multiple different coadd or subtraction methods; if you really want to track that, then that needs to go into a provnance, or needs somehow else to keep track of coadd and subtraction algorithms.

If you just want to answer the question "is this a coadd or a subtraction" without worrying about actual algorithm, or if the algorithm goes into the provanan ce but you still want this flag, an ENUM would make more sense than a text column.

models/image.py
```diff
@@ -1,10 +1,133 @@
...
+
+    type = sa.Column(
+        sa.Text,
+        nullable=False,
+        index=True,
+        doc=(
+            "Type of image. One of: science, reference, difference, bias, dark, flat. "
+        )
+    )
```
rknop, line 70: Again, if we're predefining the types, I'd make it an ENUM.

models/image.py
```diff
@@ -1,10 +1,133 @@
...
+
+    mjd = sa.Column(
+        sa.Double,
+        nullable=False,
+        index=True,
+        doc=(
+            "Modified Julian date of the exposure (MJD=JD-2400000.5). "
+            "Multi-exposure images will have the MJD of the first exposure."
+        )
+    )
```
rknop, line 112: We should probably define whether it's the MJD of the start of the exposure, or the MJD of the middle of the exposure.  MOST observatories give you the MJD of the start of the exposure in the header, but arguably the mean MJD is more useful.  Usually it doesn't matter because the timescale of the events we care about is enough longer than the typical exposure time that the difference is unimportant.  But, it's probably worth being clear about what we mean.

I'd probably be in favor of making it the mean MJD ; you could then also have the mean MJD for a coadded exposure.  (For a coadd, the MJD is a much more fraught concept, of course.)

models/image.py
guynir42: I need to add a many-to-many table.

models/image.py
guynir42: Ok, we can use the exposure_id.

models/image.py
guynir42: Let's keep it simple. I'll add an enum.

models/image.py
guynir42: ok

models/image.py
guynir42: I'm going to add an `end_mjd` to each image, so that will track the start/end times of either a single exposure or a stack of images. The middle mjd will be the average of those two, by definition. And the exposure time is the sum of exposure times, regardless of whether you use sum or average, because that is just a scalar multiplication of the pixel values, it doesn't change the fact that the image contains photons from X seconds of exposure.

models/image.py
```diff
@@ -1,10 +1,133 @@
...
+    exp_time = sa.Column(
+        sa.Float,
+        nullable=False,
+        index=True,
+        doc="Exposure time in seconds. Multi-exposure images will have the total exposure time."
+    )
+
+    filter = sa.Column(sa.Text, nullable=False, index=True, doc="Filter name")
+
+    section_id = sa.Column(
```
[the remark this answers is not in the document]
guynir42: I think we should not link the SensorSections directly to the image. 
The sensor sections are an optional feature that is used to "patch" the global instrument properties. 
In a sense, you don't care that an image was taken by a specific instance of a section ID, you care that it was taken on "CCD number 2".

guynir42:
@rknop I think most of this is finished, so you can have another look. I still need to add tests so please do not merge yet...

guynir42:
@rknop This seems ready for review. I added multiple tests, and solved our issues with the naming conventions and the single vs. multiple files issue.

models/exposure.py
```diff
@@ -352,28 +439,19 @@ def update_instrument(self, session=None):
...
-    import numpy as np
-    rnd_str = lambda n: ''.join(np.random.choice(list('abcdefghijklmnopqrstuvwxyz'), n))
-
-    e = Exposure(f"Demo_test_{rnd_str(5)}.fits", exp_time=30, mjd=58392.0, filter="F160W", ra=123, dec=-23, project='foo', target='bar')
-
-    session = Session()
-
-    session.add(e)
-    session.commit()
+    filename = '/home/guyn/Dropbox/python/SeeChange/data/DECam_examples/c4d_221104_074232_ori.fits.fz'
```
rknop, line 455: I realize this is something we're never going to run (we'll import exposure, not python it), but still it's a little ugly to have a hardcoded path on your machine here....

I have, of course, done exactly this sort of thing.  But, when I'm behaving better, what I try to do is create a subdirectory called "muckabout" or something like that where I can, well, muck about.  There I have the little hardcoded test thingies that I run on my own machine to make sure to be able to look into what's going on.

models/instrument.py
```diff
@@ -838,6 +906,79 @@ def _get_header_values_converters(self):
...
+        Returns
+        -------
+        index: int
+            The index of the HDU in the FITS file.
+            Note that the index is 0-based, as this value is
+            used in the astropy.io.fits functions/objects,
+            not in the native FITS format (which is 1-based).
+        """
+        cls.check_section_id(section_id)
+        return int(section_id) + 1
```
rknop, line 935: This isn't going to work with string section_ids (which is probably what we need to use, since we can't count on the HDUs showing up in any particular order in the FITS file).

rknop: ...I see you overrode this in the DECam class.  Perhaps put a note in the documentation that an instrument needs to override this if it will ever get a section_id that isn't an integer, and if we can't assume that the mapping of section id to order in FITS file is 1:1?

models/exposure.py
guynir42: I think I left that in by accident... I'll fix it right up

rknop (approved):

PR #39, opened 2023-06-15
End-to-end pipeline tests

guynir42:
This PR runs some tests to see that the data manages to flow from the Exposure level to the end products. 
It doesn't test the individual processing units, just that data is formatted and packaged correctly. 
It also checks that the Provenances are managed correctly.


guynir42:
@rknop I still need to add a few tests, but the basic data flow from start to finish seems to work. This is a good time to go over the data models and how they interact.

guynir42:
@rob I think this is ready for review now.

rknop (approved):

rob:
Thank you, @guynir42! I'll pass this over to @rknop!

PR #45, opened 2023-07-17
Integrate the util/config.py config system into pipeline/top_level.py

rknop:
… ; write tests in test_pipeline.py to ensure that the config is being read and passed into pipeline parameters, and that parameters can be overridden at runtime.

### DIFF
--- a/default_config.yaml
+++ b/default_config.yaml
@@ -30,3 +30,36 @@ storage:
     # (in this example, the first six characters of the provenance unique hash)
     name_convention: "{ra_int:03d}/{short_name}_{date}_{time}_{section_id}_{filter}_{prov_hash:.6s}"
 
+# ======================================================================
+# Pipeline components
+#
+# For documentation on the parameters, see the Parameters subclass
+# in the file that defines each part of the pipeline
+
+pipeline: {}
+
+preprocessing:
+  use_sky_subtraction: True
+
+extraction: {}
+
+astro_cal:
+  cross_match_catalog: Gaia
+
+photo_cal:
+  cross_match_catalog: Gaia
+
+subtraction:
+  method: hotpants
+
+detection:
+  subtraction: false
+  threshold: 5.0
+
+cutting:
+  cutout_size: 20
+
+measurement:
+  photometry_method: aperture
+  aperture_radius: 3.0
+  real_bogus_version: null
--- a/default_config.yaml
+++ b/default_config.yaml
@@ -53,8 +53,8 @@ subtraction:
   method: hotpants
 
 detection:
-  subtraction: false
-  threshold: 5.0
+  subtraction: true
+  threshold: 3
 
 cutting:
   cutout_size: 20
--- a/tests/pipeline/test_pipeline.py
+++ b/tests/pipeline/test_pipeline.py
@@ -1,4 +1,5 @@
 import os
+import pytest
 
 import sqlalchemy as sa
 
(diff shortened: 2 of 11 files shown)


tests/pipeline/test_pipeline.py
```diff
@@ -81,6 +82,51 @@ def check_datastore_and_database_have_everything(exp_id, sec_id, ref_id, session
     # TODO: add the cutouts and measurements, but we need to produce them first!
 
 
+def test_parameters( config_test ):
+    """Test that pipeline parameters are being set properly"""
+
+    # Verify that we _enforce_no_new_attrs works
+    kwargs = { 'pipeline': { 'keyword_does_not_exist': 'testing' } }
+    with pytest.raises( AttributeError ):
```
guynir42, line 90:
```suggestion
    with pytest.raises( AttributeError, match='object has no attribute' ):
```

tests/pipeline/test_pipeline.py
```diff
@@ -81,6 +82,51 @@ def check_datastore_and_database_have_everything(exp_id, sec_id, ref_id, session
...
+        failed = Pipeline( **kwargs )
+
+    # Verify that we can override from the yaml config file
+    pipeline = Pipeline()
+    assert pipeline.preprocessor.pars['use_sky_subtraction']
```
guynir42, line 95: Just FYI it also works by doing `pars.use_sky_subtraction`

tests/pipeline/test_pipeline.py
```diff
@@ -81,6 +82,51 @@ def check_datastore_and_database_have_everything(exp_id, sec_id, ref_id, session
...
+                 }
+    pipelinemodule = { 'preprocessing': 'preprocessor',
+                       'subtraction': 'subtractor',
+                       'detection': 'detector',
+                       'cutting': 'cutter',
+                       'measurement': 'measurer'
+                      }
+
+    pipeline = Pipeline( pipeline={ 'example_pipeline_parameter': -999 } )
+    assert pipeline.pars['example_pipeline_parameter'] == -999
```
guynir42, line 118:
```suggestion
    # TODO: this is based on a temporary "example_pipeline_parameter" that will be removed later
    pipeline = Pipeline( pipeline={ 'example_pipeline_parameter': -999 } )
    assert pipeline.pars['example_pipeline_parameter'] == -999
```

tests/seechange_config_test.yaml
```diff
@@ -3,3 +3,6 @@ preloads:
 
 db:
   host: seechange_postgres
+
+subtraction:
+  method: testing_testing
```
guynir42, line 8: Wouldn't this be problematic when we actually run tests with subtractions? Or should we just explicitly test each method by setting this parameter when we setup the pipeline for each test?

guynir42 (commented):
Looks good. I have minor comments. 
Please let me know when you are done fixing them, then I will approve and merge (or let you self-merge, which I like better).

default_config.yaml
```diff
@@ -30,3 +30,36 @@ storage:
...
+#
+# For documentation on the parameters, see the Parameters subclass
+# in the file that defines each part of the pipeline
+
+pipeline: {}
+
+preprocessing:
+  use_sky_subtraction: True
+
+extraction: {}
```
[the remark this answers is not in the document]
rknop: I'll put it in, but we're going to want to revisit this when we get to it for real.  A detection threshold of 3 is going to net a *lot* of junk.  Most people who do this use something closer to a 5σ threshold.

default_config.yaml
```diff
@@ -30,3 +30,36 @@ storage:
...
+  cross_match_catalog: Gaia
+
+photo_cal:
+  cross_match_catalog: Gaia
+
+subtraction:
+  method: hotpants
+
+detection:
+  subtraction: false
```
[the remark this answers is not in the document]
rknop: Yeah, but again we'll revisit this when we actually implement a subtraction.

tests/seechange_config_test.yaml
rknop: We'll have to do something to test each pipline; it may be a different config, or just updating the config at runtime.

default_config.yaml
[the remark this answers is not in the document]
guynir42: Sure, just wanted it to be consistent with the current defaults. 
If you feel like changing the defaults right now (in the Parameters object and here) you can do that too.

guynir42 (approved):

PR #48, opened 2023-08-01
Use sep package to do SExtractor work, in python

guynir42:


models/source_list.py
```diff
@@ -31,6 +38,13 @@ class SourceList(Base, FileOnDiskMixin):
         )
```
rknop, line 38: You didn't change this in this PR, but : do we need the is_sub column of the source list table?  Once the source_list points back to the image, we should be able to determine if it's a subtraction by using the is_sub property of the image object.  (If we wanted to do it as a database join, we'd join to image.ref_image is null and image.new_image is null.)

From the point of view of database normalization, I'd prefer not to have redundant (and, thus, possibly conflicting) information unless there's a compelling reason for it.

models/source_list.py
```diff
@@ -53,10 +67,106 @@ class SourceList(Base, FileOnDiskMixin):
...
+            value = value.to_records(index=False)
+
+        if not isinstance(value, np.ndarray) or value.dtype.names is None:
+            raise TypeError("data must be a pandas DataFrame or numpy recarray")
+
+        self._data = value
+        self.num_sources = len(value)
+
+    @property
+    def mjd(self):
```
rknop, line 119: THis property shouldn't be here; you can get it with a join to the image table.

rknop: That was a stupid comment. :D

My brain thought it was looking at a column definition.  Please ignore.

models/source_list.py
```diff
@@ -53,10 +67,106 @@ class SourceList(Base, FileOnDiskMixin):
...
+
+        if self.data is None:
+            raise ValueError("Cannot save source list without data")
+
+        filename = self.invent_filename()
+
+        if self.is_sub:
+            filename += '_detections'
+        else:
+            filename += '_sources'
```
rknop, line 149: I might just call it _sources always.  I don't think there's any benefit in having different filenames based on where the thing came from; at some level, the Source List class itself shouldn't care.

models/source_list.py
```diff
@@ -53,10 +67,106 @@ class SourceList(Base, FileOnDiskMixin):
...
+    'mjd',
+    'ra',
+    'dec',
+    'filter',
+    'filter_short',
+    'telescope',
+    'instrument',
+    'instrument_object',
+]:
+    setattr(SourceList, att, property(fget=lambda self, att=att: getattr(self.image, att) if self.image is not None else None))
```
rknop, line 172: You've already explicitly defined @property mjd above

pipeline/detection.py
```diff
@@ -110,10 +113,40 @@ def run(self, *args, **kwargs):
...
+        -------
+        sources: SourceList
+            The list of sources detected in the image.
+            This contains a table where each row represents
+            one source that was detected, along with all its properties.
+
+        """
         # TODO: finish this
-        # TODO: this should also generate an estimate of the PSF
+        # TODO: this should also generate an estimate of the PSF?
```
rknop, line 133: Thought needed.  The reason not to do it here is that you aren't going to want to generate a PSF estimate when detecting sources on a subtraction image.  So, we should separate source extraction and PSF estimation (with the latter requiring the former as a prerequisite).

pipeline/detection.py
```diff
@@ -110,10 +113,40 @@ def run(self, *args, **kwargs):
...
+
+        data = image.data
+
+        # see the note in https://sep.readthedocs.io/en/v1.0.x/tutorial.html#Finally-a-brief-word-on-byte-order
+        if data.dtype == '>f8':  # TODO: what about other datatypes besides f8?
+            data = data.byteswap().newbyteorder()
+        b = sep.Background(data)
```
rknop, line 140: Thought is going to be required here.  My experience is that you want to use different sky-subtraction algorithms in different places.  For extragalactic fields, the default sextractor algoirithm usually does a great job.   (Is that exact same algorithm what sep.Background uses by default?)  However, for crowded galactic fields, it oversubtracts the sky by a lot, and a slower and more complicated algorithm does better; the one I've used is in Bijaoui, 1980, A&A, 84, 81

rknop: Another thought : we don't always want to subtract a background.  The most obvious case is going to be where the image is a subtraction!  But we might also end up with sky-subtracted images already (e.g. if there's a custom processing pipeline that really does a good job of subtracting sky images, and we don't want to mess it up by doing something on top of that here).  Perhaps a field in the image table that indicates if the image is already background subtracted?

rknop (commented):
Just a few comments, some of which are "we will need to think about this in the future".

models/source_list.py
guynir42: I guess you are right. The idea was that it would be simple to tell the subtraction sources from the regular sources. Maybe there's an advantage when you can index on the `is_sub` column?

models/source_list.py
guynir42: You can also delete comments :)

models/source_list.py
guynir42: oops!

pipeline/detection.py
guynir42: The problem is that you need stars to find a PSF, and then you want to PSF to find stars in a more sensitive way (using a matched filter with the correct PSF). It would maybe make sense to make PSF finding an option to the sextractor run.

pipeline/detection.py
guynir42: I like having that extra field. Maybe even worthwhile to save the median sky brightness and/or the sky noise level so we will not have to re-estimate it. 
The thing is that even if you reduce the mean sky (or the local sky value) it still leaves the noise. If you get an Image that is already subtracted but you don't know how much was there before, you'd be missing a big piece of information which is the sky variance. 
This seems like we should make an issue with a "decision needed" on it and discuss on the next meeting.

models/source_list.py
rknop: If we index on the foreign key (which you should always do), and index in image.new_image and image.ref_image, then the database performance should be similar.

guynir42: got it.

guynir42: I can take out that column and add a property instead.

models/source_list.py
rknop: But are they different?  They're the result of this algorithm run on an image.  There's nothing different about the algorithm, so what this process adds to the filename shouldn't vary.  You can tell a subtracted image vs. a regular image based on the rest of the filename.  (Subtracted images do have a different filename pattern than regular images, don't they?   I would argue that yes, they really ought to, for the reasons you say; but, once that happens, then we don't need the _detections vs _sources difference any more.)

pipeline/detection.py
rknop: Yeah... so there are different modes of operation.

In fact, it's more involved... I'll make an issue.  Let's just leave this comment as is for now.

pipeline/detection.py
guynir42: yeah, I agree.

models/source_list.py
guynir42: What happens if the naming convention that you choose for SourceList files isn't exactly the same as the one for Image files? I'm not 100% sure you'd have the subtraction/not subtraction in the filename. 

In my mind the SourceList is actually the same model for two very different datasets (just like Image) so making them obviously different will only help us. 

But I will have another look at the whole naming of these things and maybe I will change my mind.

rknop: Since each source list comes from an image, ideally the source list filename will be the image filename with a different extension at the end (e.g. .sources.hdf5 or .sources.fits or whatever format we save it into).

I don't think of the source list datasets being all that different.  A subtraction image is different from a basic telescope image, I agree.  But, in both cases, the source list is just "what sources did I find on that image".  Deciding that they are new detections or not is really meaning that the user then ascribes to it-- if you are taking a deeper image than's been taken before of this field, you're going to have new detections in the raw image.   Detections and sources are really two different names for the same thing.

models/exposure.py
```diff
@@ -115,6 +115,31 @@ def clear_cache(self):
...
+    ----------
+    image_type: str or None
+
+    Returns
+    -------
+    str or None:
+         The short version of the image type, or None if the input was None.
+    """
+    short_types = {
+        "science": "sci",
```
rknop, line 136: We probably also want to distinguish between individual science images and science stacks (e.g. if we want to do an end-of-night stack subtraction for when the same field is imaged multiple times in one night, or in one week, or whatever).

models/base.py
```diff
@@ -322,10 +322,10 @@ def safe_mkdir(cls, path):
     )
 
     format = sa.Column(
-        im_format_enum,
+        file_format_enum,
         nullable=False,
         default='fits',
-        doc="Format of the file on disk. Should be fits or hdf5. "
+        doc="Format of the file on disk. Should be fits, hdf5, csv or npy. "
```
rknop, line 328: Don't act on this now, but : some year we might also want to support Parquet files.  (The library pyarrow handles those.  I came across these in another project.)

I'm not sure that parquet gives us anything we don't already have with hdf5, though.  I think the interface might be simpler for reading a smaller dataset (e.g. a table of objects from a single image).

models/exposure.py
```diff
@@ -115,6 +115,31 @@ def clear_cache(self):
 im_type_enum = Enum("science", "reference", "difference", "bias", "dark", "flat", name='image_type')
 
 
+def get_short_image_type(image_type):
```
rknop, line 118: What do you see as the use case for this?  It's possible that adding this will just add the ability for confusion.  I think we'd be better off just defining which types exist (sci, ref, sub, etc.), and documenting that.  I see where this could add convenience, but it also adds new potential for "sometimes" errors (e.g. somebody calls get_short_image_type("subtraction") expecting it to return "sub").

models/source_list.py
```diff
@@ -141,14 +145,15 @@ def save(self):
...
+            filename = self.image.invent_filename()
 
-        if self.is_sub:
-            filename += '_detections'
-        else:
-            filename += '_sources'
+        if filename.endswith(('.fits', '.h5', '.hdf5')):
+            filename = os.path.splitext(filename)[0]
 
         # TODO: use FITS and astropy tables instead?
```
rknop, line 155: The advantage of using FITS is that the sources file will be more easily read by hand by somebody who wants to do ad-hoc stuff.  That is, reading the .npy file is easy, but the FITS table is a bit more self-documenting, generally.  (I guess you could have all named columns and such in the .npy file as well; not sure what the sep returns.)

tests/conftest.py
```diff
@@ -267,8 +267,9 @@ def reference_entry(exposure_factory, provenance_base, provenance_extra):
 
                 session.commit()
 
+
 @pytest.fixture
 def config_test():
     # Make sure the environment is set as expected for tests
-    assert os.getenv( "SEECHANGE_CONFIG" ) == "/seechange/tests/seechange_config_test.yaml"
+    # assert os.getenv( "SEECHANGE_CONFIG" ) == "/seechange/tests/seechange_config_test.yaml"
     return Config.get( os.getenv("SEECHANGE_CONFIG") )
```
rknop, line 275: I don't understand why you removed this assert ; the tests are designed to go with that specific config file, so we want to make sure we've got the right config file before proceeding.

rknop (approved):

PR #70, opened 2023-08-30
replace integer FITS extensions with strings when loading an Image

guynir42:


guynir42:
@rknop this is (thankfully) a real easy one for once.

rknop (approved):

Issue #69, opened 2023-08-30
Remove Enums


guynir42:
I've just about had it with Enums under Postgres/SQLAlchemy

Here's a list of things we can't do:
- remove an Enum value, unless you go through a very complicated dance
- rename the text that represents an Enum value (e.g., Sci->sci will not be possible unless you manually do it)
- put the Enum definition in the base class (mixin) and inherit from others, as that breaks the tests where we do `create_table` (this works in alembic, though). 

Here's a counter-proposal:
Use integers, and have a getter/setter lookup table to translate the integers into text. 
Advantages:
- you can easily add values, just make the setter function accept new strings. 
- you can easily rename existing data, just alter the lookup tables. 

Disadvantages:
- if you are looking at the database directly, you will see a number, not a value (e.g., "2" instead of "Bias"). 

That's it. Please discuss.

nugent68:
You can do the integer and then just have a lookup table for it. Gives you some flexibility. This is what we did for the timing on the various parts of the pipeline.

guynir42:
Addressed by #73

Issue #40, opened 2023-06-19
Bad flag for images (and other data objects?)


guynir42:
Should we have some way to flag an image is bad? 
I am thinking a boolean, a text with a short tag and a comment with room for more details. 
Should this be a common thing we add to all data objects, or just add this manually to the Image class and maybe later to others?

guynir42:
32 (or 64) bit mask for specific badness which will be "bitwise or"ed when you ask for only the good images, source lists, etc. 
Each data model below Image will have to join back all the parents to see if any of them was bad (in other words join the badness).

guynir42:
Addressed by #73

Issue #37, opened 2023-06-06
enum type enforcements at set time


guynir42:
We could add some python-side code to some of the models such that when you try to add a value that isn't included in the relevant Enum, it would raise an error right away, instead of when you commit the data.

PR #73, opened 2023-08-31
Replace Enums with dict and conversion functions

guynir42:
First step: replace image type and file formats with dictionaries and converters. 
Next: to add a "bad mask" with a bit flag to all data types.


guynir42:
Addresses #69

models/enums_and_bitflags.py
```diff
@@ -0,0 +1,237 @@
...
+    25: 'Bad Pixel',
+    26: 'Bleed Trail',
+}
+cutouts_badness_inverse = {v.lower().replace(' ', ''): k for k, v in cutouts_badness_dict.items()}
+
+source_list_badness_dict = {
+    41: 'X-Match Failed',
+    42: 'Big Residuals',
+    43: 'Few Stars',
+    44: 'Many Stars',
```
rknop, line 227: Too Many Stars?

Many stars doesn't sound like a bad thing to me.

models/enums_and_bitflags.py
```diff
@@ -0,0 +1,237 @@
...
+    11: 'avro',
+    12: 'netcdf',
+    13: 'jpg',
+    14: 'png',
+    15: 'pdf',
+}
+
+allowed_image_formats = ['fits', 'hdf5']
+image_format_dict = {k: v for k, v in file_format_dict.items() if v in allowed_image_formats}
+image_format_inverse = {v: k for k, v in image_format_dict.items()}
```
rknop, line 27: Random musing : it's too bad python doesn't have a symmetric 1:1 mapping class.  As far as  I know, having an inverse dictionary is the only reasonable way to do it.

models/enums_and_bitflags.py
```diff
@@ -0,0 +1,237 @@
...
+    44: 'Many Stars',
+}
+source_list_badness_inverse = {v.lower().replace(' ', ''): k for k, v in source_list_badness_dict.items()}
+
+# join the badness:
+data_badness_dict = {0: 'Good'}
+data_badness_dict.update(image_badness_dict)
+data_badness_dict.update(cutouts_badness_dict)
+data_badness_dict.update(source_list_badness_dict)
+data_badness_inverse = {v.lower().replace(' ', ''): k for k, v in data_badness_dict.items()}
```
rknop, line 236: To catch future programming errors, it might be worth adding a test to make sure that none of (image_badness_dict, cutouts_badness_dict, source_list_badness_dict) (and anything we add later) have 0 as a key, and that none of them have overlapping keys with each other.

models/enums_and_bitflags.py
guynir42: I agree to many sources is better. I think this sort of thing could happen when SExtractor fails and thinks every pixel is a source.

models/enums_and_bitflags.py
guynir42: We only need to calculate it once here, so it is not so bad.

models/enums_and_bitflags.py
guynir42: I agree about testing against zeros. 
Overlaps in principle are not illegal. I can imagine places where different models could have overlapping keywords.

rknop: But the dict that is used for lookup will only have the latest one that got updated

guynir42: So you're worried that someone might alter the dict during runtime? We can probably make up a bidirectional dictionary for fun.

guynir42: and of course this has already been implemented: https://pypi.org/project/bidict/

guynir42:
@rknop this one is really hacky but in the end it works. I am not sure how all of these crazy outer joins will scale once we have lots of images, but I hope it doesn't end up growing the number of rows a lot more than the total number of images. 

Also, I'm leaving the implementation for Cutouts for a different PR, when I start looking at that model more closely.

guynir42:
Addresses #37 #40 #73

models/enums_and_bitflags.py
rknop: I'd still recommend changing this to "too many sources" and "too few sources"

guynir42: Yes, I will do that. The beauty of this new setup is that we can make all the changes we want down the line. There's no need to commit to these enums at all. In fact we should probably discuss which values we want and how to name them.

guynir42:
@rknop I fixed the two comments on the Enums. It is worth while that you also look at the bitflag mechanics. That's the really complicated one and I'm sure you'll have a lot to say about it...

rknop (approved):
This looks good.  Go ahead and merge it once the alembic two-heads issue is resolved.

Issue #27, opened 2023-04-18
update readme with instructions for installing out of docker


guynir42:
If a user wants to run SeeChange without a docker container, there should be some instructions on how to install all the required software to make it work locally. 
E.g., postgres with q3c, hotpants, sextractor.

guynir42:
Addressed by #77

PR #77, opened 2023-09-07
Document how to run SeeChange locally (not in docker)

guynir42:
There are some advantages to being able to run locally (outside a docker container). 
1) You can run tests directly from the IDE.
2) You don't need to worry if you have problems with your docker install (e.g., cannot edit files inside the container for some reason).
3) Provides another cross check that things work as expected on different setups. 

There are a few non-trivial installation steps that need to happen for this to work. 
I will try to document them here, but I'm sure this file will get updated the next time someone actually installs this way from scratch. 

Also, I'm opening a "docs" folder, and we should consider breaking off some of the already large readme in to files in that folder.

### DIFF
--- a/docs/run_locally.md
+++ b/docs/run_locally.md
@@ -124,6 +124,15 @@ using:
 git submodule update
 ```
 
+Note that you can just do 
+
+```bash
+git submodule update --init
+```
+
+from the root directory, which will also initialize any 
+submodules that have not been initialized yet.
+
 #### Setting up environment variables
 
 Some environmental variables are used by SeeChange.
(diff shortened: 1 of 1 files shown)


guynir42:
Addresses #27

docs/run_locally.md
```diff
@@ -0,0 +1,192 @@
...
+```bash
+cd extern/nersc-upload-connector
+git submodule init
+```
+
+If those packages require updates, you can do that from the root SeeChange directory
+using:
+
+```bash
+git submodule update
```
rknop, line 124: You can also just run 

git submodule update --init

in the main repository to have it update all submodules; the --init flag tells it to grab all submodules as necessary.

rknop: (This is in place of the two steps you mention.)

rknop: Feel free to change this or not, and then squash and merge.

rknop (approved):

docs/run_locally.md
guynir42: yeah, good idea. I will add it and merge.

Issue #54, opened 2023-08-13
Additional Image properties/columns


guynir42:
There are a few more nullable columns to be added to the Image class, such as sky background / sky noise, image FWHM or seeing, image zero point, and so on. 

These are optional properties that could go in a header but are better as their own column which makes searches much easier. 

NOTE: The background field, for example, is important as it helps us know if the image already had background removed. We should probably also have fields for "was dark/bias subtracted" and "was flat adjusted". 

NOTE 2: the properties like FWHM really depends on the source extraction, so it may be better to have it as a join with the SourceList or other object that holds a PSF (in which case, it would have its own provenance and it is not obvious which one should be loaded!). 
For zero point it is the same thing, only with the photometric calibration object (same problem with the choosing of which provenance to use).

guynir42:
we need to record where exactly the sky was subtracted (was it at the source- meaning the Exposure was already subtracted) or was it done during coaddition, etc. And then a simple flag for "was sky subtracted" is also useful.

guynir42:
Addressed by #78

guynir42:
Addressed by PR #139

PR #78, opened 2023-09-07
Add columns to Exposure and Image models

guynir42:
We clearly need a few searchable columns on Exposure and Image. 

There are a lot of open questions so I've decided to open a draft PR and have a discussion here. 

Some things I'm not sure about:
 - Image class should have similar columns to Exposure, but probably will have different values because these are sections of the main Exposure? 
 - Sky background mean/rms: should we do the mean of the sky? median? is rms a good measurement? 
 - PSF width (seeing) and zero point and limiting mag and sky background: all these measurements can be made downstream in the pipeline, and could each have multiple versions/provenances attached. If we want to use them as the value for each Image/Exposure we will have to decide which version to pick and when to update the numbers (e.g., when a new SExtractor parameter is used). 
 - What is the definition of limiting magnitude (e.g., 3 sigma? 5 sigma?) 
 - Column names: I tried to make them very clear but also short and not over-complicated. Opinions on naming are welcome as always. 
 - I am probably forgetting some more columns. Please suggest more. 
 - Should we index on all of these? If we envision using them as searchable columns, then probably indexing is the right thing to do. On the other hand, when did we ever need to filter a query based on sky rms?

### DIFF
--- a/models/exposure.py
+++ b/models/exposure.py
@@ -220,6 +220,43 @@ class Exposure(Base, FileOnDiskMixin, SpatiallyIndexed):
         doc='Name of the target object or field id. '
     )
 
+    sky_bkg = sa.Column(
+        sa.Float,
+        nullable=True,
+        doc="Sky background level in counts per second per pixel. "
+    )
+
+    sky_rms = sa.Column(
+        sa.Float,
+        nullable=True,
+        doc="Sky background RMS in counts per second per pixel. "
+    )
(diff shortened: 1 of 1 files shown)


guynir42:
Addresses #54

guynir42:
Closing as this is now being done in https://github.com/c3-time-domain/SeeChange/pull/139.

PR #79, opened 2023-09-07
Add cone_search as a sqlalchemy filter to models.base.SpatiallyIndexed

### LINKED ISSUE #35: Spatial indexing
Should add a "cone search" method to SpatiallyIndexed and make sure it works on Exposure and Image.

rknop:
resolves #35

### DIFF
--- a/models/base.py
+++ b/models/base.py
@@ -1013,19 +1013,5 @@ class SpatiallyIndexed:
 
         return func.q3c_radial_query( self.ra, self.dec, ra, dec, rad )
 
-    @cone_search.expression
-    def cone_search( cls, ra, dec, rad, radunit='arcsec' ):
-        if radunit == 'arcmin':
-            rad /= 60.
-        elif radunit == 'arcsec':
-            rad /= 3600.
-        elif radunit == 'radians':
-            rad *= 180. / math.pi
-        elif radunit != 'degrees':
-            raise ValueError( f'SpatiallyIndexed.cone_search: unknown radius unit {radunit}' )
-
-        return func.q3c_radial_query( cls.ra, cls.dec, ra, dec, rad )
-
-
 if __name__ == "__main__":
     pass
--- a/models/base.py
+++ b/models/base.py
@@ -1,5 +1,6 @@
 import sys
 import os
+import math
 import hashlib
 import pathlib
 import logging
@@ -15,6 +16,7 @@ from sqlalchemy.types import Enum
 
 from sqlalchemy.orm import sessionmaker, declarative_base
 from sqlalchemy.ext.declarative import declared_attr
+from sqlalchemy.ext.hybrid import hybrid_method
 from sqlalchemy.orm.exc import DetachedInstanceError
 from sqlalchemy.dialects.postgresql import UUID as sqlUUID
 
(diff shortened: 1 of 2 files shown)


models/base.py
```diff
@@ -978,6 +980,38 @@ def calculate_coordinates(self):
...
+        if radunit == 'arcmin':
+            rad /= 60.
+        elif radunit == 'arcsec':
+            rad /= 3600.
+        elif radunit == 'radians':
+            rad *= 180. / math.pi
+        elif radunit != 'degrees':
+            raise ValueError( f'SpatiallyIndexed.cone_search: unknown radius unit {radunit}' )
+
+        return func.q3c_radial_query( self.ra, self.dec, ra, dec, rad )
```
guynir42, line 1014: Wow that is really easier than I thought it would be :)

tests/models/test_image.py
```diff
@@ -278,6 +279,65 @@ def test_image_coordinates():
...
+            image3 = Image( 'three.fits', ra=120.0005, dec=10., provenance=provenance_base, nofile=True, **kwargs )
+            image4 = Image( 'four.fits', ra=60., dec=0., provenance=provenance_base, nofile=True, **kwargs )
+            session.add( image1 )
+            session.add( image2 )
+            session.add( image3 )
+            session.add( image4 )
+
+            sought = session.query( Image ).filter( Image.cone_search(120., 10., rad=1.02) ).all()
+            assert ( set( [ pathlib.Path( s.filepath ).name for s in sought ] )
+                     == { 'one.fits', 'two.fits' } )
```
guynir42, line 313:
```suggestion
            assert image1.id in [im.id for im in sought]
            assert image.id in [im.id for im in sought]
```

guynir42: This has two advantages:
- You don't depend on other tests to clean up for this to work (i.e., imagine the database is not empty when you start this test). 
- You don't need to know the filenames for the images (which I intend to modify in the near future when we enforce having a file on disk before committing an image). 

Also, this is the way I've been testing these kinds of queries until now.

tests/models/test_image.py
```diff
@@ -278,6 +279,65 @@ def test_image_coordinates():
...
+                     == { 'one.fits', 'two.fits', } )
+
+            sought = session.query( Image ).filter( Image.cone_search(60, -10, 1.) ).all()
+            assert len( sought ) == 0
+
+            with pytest.raises( ValueError, match='.*unknown radius unit' ):
+                sought = Image.cone_search( 0., 0., 1., 'undefined_unit' )
+        finally:
+            for i in [ image1, image2, image3, image4 ]:
+                if ( i is not None ) and sa.inspect( i ).persistent:
```
guynir42, line 338: Nice

tests/models/test_image.py
```diff
@@ -278,6 +279,65 @@ def test_image_coordinates():
...
+                    session.delete( i )
+
```
guynir42, line 340:
```suggestion
session.commit()
```

guynir42 (commented):
Looks good. I think this is a really important thing to have. 
I wonder if there is a way to test that the q3c indices are there. 
I've seen them being added and removed by alembic and I'm not sure what is going on.

tests/models/test_image.py
guynir42: Ok, that's true. But I still would test on the ids not on the filenames, and I would not assume the DB has no other images. This is a good way to make the tests break later on and never figure out why (when some other test leaves an Image in the DB and you suddenly - and only sometimes - have a failing test).

guynir42 (approved):

Issue #36, opened 2023-06-06
Four corners


guynir42:
Need to have a mixin that keeps track of the four corners of the image.
Or maybe this can just be part of the Image class? 
I think the corners should be arranged as upper/lower for RA/Dec so it is clear how to search against a coordinate ("is this coordinate inside the square?") instead of the image's left/right/up/down, so it would avoid the question of the image orientation.

guynir42:
@rknop should I just add this to the Image class?

PR #84, opened 2023-09-07
Add the FourCorners mixin

rknop:
Addresses #36


tests/models/test_image.py
```diff
@@ -279,6 +288,101 @@ def test_image_coordinates():
     assert abs(image.gallon - 160.922) < 0.01
 
 
+# Really, we should also do some speed tests, but that
+# is outside of the scope of the always-run tests
+def test_four_corners( provenance_base ):
+    # There is a small chance that this test will randomly fail
+    # because the demo_image fixture (and others) that are
+    # generated with a random RA and Dec, if they just happen
+    # to fall within the radius of one of the searches below.
```
guynir42, line 297:
```suggestion
# There is a small chance that this test will randomly fail
# because the demo_image fixture (and others) that are
# generated with a random RA and Dec, if they just happen
# to fall within the radius of one of the searches below.
@pytest.mark.flaky(reruns=2)
def test_four_corners( provenance_base ):
    
```

tests/models/test_image.py
```diff
@@ -279,6 +288,101 @@ def test_image_coordinates():
...
+                                provenance=provenance_base, nofile=True, **kwargs )
+            session.add( image1 )
+            session.add( image2 )
+            session.add( image3 )
+            session.add( imagepoint )
+            session.add( imagefar )
+
+            sought = session.query( Image ).filter( Image.containing( 120, 40 ) ).all()
+            assert ( set( [ pathlib.Path( s.filepath ).name for s in sought ] )
+                     == { 'one.fits', 'two.fits', 'three.fits' } )
```
guynir42, line 351:
```suggestion
            assert {image1.id, image2.id}.issubset([im.id for im in sought])
```

guynir42: Same comment as in the conesearch PR. You can use sets if you have to test multiple images in "sought".

tests/models/test_image.py
```diff
@@ -279,6 +288,101 @@ def test_image_coordinates():
...
+
+            sought = Image.find_containing( imagefar, session=session )
+            assert len(sought) == 0
+
+            sought = session.query( Image ).filter( Image.within( image1 ) ).all()
+            assert ( set( [ pathlib.Path( s.filepath ).name for s in sought ] )
+                     == { 'one.fits', 'two.fits', 'three.fits', 'point.fits' } )
+
+            sought = session.query( Image ).filter( Image.within( imagefar ) ).all()
+            assert len(sought) == 0
```
guynir42, line 380: Here I would also do a `not in` test against the image ids, instead of assuming that there are not other images in the database.

tests/models/test_image.py
```diff
@@ -613,8 +717,8 @@ def test_image_from_decam_exposure(decam_example_file, provenance_base):
     # assert im.dec == -26.25
     assert im.ra != e.ra
     assert im.dec != e.dec
-    assert im.ra == 116.23984530727733
-    assert im.dec == -26.410038282561345
+    assert im.ra == 116.32126671843677
+    assert im.dec == -26.337508447652503
```
guynir42, line 721: I'm curious: what made the coordinates change?

guynir42 (commented):
This is good. I have a lot of comments but nothing major.

alembic/versions/2023_09_07_1927-b33a5b72da8b_image_fourcorners.py
```diff
@@ -0,0 +1,58 @@
...
+# revision identifiers, used by Alembic.
+revision = 'b33a5b72da8b'
+down_revision = 'e78c1e8bec33'
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('images', sa.Column('ra00', sa.REAL(), nullable=False))
```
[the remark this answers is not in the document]
rknop: REAL is 32-bit

For the corners of the image, that should be good enough, and since there are eight of these fields, it's worth using the smaller size.

(32-bit floats have 23 bits in the mantissa.  At an angle of 360 degrees, 360/2^23 = about 0.15 arcseconds.  That's not good enough for positions of objects (it is for LS4, but not for everything), so we want to use doubles in SpatiallyIndexed, but it's plenty good enough to decide if an image overlaps a spot on the sky.)

models/base.py
```diff
@@ -978,6 +981,152 @@ def calculate_coordinates(self):
...
+    """A mixin for tables that have four RA/Dec corners"""
+
+    ra00 = sa.Column( sa.REAL, nullable=False, index=True, doc="RA of the low-RA, low-Dec corner (degrees)" )
+    ra01 = sa.Column( sa.REAL, nullable=False, index=True, doc="RA of the low-RA, high-Dec corner (degrees)" )
+    ra10 = sa.Column( sa.REAL, nullable=False, index=True, doc="RA of the high-RA, low-Dec corner (degrees)" )
+    ra11 = sa.Column( sa.REAL, nullable=False, index=True, doc="RA of the high-RA, high-Dec corner (degrees)" )
+    dec00 = sa.Column( sa.REAL, nullable=False, index=True, doc="Dec of the low-RA, low-Dec corner (degrees)" )
+    dec01 = sa.Column( sa.REAL, nullable=False, index=True, doc="Dec of the low-RA, high-Dec corner (degrees)" )
+    dec10 = sa.Column( sa.REAL, nullable=False, index=True, doc="Dec of the high-RA, low-Dec corner (degrees)" )
+    dec11 = sa.Column( sa.REAL, nullable=False, index=True, doc="Dec of the high-RA, high-Dec corner (degrees)" )
```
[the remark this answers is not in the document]
rknop: *We* are going to be typing it a lot... or, at least, I already did.  I'd just assume leave it as it is.  Given that there are four of them, it'll probably be pretty obvious to anybody looking at it, and the documentation does say what it is.

models/base.py
```diff
@@ -978,6 +981,152 @@ def calculate_coordinates(self):
...
+
+        Parameters
+        ----------
+          ras, decs: list of float
+             Four ras and four decs
+
+        Returns
+        -------
+          Two new lists (ra, dec) sorted so they're in the order:
+          (lowRA,lowDec), (lowRA,highDec), (highRA,lowDec), (highRA,highDec)
```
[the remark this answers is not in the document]
rknop: Or if the image isn't square to the sky.  Look at image2 in test_four_corners in test_image.py

models/base.py
```diff
@@ -978,6 +981,152 @@ def calculate_coordinates(self):
...
+
+        Returns
+        -------
+          An expression usable in a sqlalchemy filter
+
+        """
+
+        return func.q3c_poly_query( self.ra, self.dec,
+                                    sqlarray( [ fourcorn.ra00, fourcorn.dec00, fourcorn.ra01, fourcorn.dec01,
+                                                fourcorn.ra11, fourcorn.dec11, fourcorn.ra10, fourcorn.dec10 ] ) )
```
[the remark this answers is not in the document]
rknop: The code would be exactly the same except for that change, though; it's just that "self" is a class when called in a class context.  It works with just the single method -- the tests use it as a class method.  (...in fact, will Python do that anyway?  We may not even need the hybrid decorator here.)

models/base.py
```diff
@@ -978,6 +981,152 @@ def calculate_coordinates(self):
...
+                                   f"  AND LEAST(i.ra00, i.ra01, i.ra10, i.ra11 ) <= :ra "
+                                   f"  AND GREATEST(i.dec00, i.dec01, i.dec10, i.dec11 ) >= :dec "
+                                   f"  AND LEAST(i.dec00, i.dec01, i.dec10, i.dec11 ) <= :dec" ),
+                          { 'ra': ra, 'dec': dec } )
+            query = sa.text( f"SELECT i.id FROM temp_find_containing i "
+                             f"WHERE q3c_poly_query( {ra}, {dec}, ARRAY[ i.ra00, i.dec00, i.ra01, i.dec01, "
+                             f"                                          i.ra11, i.dec11, i.ra10, i.dec10 ])" )
+            objs = sess.scalars( sa.select( cls ).from_statement( query ) ).all()
+            sess.execute( sa.text( "DROP TABLE temp_find_containing" ) )
+            return objs
```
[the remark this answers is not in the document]
rknop: I don't think it's worth the time figuring out the arcane sqlalchemy of the right subquery.  SQLAlchemy is enough of a nightmare trying to figure out how to do things that are straightfoward in SQL already.  Also, we won't *really* know which we want until we can test it on a big enough database to see the speed trials.

I'm sure we can get something that produces the same results, but the point of using the temporary table was so that the query that didn't have indexes to take advantage of would be working on a much smaller table than the full images table.

tests/models/test_image.py
rknop: We don't want to use a subset here; it's not just that these images are found, it's also that the other ones are not found.  So, the sets have to be equal.

tests/models/test_image.py
rknop: The code had NAXIS1 and NAXIS2 backwards in a wcs.pixel_to_world call in image.from_exposure.  NAXIS1 is width and NAXIS2 is height.  If you want to find the coordinates of x along the horizontal axis and y along the vertical axis, you need to do wcs.pixel_to_world(x, y).  But, if you want to find that pixel in the numpy data array, it's data[y, x].

models/base.py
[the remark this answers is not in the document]
guynir42: It isn't too late to do a search-and-replace :) 

I think adding "corner" into those names is going to make things a lot clearer. It will be harder to do later on.

models/base.py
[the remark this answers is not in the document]
guynir42: If you replace "self" with "cls" and the code stays the same, then this is fine. In this case the hybrid will allow you to also call this as an expression. Just that usually the expression is something more complicated. It doesn't have to be.

models/base.py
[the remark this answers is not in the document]
guynir42: sounds good to me.

tests/models/test_image.py
guynir42: see my comment on the other PR.

tests/models/test_image.py
guynir42: great!

guynir42 (approved):
Looks good!

PR #86, opened 2023-09-08
Write md5sum when file is saved locally iff archive is null

rknop:
The first part of the two-part thing we talked about.


guynir42 (approved):
Looks good!

guynir42:
Merging it as I want to write the other part and everyone else is still asleep.

PR #90, opened 2023-09-12
Make auto-incrementing ID an optional mixin, use other unique things for primary key of provenance, code versions

rknop:
This will solve the problem of creating Provenances that turn out to be identical to what's in the database, but not knowing the primary key without a session.  The primary key is now the hash (which was unique anyway, and so might as well be the primary key), which can be calculated without any reference to anything done on the database server (like the auto-incrementing counter that most models use as their primary key).  Likewise, the "version" is now the primary key of code versions, as that was already unique.

This also includes the changes from PR #89 (which I merged on my branch).  After that one is merged to main, I'll make sure it's all synced right.


guynir42:
I'm not sure this will actually solve the problem. When you make a provenance you can only get the right hash if you also know the upstreams. You'd probably still need to go to the database...

rknop:
It doesn't solve every problem, but it solves some problems.

If you already have the previous provenances, then you don't have to go to the database to create the one for the step you're about to do right now.

It solves the problem of creating a provenance for an exposure, which lets me take out some of those database connections you objected to in the decam pull PR.

models/provenance.py
```diff
@@ -52,7 +51,7 @@ def update(self, session=None):
         if git_hash is None:
             return  # quietly fail if we can't get the git hash
         with SmartSession(session) as session:
-            hash_obj = session.scalars(sa.select(CodeHash).where(CodeHash.hash == git_hash)).first()
+            hash_obj = session.scalars(sa.select(CodeHash).where(CodeHash.id == git_hash)).first()
```
guynir42, line 54: Are you sure you also changed CodeHash's id and not only the CodeVersion?

models/provenance.py
```diff
@@ -259,11 +257,11 @@ def update_hash(self):
             process=self.process,
             parameters=self.parameters,
             upstream_hashes=self.upstream_hashes,
-            code_version=self.code_version.version
+            code_version=self.code_version.id
         )
         json_string = json.dumps(superdict, sort_keys=True)
 
-        self.unique_hash = base64.urlsafe_b64encode(hashlib.sha256(json_string.encode("utf-8")).digest()).decode()[:20]
+        self.id = base64.urlsafe_b64encode(hashlib.sha256(json_string.encode("utf-8")).digest()).decode()[:20]
```
guynir42, line 264: Here's something unrelated to this PR but it would be a really good opportunity to solve this:

Can we make sure this string does not contain underscores? It would make our filenames more easy to parse. 

I'm not sure what the best way to remove them. I used `urlsafe` which gets rid of other things like dashes (if I recall) and replaces them with underscores, so maybe it would be good enough just to not use `urlsafe`. 
Another option is to just decide here (the only place where we make these strings) that we will always use `replace` and change the underscores to something else. 

What do you think?

pipeline/data_store.py
```diff
@@ -327,15 +327,8 @@ def get_provenance(self, process, pars_dict, upstream_provs=None, session=None):
...
-            existing_p = session.scalars(
-                sa.select(Provenance).where(
-                    Provenance.unique_hash == prov.unique_hash
-                )
-            ).first()
-
-            if existing_p is not None:
-                prov = existing_p
+            prov.update_id()
+            prov = session.merge( prov )
```
guynir42, line 331: Yeah, this is much easier now.

guynir42 (commented):
Most of these comments are minor. It looks really good. I think this will make life easier down the line.

alembic/versions/2023_09_12_1726-93d7c3c93a06_code_provenance_pk.py
```diff
@@ -0,0 +1,103 @@
...
+    op.drop_column( 'code_hashes', 'id' )
+    op.alter_column( 'code_hashes', 'hash', new_column_name='id' )
+    op.create_primary_key( 'pk_code_hashes', 'code_hashes', [ 'id' ] )
+
+    op.drop_column( 'code_hashes', 'code_version_id' )
+    op.drop_column( 'provenances', 'code_version_id' )
+    op.drop_index('ix_code_versions_version', table_name='code_versions')
+    op.drop_column( 'code_versions', 'id' )
+    op.alter_column( 'code_versions', 'version', new_column_name='id' )
+    op.create_primary_key( 'pk_code_versions', 'code_versions', [ 'id' ] )
```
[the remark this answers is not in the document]
rknop: What primarily motivated this was models.base.safe_merge ; that works as is if the primary key for all models is "id".

I originally defined a class variable _pk that defined what the primary key was, and then did a getattr in safe_merge.  However, I decided that that was going to be more fiddily if we ever needed to create another class with something besides id as the primary key, and it was simpler to just keep id as the primary key for everything.

models/provenance.py
rknop: It's sad that python's base64 encode has an altchars for + and /, but not _

One solution would be to go from base64 to md5sum.  md5sum is all hexidecimal digits.  (hexits?)

rknop: ...or, just take the b64encode off.  Replace hashlib.sha256().digest() with hashlib.sha256().hexdigest() ; same result.  20 hexidecimal digits is almost certainly going to be safe, even if it's 4x fewer combinations than 20 base64 characters.

rknop: We *could* make the id field a UUID rather than a string.  I think that will have more efficient indexing, but indexing of the provenances is not going to be an issue.  (We're not going to have millions of provenances in the database... or, if we do, I'm very sad.)

alembic/versions/2023_09_12_1726-93d7c3c93a06_code_provenance_pk.py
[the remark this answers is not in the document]
guynir42: OK, that's a good reason. I guess it is kind of a hassle to update it all to point to id but since you already took care of that, it would probably save us headaches in the future.

devshell/docker-compose.yaml
```diff
@@ -44,5 +48,35 @@ seechange:
...
+         condition: service_healthy
+    volumes:
+      - type: bind
+        source: ..
+        target: /seechange
+    working_dir: /seechange
+    user: "${USERID:?err}:${GROUPID:?err}"
+    entrypoint: [ "tail", "-f", "/etc/issue" ]
+
+
```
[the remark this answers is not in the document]
guynir42: ok, that makes sense.

models/provenance.py
guynir42: I was really happy to have more than 16 options per digit because I could reduce the number of digits. For example, if you shorten to 6 digits for filenames it helps to have more options. 

Do you think we could just replace the underscores with dashes or something?

guynir42: The `urlsafe` replaces + with - and / with _. I actually like having no slashes, but would prefer to replace underscore with dash. I don't see why we shouldn't just add a `replace('_', '-')` after the `urlsafe` gets rid of the dashes and slashes first.

rknop: We could... I guess it bothers me aesthetically to have an ecoding that's not reversable (since + is already replaced with -), though the hash is already not reversable, so probably not a big deal.

I'm trying to think of any other ASCII character we could put in a filename, but all of them (other than a-zA-Z0-9, -, and _) are potentially troublesome (even though legal).  We could always just use greek letters....

How about base32? That gives you 30 bits of variation in 6 characters, which is probably enough.

models/provenance.py
rknop: Yeah -- I renamed hash to id, and changed it from unique and index to primary key (which makes it unique and indexed).  Since I don't put in the AutoIDMixin here, it gets rid of the old auto-incrementing integer id.

models/provenance.py
guynir42: How would that work, exactly?

rknop: Just replace base64.urlsafe_b64encode with base64.b32hexencode

rknop: I did this, but used b32encode instead of b32hexencode ; the latter is new in Python 3.10, and the current test dockerfile has python 3.9.  (I'll be bumping that to 3.11 with one of my upcoming pull requests.)

guynir42 (approved):

PR #91, opened 2023-09-15
Enum inheritance

rknop:
As I started working on astrometry, I wanted a couple of new enums, and realized that each time we added an enum, we were going to write a new nearly-identical *_converter() function.  I refactored it all to use classes and class methods so that now there's just a single convert function, and we just have to define the dictionary and allowed values for each new enum.  (We still have to create the hybrid properties for each database model that uses an enum.)

Note that the diffs are a bit of a disaster -- the diff algorithm thought there was all kinds of clever mixing, making it hard to read.  It might be better to just look at the full enums_and_bitflags.py file.

### DIFF
--- a/tests/models/test_enums.py
+++ b/tests/models/test_enums.py
@@ -17,7 +17,7 @@ def test_enums_zero_values():
 
 def test_converter_dict():
     # Probably should test them all, but test just these
-    #  three and trust that it it works, then the inheritance
+    #  three and trust that if it works, then the inheritance
     #  is working for all of them.
 
     assert ImageTypeConverter.dict == {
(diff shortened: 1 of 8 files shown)


tests/models/test_enums.py
```diff
@@ -1,11 +1,75 @@
...
-    assert 0 not in image_type_dict
-    assert data_badness_dict[0] == 'Good'
\ No newline at end of file
+    assert 0 not in FormatConverter.dict
+    assert 0 not in ImageTypeConverter.dict
+    assert data_badness_dict[0] == 'Good'
+
+def test_converter_dict():
+    # Probably should test them all, but test just these
+    #  three and trust that it it works, then the inheritance
```
guynir42, line 20:
```suggestion
    #  three and trust that if it works, then the inheritance
```

guynir42 (approved):
Looks good. I have a couple of minor comments:

PR #93, opened 2023-09-21
Algorithm to calculate sky flats

guynir42:
Add a simulator class that makes semi-realistic images of stars, including many of the imperfections and artefacts in real images. 

Then use this to make images we can add together to produce sky flats, and check that they are not bumpy (due to stars) and that the background values we get are consistent with the simulation inputs.


rknop (commented):
While this looks fine as it is, really doing image simulation is a pretty big issue.  In addition to needing to actually implement bleeding, there's also the issue that very bright stars have bigger psfs (leading to very extended wings around bright stars), plus galaxies.

For testing flatfielding-building, what would be the advantage of using this over just getting a night's worth of images from a couple of different telescopes and doing it on real data?  One thing I hope to do "soon" is add a PTF archive "instrument" that will let us get PTF images which should be kind of close to what we'll see in LS4 (in terms of depth and undersampling).

guynir42:
The thing I was going for, when I started out and thought this simulator will be a quick and simple thing, was to get the actual flat (from the pixels and the telescope vignette pattern). I wanted to have ground truth which you don't get from looking just at data. 

I will tag you for review when I add the actual flat field algorithm on top of the simulator.

improc/simulator.py
```diff
@@ -0,0 +1,886 @@
...
+        pass
+
+    def electrons_to_adu(self):
+        """
+        Convert the number of electrons in each pixel
+        to the number of ADU (analog to digital units)
+        that will be read out.
+        """
+        self.average_counts = self.electrons * self.sensor.pixel_gain_map
+        self.noise_var_map = self.average_counts + self.sensor.read_noise ** 2
```
rknop, line 718: Poisson noise should be based on the number of electrons, not ADU.

improc/simulator.py
```diff
@@ -0,0 +1,886 @@
...
+        self.oversampling = oversampling
+
+        if self.atmos_psf_mode.lower() == 'gaussian':
+            self.atmos_psf_image = make_gaussian(self.atmos_psf_pars['sigma'] * self.oversampling)
+            self.atmos_psf_fwhm = self.atmos_psf_pars['sigma'] * 2.355
+        else:
+            raise ValueError(f'PSF mode not recognized: {self.atmos_psf_mode}')
+
+
+class SimStars:
```
rknop, line 319: Obviously not for this PR, but: do you have any ambitions for SimGalaxies?  That's a pretty gigantic project, and it's probably worth looking into GalSim or something like that which has already done simulations of galaxy profiles.

improc/simulator.py
```diff
@@ -0,0 +1,886 @@
...
+        # ) - self.sensor.pixel_bias_std ** 2
+        self.sensor.pixel_bias_map = np.random.normal(
+            self.sensor.bias_mean,
+            self.sensor.pixel_bias_std,
+            size=self.pars.imsize
+        )
+
+        self.sensor.gain_mean = self.pars.gain_mean
+        self.sensor.pixel_gain_std = self.pars.gain_std
+        self.sensor.pixel_gain_map = np.random.normal(
```
rknop, line 448: Not suggesting a change, but musing : are the pixel gains on a typical real CCD distributed normally? I suspect not exactly, and I also suspect they're not all uncorrelated.

improc/simulator.py
```diff
@@ -0,0 +1,886 @@
...
+        Generate an instance of a sky. This will usually stay the same
+        when taking a series of images at the same pointing.
+        We will assume that if the pointing changes, the sky changes
+        values like seeing and background.
+
+        """
+        self.sky = SimSky()
+        self.sky.background_mean = self.pars.background_mean
+        self.sky.background_std = self.pars.background_std
+        self.sky.background_instance = np.random.normal(self.sky.background_mean, self.sky.background_std)
```
rknop, line 496: Another thing worth doing at some point in the distant future is a variable sky background.  Of course, there's all the "wings of bright stars" stuff we talked about, but just to start some sort of parameterized polynomial.

rknop (commented):
Only one minor thing worth fixing -- the Poisson noise should be based on electrons, not ADU.

improc/simulator.py
guynir42: I think that might be a nice addition (as an optional parameter for the simulator). But really you can get a lot of the way there just by putting some elongated Gaussians as stand-ins for galaxies. 
The same is probably true for cosmic rays and bleeds. In terms of figuring out sky flat algorithms (and maybe other things) it may be close enough that it wouldn't matter if we hack something really simple together (like the very simple Gaussian PSF I am using).

improc/simulator.py
guynir42: Something like a moon illumination? I think that might actually be relevant to sky flats. I calculate the flats pixel by pixel, but I also start by normalizing each image based on the median sky brightness, so it could make a difference in principle. I will have to think about that.

improc/simulator.py
guynir42: You are right! I knew it would be useful to have someone else go over the actual simulation code :)

rknop (approved):

PR #103, opened 2023-10-27
ZOGY algoithm

guynir42:


tests/improc/test_zogy.py
```diff
@@ -0,0 +1,436 @@
...
+        for i, f in enumerate(fluxes):
+            x = y = int(pos[i])
+            c = S[y-edge:y+edge, x-edge:x+edge]
+            expected[j, i] = f * np.sqrt(np.sum(P ** 2) / B)
+            expected[j, i] *= np.sqrt(B / (B + f * np.sum(P ** 2)))  # add the Poisson noise from the source
+            measured[j, i] = np.nanmax(c)
+
+    mean = np.nanmean(measured, axis=0)
+    err = np.nanstd(measured, axis=0)
+    exp = expected[0, :]
```
rknop, line 226: Why are you comparing the means of the measured values to the first expected value?  If the background_instance is different in each expected value, shouldn't you compare each measured to each expected, or the mean of measured to the mean of expected?

tests/improc/test_zogy.py
```diff
@@ -0,0 +1,436 @@
...
+    # verify no stars are on top of our injection positions
+    # ZOGY still detects the new source on top of the star, but the S/N is reduced
+    for p in pos:
+        for i in range(len(sim.stars.star_mean_fluxes)):
+            if abs(p - sim.stars.star_mean_x_pos[i]) < 10:
+                sim.stars.star_mean_x_pos[i] += 15  # bump this star out of the way
+            if abs(p - sim.stars.star_mean_y_pos[i]) < 10:
+                sim.stars.star_mean_y_pos[i] += 15  # bump this star out of the way
+
+    # beware the 4-loop!
```
rknop, line 275: Fortunately, 3^4 is only 81

tests/improc/test_zogy.py
```diff
@@ -0,0 +1,436 @@
...
+                    sim.stars.remove_stars(len(sim.stars.star_mean_fluxes) - sim.pars.star_number)
+
+                    sim.make_image(new_sky=True)
+                    truth1 = sim.truth
+                    im1 = sim.apply_bias_correction(sim.image)
+                    im1 -= truth1.background_instance
+                    psf1 = truth1.psf_downsampled
+                    bkg1 = truth1.total_bkg_var
+
+                    # add a few sources
```
rknop, line 295: ...and 3^5 is only 243

tests/improc/test_zogy.py
```diff
@@ -0,0 +1,436 @@
...
+                    P = zogy_psf
+
+                    for i, f in enumerate(fluxes):
+                        x = y = int(pos[i])
+                        c = S[y - edge:y + edge, x - edge:x + edge]
+                        expected = f * np.sqrt(np.sum(P ** 2) / B)
+                        expected *= np.sqrt(B / (B + f * np.sum(P ** 2)))  # add the Poisson noise from the source
+                        measured = np.nanmax(c)
+                        # print((expected - measured) / expected)
+                        # TODO: figure out why we still have cases where the S/N is reduced by 50%
```
rknop, line 330: What does the histogram of S/N expected vs. measured look like?

tests/improc/test_zogy.py
```diff
@@ -0,0 +1,436 @@
...
+        star_number=num_stars,
+    )
+
+    jitter_values = [0.0, 0.1, 0.25, 0.5]
+    fluxes = [300, 500, 1000]
+    pos = np.linspace(0, imsize, len(fluxes) + 1, endpoint=False)[1:]
+
+    sim.make_stars()
+    # verify no stars are on top of our injection positions
+    # ZOGY still detects the new source on top of the star, but the S/N is reduced
```
rknop, line 368: ...of course, we're eventually going to want to test exactly that!  Supernovae are often found near the centers of galaxies.

rknop (approved):
This looks good.   I haven't dug into the details of the algorithm to make sure it actually matches what's in the ZOGY paper.

improc/zogy.py
```diff
@@ -0,0 +1,286 @@
...
+    any peak found in the score image has a value that is equivalent to the S/N of that detection.
+
+    Besides the classical score image (which is optimal assuming background dominated noise), the function also
+    returns the corrected score image, which takes into account the noise from bright sources.
+
+    The function also applies the astrometric noise correction when calculating score_corr, but only if
+    dx and dy are given (and if they are non-zero).
+    These values are... TODO: finish this
+
+    The last two outputs (alpha, alpha_std) represent a PSF-photometry measurement of the
```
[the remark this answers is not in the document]
guynir42: This is a good question. I think the math behind the algorithm will break down, but that doesn't mean it will necessarily fail or perform worse than deconvolutional methods.

improc/zogy.py
```diff
@@ -0,0 +1,286 @@
...
+
+    Parameters
+    ----------
+    image_ref : numpy.ndarray
+        The reference image, background subtracted, in units of electrons (gain corrected).
+    image_new : numpy.ndarray
+        The new image, background subtracted, in units of electrons (gain corrected).
+    psf_ref : numpy.ndarray
+        The PSF of the reference image. Must be normalized to have unit sum. Will be zero-padded to size of images.
+    psf_new : numpy.ndarray
```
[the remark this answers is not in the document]
guynir42: I'm thinking about cutting the image into subsections and stitching it back together (either interpolating between sections or choosing the max).

improc/zogy.py
```diff
@@ -0,0 +1,286 @@
...
+    Pn = pad_to_shape(psf_new, N.shape)
+    Pr = pad_to_shape(psf_ref, R.shape)
+
+    # make sure all masked pixels in one image are masked in the other
+    nan_mask = np.isnan(R) | np.isnan(N)
+    if np.sum(~nan_mask) == 0:
+        raise ValueError("All pixels are masked or no overlap between images.")
+
+    # must replace all NaNs with zeros, otherwise the FFT will be all NaNs
+    N[nan_mask] = 0
```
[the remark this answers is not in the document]
guynir42: Yes, NaNs are a problem. I am leaving the interpolation to the stage before the subtraction, but having even one NaN in the FFT is catastrophic so I must avoid it as a precaution.

improc/zogy.py
```diff
@@ -0,0 +1,286 @@
...
+        )
+
+    if any(s < a for s, a in zip(shape, arr.shape)):
+        raise ValueError(
+            f"The shape must be larger/equal to the array shape on all dimensions. Got {shape} and {arr.shape}."
+        )
+
+    new_arr = np.zeros(shape, dtype=arr.dtype)
+    if value != 0:
+        new_arr[:] = value
```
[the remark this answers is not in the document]
guynir42: nice, thank you!

tests/improc/test_zogy.py
```diff
@@ -0,0 +1,436 @@
...
+        S = np.ones(zogy_score_corr.shape) * np.nan
+        S[edge:-edge, edge:-edge] = zogy_score_corr[edge:-edge, edge:-edge].copy()
+        B = truth2.background_instance
+        P = zogy_psf
+
+        for i, f in enumerate(fluxes):
+            x = y = int(pos[i])
+            c = S[y-edge:y+edge, x-edge:x+edge]
+            expected[j, i] = f * np.sqrt(np.sum(P ** 2) / B)
+            expected[j, i] *= np.sqrt(B / (B + f * np.sum(P ** 2)))  # add the Poisson noise from the source
```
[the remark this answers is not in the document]
guynir42: You are right. I was playing around with the formulas to get the right behavior including the source noise. 
In a sense, the first line is what I expected and then the next line is a correction, but there's no reason to split them up.

tests/improc/test_zogy.py
guynir42: The expected is calculated and is the same for each `j` (each iteration), and then we average the measured (which is different for each iteration) but choose only the first of each expected. I tried to remove the superfluous calculation but it made the code more complicated because of the order of the loops I am using.

guynir42: In this case the background_instance is the same all the time, but if we ever change that, we will need to use an average. I will put the expected values as a mean to make it more robust.

tests/improc/test_zogy.py
guynir42: It is more of a warning to myself not to get to generous with the list of parameters to test.

tests/improc/test_zogy.py
guynir42: Yes, I'm going to add galaxies in one of the next PRs and we can see if it still works. But you lose S/N because of all the source noise from the galaxy. Mostly it makes the tests less reliable if you put random stars with random intensity under the transient.
Maybe I should run a test where I put a galaxy with known flux under the transient and see if I can figure out the correct S/N.

tests/improc/test_zogy.py
guynir42: zogy_snr_histograms

guynir42: The distributions look ok, but there's definitely a mismatch of 15% loss of S/N for fluxes above the background. 
I will try to figure this out, but I think we should merge and keep at it on another PR.

PR #106, opened 2023-10-31
PSF & complete Sextractor extraction

rknop:


pipeline/detection.py
```diff
@@ -74,7 +167,16 @@ def run(self, *args, **kwargs):
...
                     )
 
-                detections = self.extract_sources(image)
+                # TODO -- should probably pass **kwargs along to extract_sources
+                #  in any event, need a way of passing parameters
+                if self.pars.method == 'sep':
+                    detections = self.extract_sources_sep(image)
+                elif self.pars.method == 'sextractor':
+                    psffile = None if self.pars.psf is None else self.pars.psf.get_fullpath()
+                    detections, _ = self.extract_sources_sextractor( image, psffile=FIX_THIS )
```
guynir42, line 176: Not sure what is going on. Why not just put `psffile` into the function?

pipeline/detection.py
```diff
@@ -98,20 +201,422 @@ def run(self, *args, **kwargs):
...
                         raise ValueError('Provenance mismatch for sources and provenance!')
+                psf.image_id = image.id
+                if psf.provenance is None:
+                    psf.provenance = prov
+                else:
+                    if psf.provenance.id != prov.id:
+                        raise ValueError('Provenance mismatch for pfs and provenance!')
 
             ds.sources = sources
+            ds.psf = psf
```
guynir42, line 219: Now I'm happy we use a data_store to pass around all the different objects.

pipeline/detection.py
```diff
@@ -98,20 +201,422 @@ def run(self, *args, **kwargs):
...
+        apers = ( np.array( self.pars.apers ) if self.pars.apers is not None
+                  else np.array( [1., 2., 3., 4., 5., 7., 10.] ) )
+
+        if self.pars.measure_psf or ( self.pars.psf is None ):
+            # Run sextractor once without a psf to get objects from
+            # which to build the psf.
+            #
+            # As for the aperture, if the units are FWHM, then we don't
+            # know one yet, so guess that it's 2".  This doesn't really
+            # matter that much, becauser we're not going to save these
```
guynir42, line 262:
```suggestion
            # matter that much, because we're not going to save these
```

pipeline/detection.py
```diff
@@ -98,20 +201,422 @@ def run(self, *args, **kwargs):
...
+            File that has the PSF to use for PSF photometry.  If None,
+            won't do psf photometry.
+
+          tempname: str
+            A filename base for where the catalog will be written.  The
+            source file will be written to
+            "{FileOnDiskMixin.temp_path}/{tempname}.sources.fits".  (A
+            temporary image, weight, and mask file will be written with
+            the same name base to the same directory, but will be
+            deleted by this routine unless do_not_cleanup is False.)  It
```
guynir42, line 331:
```suggestion
            deleted by this routine unless do_not_cleanup is True.)  It
```

pipeline/detection.py
```diff
@@ -98,20 +201,422 @@ def run(self, *args, **kwargs):
...
+            is the responsibility of the calling routine to delete this
+            temporary file.
```
guynir42, line 333: It is a little bit confusing. Can you emphasize that there's a single file that must be written out and deleted by external function, and besides that there are also these additional temporary files that may or may not be deleted by this function.

pipeline/detection.py
```diff
@@ -98,20 +201,422 @@ def run(self, *args, **kwargs):
...
+
+          do_not_cleanup: bool, default False
+            This routine writes some temp files with the image, weight,
+            mask, and sourcelist data in it, muchof which is probably
```
guynir42, line 337:
```suggestion
            mask, and sourcelist data in it, much of which is probably
```

pipeline/detection.py
```diff
@@ -98,20 +201,422 @@ def run(self, *args, **kwargs):
...
+            raise FileNotFoundError( "Can't figure out where astromatic config directory is" )
+        if not astromatic_dir.is_dir():
+            raise FileNotFoundError( f"Astromatic config dir {str(astromatic_dir)} doesn't exist "
+                                     f"or isn't a directory." )
+
+        # TODO : make these configurable by instrument (at least!)
+        # For now we're using the astromatic defaults that everybody
+        # just uses....
+        conv = astromatic_dir / "default.conv"
+        nnw = astromatic_dir / "default.nnw"
```
guynir42, line 386: We should think about whether we want to config the pipeline to have instrument-specific behavior, or just assume you'd always run it on a single instrument, then you can just choose all of these parameters in the one config file for the whole system.

pipeline/detection.py
```diff
@@ -98,20 +201,422 @@ def run(self, *args, **kwargs):
...
+        else:
+            paramfile = tmpparams
+            with open( paramfilebase ) as ifp:
+                params = [ line.strip() for line in ifp.readlines() ]
+            for i in range(len(params)):
+                if params[i] in [ "FLUX_APER", "FLUXERR_APER" ]:
+                    params[i] = f"{params[i]}({len(apers)})"
+            with open( paramfile, "w") as ofp:
+                for param in params:
+                    ofp.write( f"{param}\n" )
```
guynir42, line 416: So much clunkiness I can't take it anymore :exploding_head:

pipeline/detection.py
```diff
@@ -98,20 +201,422 @@ def run(self, *args, **kwargs):
...
+                tmpparams.unlink( missing_ok=True )
+                if tempname is None: tmpsources.unlink( missing_ok=True )
+
+
+    def _run_psfex( self, tempname, image_id, psf_size=None, do_not_cleanup=False ):
+        """Create a PSF from a SExtractor catalog file.
+
+        Will run psfex twice, to make sure it has the right data size.
+        The first pass, it will use a resampled PSF data array size of
+        psf_size in x and y (or 25, of psf_size is None).  The second
```
guynir42, line 512:
```suggestion
        psf_size in x and y (or 25, if psf_size is None).  The second
```

guynir42 (approved):
This looks good. This is a really big block you've gotten us over. 
I have some small comments and questions. If you feel like you addressed all of them, feel free to merge.

tests/models/test_source_list.py
```diff
@@ -95,65 +95,65 @@ def test_source_list_bitflag(sources, demo_image, provenance_base, provenance_ex
         assert sources.id not in [s.id for s in sources5x]
 
 def test_read_sextractor( example_source_list ):
```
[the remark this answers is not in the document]
rknop: Yeah, makes sense.

models/psf.py
```diff
@@ -0,0 +1,479 @@
...
+
+    @format.inplace.setter
+    def format( self, value ):
+        self._format = PSFFormatConverter.convert( value )
+
+    image_id = sa.Column(
+        sa.ForeignKey( 'images.id', ondelete='CASCADE', name='psfs_image_id_fkey' ),
+        nullable=False,
+        index=True,
+        unique=True,
```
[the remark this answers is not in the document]
rknop: Ah, right-- we want image_id and provenance to be unique together.

models/psf.py
```diff
@@ -0,0 +1,479 @@
...
+        -------
+           A 2d numpy array
+
+        """
+
+        if self.format != 'psfex':
+            raise NotImplementedError( "Only know how to get resampled psf for psfex PSF files" )
+
+        xc = int( np.floor(x + 0.5) )
+        yc = int( np.floor(y + 0.5) )
```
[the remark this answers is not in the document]
rknop: They were supposed to be; I had some x and y where I wanted xc and yc.

pipeline/detection.py
```diff
@@ -74,7 +167,16 @@ def run(self, *args, **kwargs):
                         f'Cannot find a subtraction image corresponding to the datastore inputs: {ds.get_inputs()}'
                     )
 
-                detections = self.extract_sources(image)
+                # TODO -- should probably pass **kwargs along to extract_sources
+                #  in any event, need a way of passing parameters
```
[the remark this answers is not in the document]
rknop: Maybe.  Not right now.  The one I can think of that we'd configure at runtime but not need to put in the object's Parameters are the memory size parameters that go into sextractor.  (Currently they're hardcoded.)  Other methods might have other things like that.  If they're method-specific things, we probably don't want to code them into the Parmeters.  In any event, perhaps thought required.

pipeline/detection.py
rknop: ....weird.  I don't know why the tests passed, there should have been an error about a file not found.

...AH.  Yeah. I see.  I need to take that parameter in extract_sources_sextractor, and actually use it.  It didn't thrown an error because it was just hidden and ignored in **kwargs

rknop: (This is what happens when you realize that a meeting is starting when you're right in the middle of typing a line of code.)

pipeline/detection.py
rknop: We almost certainly will need instrument-specific behavior.  I can easily imagine us ending up needing to put together multiple different instruments, even if we don't want to.  (There will be some program that, say, taktes data on DECam at the same time as LS4, and they want us to run it all through the pipeline to compare discovered candidates, etc.)

pipeline/detection.py
rknop: Think about it like this.

Whenever you use a library, you have to interface to that library.  If it's doing anything that's non-trivial, that interface may make decisions and assumptions that differ from what you would have.  This kind of thing happens with Pandas, for instance, where sometimes to import things you have to wrangle the data a bit.  In astronomy, I've seen various libraries take weight (inverse variance) images, inverse noise images, noise images, and variance images.  (SExtractor lets you tell it which one you're passing, but others just demand a certain thing.) In that case, you'd have to reformat your data, even if you're just making a library call and not running a subprocess.  All of this is a little bit clunky, but it's the price you pay for using libraries that don't do everything exactly the way you might have wanted to, or that don't do everything in a way that perfectly matches your current use case.

These external programs we're calling just use reading and writing temporary files, and subprocess.run(), as part of their interface.  It's a little clunkier than just a function call passing a whole bunch of parameters, but not *that* much, given that you might have had to adapt your data for what the function's parameters wanted anyway.

Right here, we could make this a little bit less clunky by not having a default parameters file at all, and just keeping a list of the parameters we want as part of the code, and writing and deleting the temporary parameters file every time.  I considered doing that, but decided not to, because for debugging purposes it was convenient to have a the base parameters file just sitting there ready to use.

models/psf.py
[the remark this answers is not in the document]
guynir42: Someday you'll start using pycharm and it will highlight stuff like unused variables for you...

pipeline/detection.py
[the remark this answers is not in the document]
guynir42: My hope is that we can call `run()` without having to add more parameters. 

And if some of the SExtractor configuration stays in their own config file (until someone wants to alter it for some reason) that's fine too.

pipeline/detection.py
guynir42: yes, but even then, would it not be easier to just have some machines running with config file A and other machines running config file B? 

If we are going to split off one CCD per computer / processor then I don't see that same processor running DECam one minute and then LSST on the next...

pipeline/detection.py
guynir42: I guess you are right. At least we don't have to build a robot arm to take floppy disks in and out of an old drive "because that's how the software runs" back in the early 90s...

rknop: ...of course, when we get to writing to tape on HPSS, there *will be* robot arms involed...

PR #114, opened 2023-11-12
Split up tests

guynir42:
I want the tests in each subfolder (currently, `models`, `pipeline` and `improc`) to run as separate workflows. 
Parallelizing the workflows will make our GA tests run faster.

### DIFF
--- a/.github/workflows/run-improc-tests.yml
+++ b/.github/workflows/run-improc-tests.yml
@@ -1,4 +1,4 @@
-name: Run Model Tests
+name: Run Image Processing Tests
 
 on:
   push:
@@ -59,4 +59,4 @@ jobs:
 
       - name: run test
         run: |
-          TEST_SUBFOLDER=models docker compose run runtests
+          TEST_SUBFOLDER=improc docker compose run runtests
--- a/.github/workflows/run-model-tests.yml
+++ b/.github/workflows/run-model-tests.yml
@@ -1,0 +1,62 @@
+name: Run Python Tests
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  tests:
+    name: run tests in docker image
+    runs-on: ubuntu-latest
+    env:
+      REGISTRY: ghcr.io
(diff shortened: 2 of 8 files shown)


rknop:
A couple of comments:

(1) You also need a util test, as right now those tests won't be run.

(2) There is an inefficiency here: each test has to complete the "bake" step separately.  As long as the image is already saved somewhere, that's about a 2-3 minute step.  (I spent a bunch of time a while back trying to figure out why it's even that long, but ended up decided we'd just live with it.)  If the Dockerfile has changed, then it will need to rebuild the image, and all of the steps will probably rebuild the image in parallel, meaning they'll all be spending time that isn't necessary.

If you look at the tests that were run here, for two of the three, the bake step was more than half of the time that the whole test took.  As such, there's not really all that much efficiency gained.  The longest test (the model test) took 14m; if we added the full "run tests" step runtime of the other tests (which may be too conservative, as there are some cached steps that wouldn't have to be redone), that would only raise the runtime to just under 20m.  It's possible in the future that both the "models" and "pipeline" tests will have long enough runtimes that this is worth it, but at the moment, I'm not convinced that this modest time gain is worth the increase in complexity.

guynir42:
I agree that there's a lot of overhead, but we are not paying for it, only github is. So from here on out, each test we add will have a 1/3 chance of being in one of the categories (1/4 if we include utils) that is not the slowest test. I'd say that is efficient for us (not for github!). 

The real motivation is that I think improc will have a lot of heavy tests (think about subtractions) and the pipeline may start to get big, too, if we have source extraction and calibrations. Sooner or later you will be happy we split this up. 

In terms of complexity, we only need to have a few workflow files, and never look at them again. Unless I'm somehow also interfering with the way you run tests locally (I don't think I am). 

Also, I will add utils right away.

rknop:
I guess the one thing I'd ask for is a new target in the docker-compose file "runalltests" that runs all of them at once.  But, I can do that in some future PR when it strikes me.

Mostly, right now, I just run specific tests while I'm developing, and I only run everything before making the PR.

rknop (approved):

PR #115, opened 2023-11-12
Cleanup exposure fixtures

guynir42:


rknop (approved):
This looks fine.

Echoing something I've commented in a couple of other places: it might be worth having an optional rng argument to make_new_exposure (et al), so that one could pass in a pre-seeded rng, allowing for deterministic tests on these demo exposures.

guynir42:
I am starting to warm up on the idea of having the option to pre-seed the rng. 
Maybe we should do that as a global parameter that somehow makes it into the tests? 
Or should we focus on making each part of the test suite random/pre-seeded separately? 
Maybe we should make an issue for this.

PR #116, opened 2023-11-12
Catalog tools

guynir42:
Moving all the tools related to catalogs into a separate file. This includes downloading catalogs, manipulating and matching lists of sources, and calculating the best filter (see the bandpass thing below). 

Some of the tools have to stay in the instrument or in the catalog excerpt classes. For example, the magnitude transformations between DECam and Gaia are going to stay in the `DECam` instrument even though they could theoretically also fit into the CatalogExcerpt or the catalog tools. 

Using a `Bandpass` class we simply define the lower/upper limits of the wavelength range of each filter (assuming all filters are top-hat filters). Then we can override the list of filters for instruments and for catalogs. 
Once these are defined we can find the filter with the biggest overlap and use that one.


guynir42:
We should consider if "most overlap" is the best filter, e.g., what happens if two filters have the same overlap (or nearly the same) but one has a much broader bandwidth? (this happens in Gaia pretty easily). Maybe add a penalty to the overlap based on the total bandwidth?

guynir42:
Also need to consider where this decision is made? Is it when you call the fetch catalog from astrometric/photometric calibration? I think it may be wise to wait for the photometric calibration code to be available and work off that.

models/instrument.py
```diff
@@ -979,8 +1061,9 @@ def get_gain_at_pixel( self, image, x, y, section_id=None ):
             return self.get_section( image.section_id ).gain
         elif section_id is not None:
             return self.get_section( section_id ).gain
-        elif sec.gain is not None:
-            return self.gain
+        # Can we remove this case? What is "sec" in this case??
```
rknop, line 1064: That must be vestigal code where merging didn't do the right thing.  I think sec was supposed to be the sensor section, but of course that's already hit in the previous elif.  I think these two lines can be deleted.

models/instrument.py
```diff
@@ -1156,6 +1239,33 @@ def get_short_filter_name(cls, filter):
...
+        """
+        Get a dictionary of filter name -> Bandpass object for a list of common filters.
+        The default Instrument just gives some generic filters and their bandpasses,
+        but subclasses should override (or update) this dictionary with their own
+        filters and bandpasses.
+        """
+
+        # we can probably do better than this, but I don't know if it makes any difference
+        # ref: https://en.wikipedia.org/wiki/Photometric_system
+        values = dict(
```
rknop, line 1253: We're probably always going to want to make this instrument-specific.  But, if we're going to have a default, it would probably be better to have something more modern than the Bessel filters here that are usually used with Vega-based magnitudes.  The LSST filters might be a good choice; see section 2.4.1 of:

https://www.lsst.org/sites/default/files/docs/sciencebook/SB_2.pdf

pipeline/astro_cal.py
```diff
@@ -151,6 +152,15 @@ def __init__(self, **kwargs):
 
     # ----------------------------------------------------------------------
 
+    # TODO: should we move all of these catalog fetches into a separate file?
+    #  probably need to when doing the photometric calibration
```
rknop, line 156: Yeah, I agree we should do this (commented on that in the photometric PR).

rknop: ...however, for DECam photometric calibration specifically, we won't actually use the Bandpass matching.  The transformations we have for all DECam filters are to the gaia Mag_G bandpass, parameterized by BP-RP, so the maximum filter overlap isn't really an issue there.  Specifically for Gaia, this probably makes sense, since Mag_G is so broad.  If we do something like a PanStarrs transformation, we may want to use the bandpass overlapping.

models/instrument.py
```diff
@@ -1197,37 +1196,170 @@ def fiducial_aperture( cls ):
...
+        Returns a copy of the input data structure, with only the
+        rows that are likely to be stars.
+        """
+        output = copy.deepcopy(catdata)
+
+        # color cut
+        gaiaminbp_rp = 0.5
+        gaiamaxbp_rp = 3.0
+        color = catdata['MAG_BP'] - catdata['MAG_RP']
+        dex = (color >= gaiaminbp_rp) & (color <= gaiamaxbp_rp)
```
rknop, line 1231: ...it's possible that we might potentially want to use different color cuts for different instruments.

For now I'd just document in the function docstring that it does this color cut, so that if somebody in the future has an instrument that needs a different color cut, they know they have to override this function in their instrument definition.

rknop (approved):
Just minor comments.  (Before we actually *use* the bandpass overlap detectors, we'll probably want to decide what get_filter_bandpasses should return; I think that returning both Bessel and LSST filters together isn't going to be the right thing to do, since any given use case is more likely to want one or the other, not both.)

models/instrument.py
```diff
@@ -1197,37 +1196,170 @@ def fiducial_aperture( cls ):
...
+            u=Bandpass(320, 400),
+            g=Bandpass(400, 552),
+            r=Bandpass(552, 691),
+            i=Bandpass(691, 818),
+            z=Bandpass(818, 922),
+            y=Bandpass(950, 1080),
+        )
+
+        values = lsst
+        values.update(wikipedia)  # TODO: should we remove these or add them as options as well?
```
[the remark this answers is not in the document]
guynir42: I will return only the LSST filters and leave the other filter set commented out in case someone wants it later.

models/instrument.py
guynir42: I just put those parameters as function arguments with defaults. I don't know why I didn't just do that to begin with.

tests/pipeline/test_astro_cal.py
```diff
@@ -31,13 +31,13 @@ def gaiadr3_excerpt( example_ds_with_sources_and_psf ):
         catexp = catexp.recursive_merge( session )
         catexp.delete_from_disk_and_database( session=session )
 
+
 def test_download_GaiaDR3():
```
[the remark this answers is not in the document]
guynir42: You are right. Moved the tests into a separate file.

PR #117, opened 2023-11-13
Photometric

rknop:
I made a change to extraction, to use the SPREAD_MODEL SExtractor parameter.  This is a much better star/galaxy separator than the CLASS_STAR neural network that it uses... at least on the one test DECam image.  (The documentation also claims it's better.)  Unfortunately, it's *slow*.  SExtractor goes from running in a few seconds to taking a minute to run.  I wanted to do this, because for photometric calibration we need to use *stars*, and so it's really helpful to not just throw out a whole bunch of stars because of weak star/galaxy separation.

I think we can handle an extra minute of overhead on each image.  I anticipate with everything that we do, we should expect each image's processing to take between 5 and 10 minutes, based on experience with the lensgrinder pipeline.  We'll want to run a whole bunch of processes at once.  (With lensgrinder, I usually have 4 nodes going, and each node has 60 processes, one process per chip of an exposure.)  However, it's irritating that this is so slow.  Another thing to investigate in the future is if we can get SExtractor to multithread.  (On NERSC, if we run 60 processes at once, we can give 4 CPUs to each process.)  (Or we can rewrite SExtractor from scratch to use GPUs.)


guynir42:
General comment before I start looking at the code:
This feels like it could easily be our bottleneck. I don't think we should be running much slower than real time (which is about 1 minute per image). And I definitely don't think we need to break up SExtractor into multiple processors/GPU. 

For now, we can live with it being slow, but we should definitely think about alternatives, and have a parameter that switches between the slow and fast option. 

Here are a couple of things we can consider down the line to make this faster:
1) what if we don't rely on our code to find stars, but instead just look for things that are close to the catalog's stars. If a source is matched in position (and rough flux estimate) to a star that we know is there, we have a pretty good prior to say it is a star. 
Maybe I'm missing something, but it seems to me a lot easier to identify star/galaxy this way. 

2) we can come up with our own galaxy / star separator and make sure it runs quickly (I don't know what SExtractor is doing that takes so long, which is part of the reason I don't trust it). One way can be just match-filtering with multiple widths and throwing away too-wide stars (and too-narrow as they can be cosmic rays). Another way is to train our own CNN, but we'd probably do that using labels for stars/galaxies that we find by comparing to catalogs (see option 1) in which case we should probably just use option 1. 

I'll put this in a separate issue.

models/zero_point.py
```diff
@@ -17,9 +18,19 @@ class ZeroPoint(Base, AutoIDMixin):
...
     )
 
+    # TODO : figure this out
+    # image = orm.relationship(
+    #     'Image',
+    #     secondary='source_lists',
+    #     primaryjoin='zero_points.c.source_list_id == source_lists.c.id',
+    #     secondaryjoin='source_lists.c.image_id == images.c.id',
+    #     single_parent=True
+    # )
```
guynir42, line 32: I think you can do this really easily with an `associationproxy`. 

https://docs.sqlalchemy.org/en/20/orm/extensions/associationproxy.html#scalar-relationships

models/zero_point.py
```diff
@@ -41,3 +52,65 @@ class ZeroPoint(Base, AutoIDMixin):
...
+        index=False,
+        doc="Uncertainty on zp"
+    )
+
+    aper_cor_apers = sa.Column(
+        sa.ARRAY( sa.REAL ),
+        nullable=True,
+        default=None,
+        index=False,
+        doc="Pixel radii of apertures whose aperture corrections are in aper_cor_aper."
```
guynir42, line 75: How about naming this `aper_cor_radii`?

guynir42: Also I think you meant to say the corrections are in `aper_cor` not `aper_cor_aper`.

models/zero_point.py
```diff
@@ -41,3 +52,65 @@ class ZeroPoint(Base, AutoIDMixin):
...
+          aper_cor: float
+             Defined so that mag = -2.5*log10(aperflux_adu) + zp + aper_cor
+
+        """
+
+        if self.aper_cor_apers is None:
+            raise ValueError( "No aperture corrections tabulated." )
+
+        for aprad, apcor in zip( self.aper_cor_apers, self.aper_cors ):
+            if np.fabs( rad - aprad ) <= 0.01:
```
guynir42, line 113: Is there a difference between `np.fabs` and just regular `abs`?

tests/pipeline/test_photo_cal.py
```diff
@@ -0,0 +1,64 @@
...
+            pyplot.close( fig )
+
+        # WORRY : zp + apercor (for the first aperture) is off from the
+        # aperture-specific zeropoint that the lensgrinder pipeline
+        # calculated for this image by 0.13 mags.  That was calibrated to
+        # either DECaPS or PanSTARRS (investigate this), and it's
+        # entirely possible that it's the lensgrinder zeropoint that is
+        # off.
+        assert ds.zp.zp == pytest.approx( 30.165, abs=0.001 )
+        assert ds.zp.dzp == pytest.approx( 1.38e-7, rel=0.01 )   # That number is absurd, but oh well
```
guynir42, line 60: Why is it absurd? because the error is way too small?

pipeline/photo_cal.py
```diff
@@ -28,12 +70,185 @@ class PhotCalibrator:
...
+            dec from x and y for the objects in sources.
+
+          catexp: CatalogExcerpt
+            Gaia DR3 catalog excerpt overlapping image
+
+          min_matches: int, default 20
+            Minimum number of matches between the two catalogs to return
+            a zeropoint.
+
+          match_size : float, default 1.
```
guynir42, line 102: Maybe call this `match_radius`?

pipeline/photo_cal.py
```diff
@@ -28,12 +70,185 @@ class PhotCalibrator:
...
+                continue
+            else:
+                # If there are more than one, just pick the first one in the list
+                # that matches.
+                # TODO : think about whether we can do better.
+                catdex.append( i )
+                sourcedex.append( w[0] )
+
+        _logger.debug( f"Matched {len(catdex)} stars between Gaia and the image source list" )
+
```
guynir42, line 181: We can probably improve this using `argmin` and then a check if it is inside the match radius. 

In the interest of "if it's not broken don't fix it" I'll make an issue to come back and improve this part later.

pipeline/photo_cal.py
```diff
@@ -28,12 +70,185 @@ class PhotCalibrator:
...
+        self.individual_cols = cols
+        self.individual_mags = catdata[ 'MAG_G' ]
+
+        wgood = ( ~np.isnan( zps ) ) & ( ~np.isnan( zpvars ) )
+        zps = zps[wgood]
+        zpvars = zpvars[wgood]
+        _logger.debug( f"{len(zps)} stars survived the not-NaN zeropoint check" )
+
+        zpval = np.sum( zps / zpvars ) / np.sum( 1. / zpvars )
+        dzpval = 1. / np.sum( 1. / (zpvars ) )
```
guynir42, line 236: If you have one really tiny error (by chance) it will dominate the value of `dzpval`. I think there must be a better way to calculate the uncertainty on `zpval`.

pipeline/photo_cal.py
```diff
@@ -28,12 +70,185 @@ class PhotCalibrator:
...
+
+        return zpval, dzpval
+
+    def run(self, *args, **kwargs):
+        """Perform the photometric calibration.
+
+        Gets the image sources, finds a catalog Excerpt (using code in
+        astro_cal.py), matches the two lists on RA/Dec,
```
guynir42, line 244: Consider (in a separate PR) to make a "catalog tools" class or something like that (maybe put all that code into `CatalogExcerpt` itself? 

We should take inventory of all the catalog related stuff, considering that we also want to add non-Gaia catalogs, and think of a good design. I think we should discuss this on the next meeting.

pipeline/photo_cal.py
```diff
@@ -45,37 +260,40 @@ def run(self, *args, **kwargs):
...
-            else:
-                if zp.provenance.id != prov.id:
-                    raise ValueError('Provenance mismatch for zp and provenance!')
+                raise ValueError( f'Cannot find a wcs for image {image.filepath}' )
+
+            # Need an AstroCalibrator to use its ability to fetch catalogs
+            self.astrometor = AstroCalibrator( cross_match_catalog=self.pars.cross_match_catalog,
+                                               max_catalog_mag=self.pars.max_catalog_mag,
+                                               mag_range_catalog=self.pars.mag_range_catalog,
+                                               min_catalog_stars=self.pars.min_catalog_stars )
```
guynir42, line 281: See my comment above regarding catalog tools.

guynir42 (approved):
Looks good. I'm really happy with how this is moving along. 

I have some ideas on how to refactor some of the code into a "catalog" module but we should discuss this on the next meeting. 

Otherwise I have some small suggestions, as usual.

models/instrument.py
```diff
@@ -1150,6 +1150,78 @@ def get_short_filter_name(cls, filter):
...
+        Parmaeters
+        ----------
+          filter: str
+            The short filter name of the magnitudes we want.
+
+        Returns
+        -------
+          numpy array
+
+        """
```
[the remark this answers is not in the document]
rknop: The problem is that it's going to be instrument specific.  So, if we put it in Catalog, there will be functions named after instruments; if we put it in Instrument, there will be functions named after catalogs.  Short term, at least, I expect us to have more instruments than catalogs (I expect at least DECam, PTF something, and of course LS4).  From the point of view of other people using the code, they're more likely to add an instrument than a catalog, so having all the stuff that they have to customize in (more or less) once place makes that more manageable.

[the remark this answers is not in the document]
guynir42: What things in the Catalog code depend on the instrument?

pipeline/photo_cal.py
guynir42: Agreed. Although I need to think about this weighted sum thing some more.

models/zero_point.py
rknop: I grew up on C, where "abs" was an integer operation, so you had to make sure to use "fabs" when working with floating point.  I *think* in numpy they're the same thing, but long years of habit have made me cautious.  (It's much like including a . on all float constants, even though in python 3 / is floating-point division, not integer-division, when the things you're dividing are floats.)

guynir42: ok, I just wanted to make sure I wasn't missing anything.

rknop:
I made some changes that might be worth your taking another look at.  In addition to the small stuff:

* I removed the SPREAD_MODEL calculation.  We can decide if we want to make that a parameter in the future.  For immediate use in photo calibration, I added the Gaia star/galaxy catalog statistic to what gets pulled down, and use that instead to decide what is a star.  That's much faster than using the SExtractor SPREAD_MODEL.

* I added a "RUN_SLOW_TESTS" environment variable for the slow tests, so they're not run by default.  But, then I added that enviornment variable to the docker-compose.yaml file, so that they will be run when you do "docker compose run runtests", which is what happens on GA.

* I marked a couple of the zogy tests skip.  I'm having them fail roughly a quarter of the time, even with them marked as flaky.  I noted in the skip comment that we could resolve this problem if we were able to pass in pre-seeded random number generators, making the tests deterministic.

* Although I didn't do it here, I agree that it makes sense to have some sort of catalog utilities module for doing things like downloading catalogs and matching RA/Dec to another list.  (It's ugly right now that the photometry class needs an instance of the astrometry class to get catalogs.)  We *could* just put all of that in CatalogExcerpt, but it probably makes more sense to have some sort of Catalog base class that defines the interface, and then we can implement a derived class for GaiaDR3 (and, presumably, others later).  Whether this should be its own module, or be in pipeline, or somewhere else, isn't immediately obvious to me.  This will still leave open the issue as to what is the right place to put the instrument / catalog magnitude transformations, because there be one transformation for each (instrument, catalog) pair.  (...there might also be catalog/catalog and instrument/instrument magnitude transformations...)  I'm inclined to leave it in Instrument, because that's where individuals are more likely to be adding custom things to support their own system.

guynir42:
I will have another look tomorrow morning. 
As for the question of the catalog vs. instrument transformations, we should definitely spend some more time thinking about it. 
One option that pops to mind is to have a file that does all the conversions between filter systems, and then each instrument has a filter system enum and each catalog has one (they can be the same set of enums for all I care) and they will both lookup the right function to call in that one file (with an entry function that has a lookup table) and then you get the right function. 

Ideally we would also be able to do transformations of individual filters based on the correct filter in the other system (like the thing I was working on with the filter overlap). I guess in Gaia/DECam you have some measured coefficients that only work for G + colors. But even that function has different coefficients for different filters on the DECam side. 

Anyway, I will think about this.

rknop:
That sounds too complicated to me.  You'll be adding a bunch of boilerplate
code for lookups and such to solve something that's more easily solved just
having functions in the derived Instrument objects.

Practically speaking, you're very unlikely to use exactly the same
transformations for two different instruments, even if nominally they're on
the same filter system.  (I base this on experience.)

tests/improc/test_zogy.py
```diff
@@ -269,6 +270,7 @@ def test_subtraction_snr_histograms(headless_plots):
     plt.savefig(os.path.join(plot_path, "zogy_snr_histograms.png"))
 
 
+@pytest.mark.skip( reason="This test frequently fails even with the flaky.  Can we use a random seed?" )
```
guynir42, line 273: I don't think we should set a constant seed, because this test should pass statistically. If we limit to one random seed it may blind us to the fact that most of the time it doesn't work. 

I am surprised this fails so often. I haven't seen any failures since I added the flaky mark. Maybe instead just put a higher `max_runs` like 5?

I'm generally worried, that I put pretty large margins on this test and it should pass most of the time, but it doesn't. 
I think skipping it altogether will lead us to forget about it completely, so I prefer to increase the `max_run` instead.

guynir42 (approved):
Looks good to me. 

I would prefer not to skip all the ZOGY tests, but instead increase the `max_run` parameter. 
I will need to figure out why there are so many failures there.

tests/improc/test_zogy.py
rknop: I'm going to leave it as skipped for here; probably we should create an issue to look into this.

PR #120, opened 2023-11-14
Image simulator upgrades

guynir42:
I'm now adding some extra features that could make our tests a little more realistic. 
Note that a lot of the "physics" here is not very realistic but it gives images that sort-of look like real astronomical images. 
I want to be able to test image processing tools against "something that is a little more spread out and non-symmetrical, like a galaxy". This is not to be used to test galaxy classification tools, really. 

Things I'm going to try to get to in this PR:
- [x] galaxies
- [x] satellite trails
- [x] cosmic ray tracks 
- [ ] cosmic ray worms (this may be left to another PR)
- [x] saturated stars with pixel bleeds


improc/simulator.py
```diff
@@ -453,6 +501,331 @@ def get_star_flux_values(self):
...
+
+    def make_galaxy_list(self, imsize):
+        """
+        Make a field of galaxies.
+        Uses the power law to draw random mean fluxes
+        and widths, and uniform orientations (both in inclination
+        and in position angle) and positions.
+        The input, imsize, is a tuple of (imsize_x, imsize_y).
+        """
+        rng = np.random.default_rng()
```
rknop, line 549: Here and elsewhere, it might be worth having an optional "rng" parameter, and using np.random.default_rng() if that parameter is none.  That would allow us to use a pre-seeded random number generator in order to make tests deterministic.

improc/simulator.py
```diff
@@ -453,6 +501,331 @@ def get_star_flux_values(self):
...
+        alpha_f = abs(self.galaxy_flux_power_law) - 1
+        self.galaxy_mean_fluxes = self.galaxy_min_flux / rng.power(alpha_f, self.galaxy_number)
+
+        self.galaxy_mean_x_pos = rng.uniform(-0.01, 1.01, self.galaxy_number) * imsize[1]
+        self.galaxy_mean_y_pos = rng.uniform(-0.01, 1.01, self.galaxy_number) * imsize[0]
+
+        alpha_w = abs(self.galaxy_flux_power_law) - 1
```
rknop, line 556: It's probably worth explicitly saying that self.galaxy_flux_power_law ought to be negative.  Perhaps you could even verify that.  (You could then also not put the abs in here, make the -1 into a +1, and multiply by rng.power( alpha_w, self.galaxy_number ) instead of divide.)

improc/simulator.py
```diff
@@ -453,6 +501,331 @@ def get_star_flux_values(self):
...
+
+        if sersic_flux_ratios is None:
+            sersic_flux_ratios = 10 ** rng.uniform(-2, -1, number)
+        elif isinstance(sersic_flux_ratios, (int, float)):
+            sersic_flux_ratios = np.full(number, sersic_flux_ratios)
+        self.galaxy_sersic_flux_ratios = np.append(self.galaxy_sersic_flux_ratios, sersic_flux_ratios)
+
+        if cos_is is None:
+            cos_is = rng.uniform(0.1, 1, number)
+        elif isinstance(cos_is, (int, float)):
```
rknop, line 687: Worth verifying that it's between 0.1 and 1?

What will happen if you pass 0?  I believe you'll get an infinitely thin disk, since the model is just a 2-d sersic profile, and doesn't have any sort of disk thickness built in.  (As such, it's probably worth verifying that cos_is isn't too small.)

improc/simulator.py
```diff
@@ -453,6 +501,331 @@ def get_star_flux_values(self):
...
+        if self.galaxy_position_std is not None:
+            x += np.random.normal(0, self.galaxy_position_std, len(x))
+
+        return x
+
+    def get_galaxy_y_values(self):
+        """
+        Return the positions of the galaxies (in pixel coordinates)
+        after possibly applying small astrometric shifts (e.g., scintillations)
+        to the mean galaxy positions.
```
rknop, line 727: Aside: another (more painful to implement) use of shifting positions is differential chromatic refraction.

improc/simulator.py
```diff
@@ -453,6 +501,331 @@ def get_star_flux_values(self):
...
+        # first make the bulge using a sersic profile
+        r0 = sersic_scale
+        bulge = np.exp(-7.67 * (r / r0) ** (1 / 4))
+        bulge *= sersic_flux / np.sum(bulge)
+
+        # now make the disk
+        r0 = exp_scale
+        disk = np.exp(-1.67 * np.sqrt((x / r0 / cos_i) ** 2 + (y / r0) ** 2))
+        disk *= exp_flux / np.sum(disk)
+        disk *= cos_i  # correct for the projection of the disk
```
rknop, line 821: I don't understand why this is here.  Ignoring the effects of dust, galaxies don't have directed radiation.  The total flux should be the same in any event-- there are the same number of stars radiating, no matter how the galaxy is oriented relative to us.  An inclined galaxy would just have a higher surface brightness.

improc/simulator.py
```diff
@@ -453,6 +501,331 @@ def get_star_flux_values(self):
...
+            The y position of the galaxy center inside the image.
+            If given as None, will position at the center of the image.
+        exp_scale: float, default 1.0
+            The length scale used to make the disk.
+        sersic_scale: float, default 1.0
+            The length scale used to make the bulge.
+        exp_flux: float, default 1.0
+            The total flux of the disk. The disk flux is modulated
+            by a factor cos_i, so the total flux of the disk,
+            if it were viewed face-on, is exp_flux, but we only
```
rknop, line 773: See below where you do this cos_i multiplication.

rknop (commented):
This looks good, but there is a cosine multiplication that I don't believe should be there.  (See comments.)

improc/simulator.py
guynir42: I was thinking about galaxies as optically thick objects, but I realize they are not. 
The reason I put in this cos_i term (after a lot of deliberation) is that if you put the same total flux for two galaxies, one edge-on and one face-on, the face-on one will have a much lower surface brightness, which I wanted to avoid. 

Now if you convince me that galaxies are total transparent to their own starlight, then indeed we should remove this term. But can we actually totally ignore the dust? I know this is getting really low in the weeds for such a simple simulator, but my intuition is that edge-on galaxies are not that much brighter than face on, but maybe I need to go look at more deep field images or something...

rknop:
Galaxies are not transparent to their own starlight, but a zero-th order
model of a spiral galaxy is a disk that has some thickness, with a plane of
dust in the middle.  (It is true that the scale height of the dust disk
tends to be less than the scale height of the star disk.)  So, you're only
ever seeing the "top" half of the galaxy anyway... and no matter how
inclined it is (as long as it's less than π/2), you're seeing the same top
half, it's just that it gets more compressed on the sky as it has a higher
inclination.

improc/simulator.py
```diff
@@ -802,29 +890,582 @@ def make_galaxy_image(
...
         x = (x0 - center_x) * np.cos(rotation) + (y0 - center_y) * np.sin(rotation)
         y = (y0 - center_y) * np.cos(rotation) - (x0 - center_x) * np.sin(rotation)
         r = np.sqrt(x ** 2 + y ** 2)
+        cls.runtime['transform'] += time.time() - t0
 
         # first make the bulge using a sersic profile
+        t0 = time.time()
         r0 = sersic_scale
-        bulge = np.exp(-7.67 * (r / r0) ** (1 / 4))
+        bulge = np.exp( -7.67 * np.power(r / r0, 1 / 4) )
```
rknop, line 907: Is there any benefit to doing np.power instead of **?  I was under the impression that numpy would do the same thing in either case.

improc/simulator.py
```diff
@@ -802,29 +890,582 @@ def make_galaxy_image(
...
+        cls.runtime['add'] += time.time() - t0
+
+        # add cutoff
+        if cutoff_radius is not None:
+            t0 = time.time()
+            cutoff = np.ones(galaxy_image.shape)  # inside the radius there is no change
+            # use the disk scale as the length scale for the cutoff
+            cutoff[r > cutoff_radius] = np.exp(-5.0 * (r[r > cutoff_radius] - cutoff_radius) / exp_scale)
+            galaxy_image *= cutoff
+            cls.runtime['cutoff'] += time.time() - t0
```
rknop, line 930: This is fine, no need to change it, but I use time.perf_counter() for this sort of thing -- that's what the time documentation suggests is best for this kind of thing.  I think unless you really want high resolution, it doesn't matter that much.

rknop (approved):
Looks good.

improc/simulator.py
guynir42: I think there is not difference. I was just trying both options to make sure it doesn't change the runtime. It doesn't.

improc/simulator.py
guynir42: Yes I will start using that.

PR #122, opened 2023-11-15
Image Alignment

rknop:
A preface to subtraction.  This warps one image to align with another, using scamp and swarp.

I moved one of the scamp functions from AstroCalibrator out to a file in improc, because I wanted to do basically exactly the same thing within the warping.


pipeline/alignment.py
```diff
@@ -0,0 +1,305 @@
...
+            str,
+            'Alignment method.  Currently only swarp is supported',
+            critical=True
+        )
+
+        self.enforce_no_new_attrs = True
+        self.override( kwargs )
+
+    def get_process_name(self):
+        return 'warp'
```
guynir42, line 36: Should this process name be "alignment"?

models/enums_and_bitflags.py
```diff
@@ -162,6 +162,7 @@ class ImageTypeConverter( EnumConverter ):
         13: 'TwiFlat',
         14: 'ComTwiFlat',
         15: 'Fringe',
+        16: 'Warped',
```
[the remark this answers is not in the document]
guynir42: Ok, I see why you need this. Is there a plan to delete these images after using them? Or are we going to save warped copies?

pipeline/data_store.py
```diff
@@ -24,6 +24,7 @@
     'detection': ['subtraction'],
     'cutting': ['detection'],
     'measurement': ['detection', 'photo_cal'],
+    'warp': [ 'photo_cal' ]
```
guynir42, line 27: I know dictionaries are not ordered but can we move this process between photo_cal and subtraction?

tests/models/test_image.py
```diff
@@ -1140,6 +1140,13 @@ def test_image_filename_conventions(demo_image, provenance_base):
 
         new_convention = '{ra_int:03d}/foo_{date}_{time}_{section_id_int:02d}_{filter}'
         cfg.set_value('storage.images.name_convention', new_convention)
+        # This next line was added because I changed image.save() so
+        # that it will use self.filepath if it is non-None.  (I needed
+        # this change in behavior ato be able to control filenames when
```
guynir42, line 1145:
```suggestion
        # this change in behavior to be able to control filenames when
```

guynir42 (approved):
Looks good!

models/enums_and_bitflags.py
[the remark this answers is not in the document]
rknop: I had originally figured we'd save them, but thinking about it, that may not be necessary.  We will save thumbnails off of the warped reference when saving cutouts, but I don't anticipate needing the full images often enough to make it worth saving them; we can just regenerate them.  (This does have implications for what subtractions should list as their reference image.)

pipeline/data_store.py
rknop: Yeah, will do.

Turns out dictionaries *are* ordered.  I was surprised to find that out, since I'd never treated them that way.

PR #125, opened 2023-11-17
Fix the pipeline end-to-end test

guynir42:
This test has been skipped due to some issue with the header.
I'm not sure if that issue is resolved or not, but to get to it I'm trying to sort out some other problems along the way.


guynir42:
I need to integrate this with better simulated images or just replace it with a decam example.

tests/pipeline/test_extraction.py
```diff
@@ -327,9 +325,9 @@ def test_run_detection_sextractor( decam_example_reduced_image_ds ):
     # wrong.
 
     assert ds.sources.psffluxadu()[0].min() == 0.0
-    assert ds.sources.psffluxadu()[0].max() == pytest.approx( 1726249.0, rel=1e-5 )
-    assert ds.sources.psffluxadu()[0].mean() == pytest.approx( 48067.805, rel=1e-5 )
-    assert ds.sources.psffluxadu()[0].std() == pytest.approx( 169444.77, rel=1e-5 )
+    assert ds.sources.psffluxadu()[0].max() == pytest.approx( 1716295.2, rel=1e-5 )
+    assert ds.sources.psffluxadu()[0].mean() == pytest.approx( 48076.523, rel=1e-5 )
+    assert ds.sources.psffluxadu()[0].std() == pytest.approx( 169310.296, rel=1e-5 )
```
guynir42, line 331: The values I get when I run this locally are a little bit different (the change is not small, but certainly outside 1e-5). 
I don't know why there would be differences (maybe sextractor versions?). 
Maybe we should make the `rel` a little bigger so it includes these variations, otherwise it would be good to figure out what causes the difference.

rknop: Weird... what version of sextractor are you running?  The one in the docker image is:

```
root@646f46408b26:/seechange# source-extractor --version
Source Extractor version 2.25.0 (2018-02-08)
```

I'm fine bumping the rel to 1e-4.  Bigger than that, I'd want to know why there were differences.

tests/pipeline/test_extraction.py
```diff
@@ -231,8 +229,8 @@ def test_run_psfex( decam_example_reduced_image_ds ):
         tmppsfxmlfile.unlink()
 
         psf = detector._run_psfex( tempname, sourcelist.image_id, psf_size=26 )
-        assert psf._header['PSFAXIS1'] == 29
-        assert psf._header['PSFAXIS1'] == 29
+        assert psf._header['PSFAXIS1'] == 27
+        assert psf._header['PSFAXIS1'] == 27
```
rknop, line 233: All of these will need to be reset when you remove the psf_sampling clamping to a minimum of 1 in detection.py.

tests/pipeline/test_extraction.py
rknop: Ah!  I bet I know the reason they were different -- it was the changes you made to psf_sampling.  When you revert that, these should go back to what they were before.

rknop (changes requested):

models/instrument.py
```diff
@@ -1914,8 +1914,8 @@ def download_and_commit_exposures( self, indexes=None, clobber=False, existing_o
         """Download exposures and load them into the database.
 
         Files will first be downloaded to FileOnDiskMixin.local_path
-        with the filename that the origin gave them.  The headers of
-        files will be used to construct Exposure objects.  When each
+        with the filename that the origin gave them. The headers of
+        files will be used to construct Exposure objects. When each
```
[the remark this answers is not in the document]
guynir42: I started to notice these double spaces everywhere and indeed you are putting them in on purpose. 
I will try to do the same, and certainly I will stop removing them!

PR #126, opened 2023-11-17
Docker file reads requirements file

guynir42:
Also could take some time to get rid of unnecessary requirements.


rknop (approved):
Looks good.

PR #129, opened 2023-11-20
Download PTF sample images

guynir42:


tests/conftest.py
```diff
@@ -856,3 +858,76 @@ def example_psfex_psf_files():
...
+    cachedpath = filename + '_cached'
+
+    if os.path.isfile(cachedpath):
+        _logger.info(f"{path} exists, not redownloading.")
+    else:
+        url = f'https://portal.nersc.gov/project/m2218/pipeline/dec+04/ccd_11/{filename}'
+        retry_download(url, cachedpath)  # make the cached copy
+
+    shutil.copy2(cachedpath, path)  # make a temporary copy to be deleted at end of test
+    prov = provenance_base
```
rknop, line 881: Is there enough code overlap between this and DECam that it'd be worth breaking some of it out into a function?  (I'm guessing probably not, the savings would just be a handful of lines.)

rknop (approved):
Looks good.

tests/conftest.py
guynir42: I have an idea for what to do to clean up the fixtures and the caching, but that will come later today.

PR #130, opened 2023-11-20
SmartSession to get multiple inputs

guynir42:
Instead of having to do `session = self.session if session is None else session` each time you want to call `SmartSession` with more than one possible input, simply do `SmartSession(session, self.session)` and it will pick the first one which is a session object.

### DIFF
--- a/pipeline/data_store.py
+++ b/pipeline/data_store.py
@@ -480,13 +480,13 @@ class DataStore:
             # database; do a quick check for mismatches.
             # (If all the ids are None, it'll match even if the actual
             # objects are wrong, but, oh well.)
-            if self.exposure_id is not None and self.section_id is not None:
-                if ( self.image.exposure_id != self.exposure_id or
-                     self.image.section_id != self.section_id ):
+            if (self.exposure_id is not None) and (self.section_id is not None):
+                if ( (self.image.exposure_id != self.exposure_id) or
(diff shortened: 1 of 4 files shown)


pipeline/data_store.py
```diff
@@ -484,13 +480,13 @@ def get_image(self, provenance=None, session=None):
             # database; do a quick check for mismatches.
             # (If all the ids are None, it'll match even if the actual
             # objects are wrong, but, oh well.)
-            if ( self.exposure_id is not None ) and ( self.section_id is not None ):
-                if ( ( self.image.exposure_id != self.exposure_id ) or
-                     ( self.image.section_id != self.section_id ) ):
+            if self.exposure_id is not None and self.section_id is not None:
```
rknop, line 483: I'm actually very fond of putting extra parentheses in if statements.  I realize that operator precedence is such that the "is not" will have higher presdence than "and", but putting in extra parentheses means not having to actually worry about operator precedence.  I think it makes the whole thing more readable as well; it clearly separates out each separate clause.

rknop (approved):
Good idea.

pipeline/data_store.py
guynir42: This is just a style choice so I will revert to what you were doing before. I don't particularly like it but I don't really care either way...

PR #133, opened 2023-11-21
Update the documentation

guynir42:
I've moved the documentation from README and expanded it quite a bit. 

There's still work to be done but this is already much more than what we had until now. 

I'm trying to make the "read the docs" project pick up the documentation automatically, I'm not sure it will work until we merge this.


.readthedocs.yaml
```diff
@@ -0,0 +1,32 @@
...
+  tools:
+    python: "3.12"
+    # You can also specify other tool versions:
+    # nodejs: "19"
+    # rust: "1.64"
+    # golang: "1.19"
+
+# Build documentation in the "docs/" directory with Sphinx
+sphinx:
+   configuration: docs/conf.py
```
rknop, line 20: I don't see the file docs/conf.py 

I haven't used sphinx or readthedocs before; does it happily ignore a missing conf.py file and just use defaults?  Do the defaults include building docs on the API?

rknop (approved):
It'll be nice to have readthedocs.

Only comment I have is on the missing conf.py file, which may just be me not understanding how the readthedocs yaml file and sphinx work.

.readthedocs.yaml
guynir42: I'm actually not sure (this is my first time too). 
I think we will not immediately use sphinx at all. If I understand correctly, sphinx is used to auto-generate docs. 
Right now I'm more worried about the docs I'm writing myself. If someone wants to see the signatures of any functions they can look at the code directly (with the fairly good docstrings in place). In the future we may want to use this but right now I will just comment it out and hope that the readthedocs works without sphinx.

guynir42, line 20:
```suggestion
# sphinx:
#    configuration: docs/conf.py
```

PR #139, opened 2023-11-23
Streamline upstream downstream

guynir42:
This is pretty big refactor but it will smooth along a lot of the handing around of data and also allow us to do image coaddition properly. 

It builds upon the previous PR https://github.com/c3-time-domain/SeeChange/pull/116 so it should only be reviewed after that part is merged and merged into this branch. 

The main thing is a new object ImageProducts which holds relationships to an Image but also to PSF, SourceList, WorldCoordinates, ZeroPoint (and maybe more if we need to add them later). 
This replaces the self-referencing table for source_images. 
Whenever you create an Image from other images (not from an Exposure) you must build ImageProducts for each of those images. A list of ImageProducts is saved as a relationship ("upstream_products") to the new Image. 
You can mark the ImageProduct that is the "reference" unto which all other images were aligned. 
You can mark the ImageProduct that is the "science" image, which means the image where you expect to see interesting things. 

In the case of a warped or difference image, there would be only two ImageProduct objects in the upstream_products list, one of which is the reference and the other is the science. 

In the case of a coadded image, the list is as long as the number of input images. The one that all others are warped into is marked as the reference. There would not be a "science" image in this case. 

The ReferenceEntry object will point to an ImageProducts object instead of an Image. We will also rename it to ReferenceProducts. It's goal remains to be a way to tag specific image (products) as THE reference for a given field, section_id and time validity range. 

Things that will change on this PR:
 - [x] Image no longer has direct relationships to ref_image and new_image. Instead it will refer to the ImageProduct list. 
 - [x] A ref_id and new_id will point to the chosen ImageProduct of the reference and new, and shortcuts to the underlying images can be added as "ref_image" and "new_image". 
 - [x] bitflag with it's crazy scheme of joining tables to get the most up to date bitflag from all upstream models is just not sustainable. I will try to add some code to make sure the bitflags are propagated but this will not happen without explicit user command, except when the objects are first generated in the pipeline. The new bitflag system is pushed into a mixin. 
 - [x] Image will have a few data products as optional attributes (psf, sources, wcs, zp) and those get filled in using the datastore, not as relationships, as it is not trivial to decide which e.g., source list goes with which image. These will have to be checked when creating new images based on other images (e.g., subtraction, coaddition) and then bundled into an ImageProduct. 
 - [x] Additional columns on the image that we were meaning to add a long time ago are added now. These include estimates for the background, fwhm, lim mag and zp.


rknop:
I don't like the whole idea behind this.  I think the image products table is a bad idea for a number of reasons, and do not think we should add it as a database table.

First, it introduces redundant information into the database, which undermines the whole database normalization thing.  WCS, zeropoint, source list already have a pointer back to the image that they are associated with; we don't need yet another table to associate them all. 

Second, this way of doing it is much less flexible than having the image pointers in WCS, zeropoint, source list, etc.  It's entirely possible that we'll have multiple source lists, or multiple WCSes, or (most likely) multiple zeropoints for one image.  That's fine; different ones have different provenances.  But, there's no provenance in this image products table.  Do we create a new image products table row every time there's a new provenance for any of the derived products?  If so, then we need a whole bunch of provenance columns in this table as well, and it make it more clear how redundant this is with information that's already in the database.  If this table doesn't have provenances, then then we have to join this table to all of the other tables while selecting in order to find the ones with the provenance we're looking for, at which point this table has introduced no advantage at all.  In either case, the number of rows in this table is going to grow every time we change the provenance of a downstream product.

From a runtime point of view we already have something that associates a source list, WCS, zeropoint with an image: DataStore.  And DataStore already handles provenance.

I do think it's worth thinking about having an image relationships table to store the source images for coadds and subtractions, to replace the source ids column that Image has right now.  But, as is, this image products table introduces complexity and raises new issues without giving us anything we need that we don't already have.

guynir42:
I'm usually not in favor of adding complexity without good reason. After a couple of restless nights I came up with this scheme and I will try to convince you there is no other way. 
- You cannot use the pointer in e.g., the WCS back to Image. There are many WCSs (with different provenances) pointing back to a certain Image. If you want to remember which WCS (and ZP, and source list and PSF) you used together to make a subtraction, you need a table that associates many-to-many Image to each of the other tables. 
- The data store doesn't get persisted. It is created as part of the pipeline, does its best to get the DB objects with provenances that match the parameters of the pipeline, and then fetches those objects from the DB. Without having the set of parameters that are requested at the exact moment the pipeline runs, you cannot do better than "get the latest provenance" and that is not always going to work. 
- Imagine you have just generated a reference image (with WCS and source list, etc). You want to save it on the References table, but you cannot save a pointer to the datastore that was used to make it. 
- Even if you go ahead and add pointers in the Reference table to each one of the specific products used (which is very close to an ImageProducts table) then you still don't have a solution for when you coadd images. Each coadded image that goes in to make the reference has its own products and you need to keep track of which exact objects came with each image. 
- There will be a new row of ImageProducts whenever you make a new downstream provenance, assuming you use the new products to produce some coadd or subtraction image from them. That's just a natural part of making versions of upstreams and downstreams. It is a many-to-many relationship(s) that you need to track in one or more tables. 
- Each one of the products is linked one-to-one to a specific ImageProducts object, so you can easily get the Provenances for each of the underlying objects, no need to keep copies of those.

rknop:
I still don't agree that this is a good idea.  It adds redundancy to the database; this means that we have more things that we have to keep manually synchronized because it's not built into the design of the database structure.

* If for whatever reason you want to get the WCS, ZP, Source List that was used to make a subtraction, you look at the upstreams in the subtraction provenance, and then you pull out the WCS, ZP, SourceList that have those provenances.  The main place I can see this being used is debugging, or tracking down issues.  In routine use, what we're going to do is ask for (say) all the subtractions of a given provenance, or all of the objects of a given provenance.

* The data store doesn't need to be persisted.  All of the linkages are already there in the foreign keys of the tables.

* You *do* have the set of parameters requested at the moment the pipeline runs; otherwise, you wouldn't be able to run the pipeline.  You're running the pipeline with some configuration, which may just be the default.  I do think we need to think about getting provenances to go with parameters (I'm not sure if we already have code to do that), but that's going to be an issue whether or not we have this image products table, because if you want to get something of a specific provenance, you have to know what that provenance is (including all the parameters that went into it).

* Re: saving a reference to the References table, why would you need to save a pointer to the datastore?  You just save the reference.  It will have all the information in provenance necessary to figure out where it all came from.  DataStore is not something that relates at all to the database (and doens't need to), it's a convenience tool for the code that keeps track of lazy loading and what things go with what at runtime, so the code doesn't have to do that everywhere it uses these things

* Specifically with respect to coadded images, I do think we should think about how we track the information.  But, that's a specific thing to coadded images.

Thinking about using this in practice: we start with an image.  We then run a number of steps:

  * source extraction and PSF creation
  * astrometric solution
  * zeropoint solution

Do we save a new row to the ImageProducts table every time we run one of these steps?  So, at the end, there is a row that has (Image + Source), another row that has (Image + Source + WCS), and another row that has (Image + Source + WCS + ZP)?  Or do we just keep adding an existing row?  If so, we then have to be careful that we're adding to the *right* row (i.e. the one that has the right provenances for the previous steps), which is going to be more complicated than just directly storing the linkages to the previous things in the step (that is, what we do right now).

If you really want to have this ImageProducts table, then we should also remove the image_id field from SourceList and PSF, and the source_list_id from WorldCoordinates and ZeroPoint.  Then, this ImageProducts table becomes the way of linking the various objects to each other.  But, I don't like that, as it's more complicated than the direct linkage we have right now.  It doesn't just store the individual linkages, but is a single table that captures a bunch of different linkages, which is really quite messy in comparison to keeping track of the linkages individually.  And, I still don't think that this adds anything that we don't have already.

rknop:
Another thing we might think about if you really want this table: remove provenance from everything except for Image.  Remove image_id and sourcelist_id from the various other tables.  Add a provenance line (that encapsulates all the provenances for all of the steps) to the ImageProducts table.  The ImageProducts table then becomes the place to track linkages between images and other things, and the provenances of those linkages; this avoids us having redundant information in the database.  This does represent a fairly big change in the design.  Among other things, you lose the notion of picking up step n+1 with a new provenance for just that step using a pre-existing provenance for everything through step n.  (You might still be able to save the redoing of steps by looking at other entries in the ImageProducts table to see if what you need is already there somewhere.)

What I'm really opposed to is storing the linkages in two different places: foreign keys in the individual tables, as well as a big table that links them all together (the ImageProducts table).  From a scalability point of view, it's more flexible to have the linkages spread out rather than all in one place.  It's easier to add new methods on individual steps if you just have to think about that one step rather than a monolithic object that encapsulates all of the steps.  On the other hand, it might simplify dealing with provenances if they're all in one place, rather than having separate provenances and upstreams on every table.  I do strongly prefer the linkages being dispersed rather than collected all in one place.

guynir42:
I think I can figure out a way to do this without an extra table using just the provenances and the direct links. Give me another restless night to figure this out and I will ask you for a real review when I'm done refactoring it.

rknop:
Consider using DataStore for this -- it already does a bunch of what we want to do here.

rknop:
How about this : refactor the image_source_self_association_table to have columns:
   combined_id
   source_wcs_id
   source_zp_id

instead of a source_image_id column.

It's a little ugly because there's redundant stuff here, but since swarp irritatingly requires a MAG column in the source lists it matches, right now the alignment requests zeropoints for the images to be warped.  The worldcoordinates linked via wcs_id has a pointer back to the source list, which in turn has a pointer back to the image, so it gives you everything you need except the zeropoint.  The zeropoint also has a pointer back to the source list and image.  That should give you everything you need to know about the images and associated data that went into a coadd.

Perhaps we should have zeropoint depend on wcs and point back to it rather than to source list, then this association table would only need the zp.

guynir42:
Just because WCS points back to a source list doesn't mean that it is the ONLY WCS that points back to that source list. 

One of the first things I did was add columns to the source image association table, but then I realized it will be easier to keep one table to bundle up all the image products than to push that data on top of an association table. 

But I believe this can be done using provenance, although not certain if it will be simpler, and we will have to make at least an assumption that all coadded images in a stack have the same provenance and their products have the same provenance.

guynir42:
I've managed to get the main functionality without an additional table. It is simpler this way!

rknop:
> Just because WCS points back to a source list doesn't mean that it is the ONLY WCS that points back to that source list.

For the use case, though, if you are doing coadds, and you point back to a wcs instead of to the image or source list, you have now uniquely specified the wcs, source list, and image that went into that coadd.

guynir42:
OK @rknop the tests are green, I think you can have a look at the code.

models/image.py
```diff
@@ -853,11 +647,14 @@ def from_images(cls, images):
         if len(images) < 1:
             raise ValueError("Must provide at least one image to combine.")
 
+        # sort images by mjd:
+        images = sorted(images, key=lambda x: x.mjd)
+
         output = Image(nofile=True)
 
         # for each attribute, check that all the images have the same value
         for att in ['section_id', 'instrument', 'telescope', 'type', 'filter', 'project', 'target']:
```
rknop, line 656: I wouldn't require the project and target to be the same.  Target is probably going to be parsed from the FITS header, and there are reasons why it might vary.  Sometimes people type differently (spaces, no spaces, capitalization), and sometimes they just get it wrong.

It's also entirely possible that we will want to coadd images from different projects in order to make a deep reference.  I've certainly done this with the NOIRLab archive.

Finally, we may well need to coadd images with different section_ids.  Even with LS4, the same field will show up on different chips sometimes, because the quartered filter on the focal plane has two i band quadrants, so sometimes a given field will be in the top half of the detector, sometimes in the bottom half.  But, in general, surveys will not always go back to exactly the same fields, so you need to build references as mosaics of mulitple other images.

models/image.py
```diff
@@ -877,23 +674,35 @@ def from_images(cls, images):
         output.exp_time = sum([image.exp_time for image in images])
 
         # start MJD and end MJD
-        output.mjd = min([image.mjd for image in images])
-        output.end_mjd = max([image.end_mjd for image in images])
+        output.mjd = images[0].mjd  # assume sorted by start of exposures
+        output.end_mjd = max([image.end_mjd for image in images])  # exposure ends are not necessarily sorted
```
rknop, line 678: If the exposure starts are sorted, then the exposure ends should be sorted too.  The only way for them to mismatch is if exposure B started while exposure A was still going, and was a short exposure that finished first.  But that can't happen of all of the images are from the same instrument.

We should add to the docstring that the images need to be sorted by mjd if the code is going to assume that.

models/image.py
```diff
@@ -877,23 +674,35 @@ def from_images(cls, images):
...
 
-        output.source_images = images
         base_type = images[0].type
         if not base_type.startswith('Com'):
             output.type = 'Com' + base_type
 
+        output.upstream_images = images
+
+        # mark the first image as the reference (this can be overriden later when actually warping into one image)
+        output.ref_image_id = output.upstream_images[0].id
```
rknop, line 691: See comments above re: ref_image_id and new_image_id.

What I would do here is add an optional argument "coadd_coord_ref" or some such that stays which image is the coordinate reference.  Then, this code would add entires into the image assocation table with assocation_type of coadd_member.  If coadd_coord_ref is not None, then that one image will have its assocation type be coadd_member_and_coord_ref (or whatever).

Another note: because you have a specific function for making a combined image from a new and ref, I'd put in the docstring that this is not for constructing the data strutures assocated with a subtraction.  It's really only in the case where all of the images are (mostly) filling the same role.  (Can you think of anything other than a coadd where we'd call this?  If we median the images, I'd still call that a "coadd".)

models/image.py
```diff
@@ -905,44 +714,70 @@ def from_ref_and_new(cls, ref, new):
...
+        if new_image is None:
+            raise ValueError("Must provide a new image.")
+
         output = Image(nofile=True)
 
         # for each attribute, check the two images have the same value
-        for att in ['section_id', 'instrument', 'telescope', 'filter', 'project', 'target']:
-            ref_value = getattr(ref, att)
-            new_value = getattr(new, att)
+        for att in ['instrument', 'telescope', 'project', 'section_id', 'filter', 'target']:
```
rknop, line 739: See comment above in from_images -- I wouldn't require project, section_id, and target to be the same.  (Requiring that, for instance, would mean we couldn't do what I did with the lensgrinder pipeline, where the references came from a deep survey previuosly performed with DECam.)

models/image.py
```diff
@@ -905,44 +714,70 @@ def from_ref_and_new(cls, ref, new):
...
+            ref_value = getattr(ref_image, att)
+            new_value = getattr(new_image, att)
 
             if att == 'section_id':
                 ref_value = str(ref_value)
                 new_value = str(new_value)
-            if ref_value != new_value:
+
+            if att in ['section_id', 'filter', 'target'] and ref_value != new_value:
```
rknop, line 747: OIC - you're only requiring section, filter, and target to be the same.  But, as I noted, we can't require section to be the same.

models/image.py
```diff
@@ -905,44 +714,70 @@ def from_ref_and_new(cls, ref, new):
...
                 )
-            output.__setattr__(att, new_value)
+
+            # assign the values from the new image
+            setattr(output, att, new_value)
         # TODO: should RA and Dec also be exactly the same??
 
+        output.upstream_images = [ref_image, new_image]
+        output.ref_image = ref_image
+        output.ref_image_id = ref_image.id
```
rknop, line 759: Doesn't setting output.ref_image also set the ref_image.id?

But also, note my comments above where I suggest removing the ref_image and new_image fields from Image, and putting the nature of the assocation in the image assocation table.

models/image.py
```diff
@@ -905,44 +714,70 @@ def from_ref_and_new(cls, ref, new):
...
-        output.ref_image = ref
-        output.new_image = new
-        output.type = 'Diff'
-        if new.type.startswith('Com'):
+        if im_type is not None:
+            if im_type.lower == 'diff':
+                output.type = 'Diff'
+            elif im_type.lower == 'warped':
+                output.type = 'Warped'
+            else:
```
rknop, line 778: Why would it ever be warped?  This is just for making subtractions, isn't it?

models/image.py
```diff
@@ -1242,6 +1077,244 @@ def load(self):
...
+        from models.zero_point import ZeroPoint
+
+        with SmartSession(session) as session:
+            im_ids = [im.id for im in self.upstream_images]
+
+            # get all the products for all images at once
+            sources_result = session.scalars(
+                sa.select(SourceList).where(
+                    SourceList.image_id.in_(im_ids),
+                    SourceList.provenance_id.in_(prov_ids),
```
rknop, line 1165: This is going to have trouble again....  It's possible that the various provenances that went into the reference image for a subtraction will not match the current provenances, as I note elsewhere.  That means the source list won't have the same provenance (because of upstreams), even if it was run using the same source list code.

In fact, this will have trouble in a subtraction even if nothing in the code has changed.  The upstream ref is probably a coadd, but the new could well be a single image, so their provenances won't match (and thus the provenances of things like source lists that come out of them won't match).

models/image.py
```diff
@@ -1242,6 +1077,244 @@ def load(self):
...
+            raise ValueError("Cannot combine PSFs with different provenance IDs.")
+        if len(set([p.id for p in wcs_provs])) > 1:
+            raise ValueError("Cannot combine WCSs with different provenance IDs.")
+        if len(set([p.id for p in zp_provs])) > 1:
+            raise ValueError("Cannot combine ZPs with different provenance IDs.")
+
+        return image_provs + sources_provs + psf_provs + wcs_provs + zp_provs
+
+    def load_upstream_products(self, session=None):
+        """Make sure each upstream image has its related products loaded.
```
rknop, line 1127: Is the use case for this e.g. when you're making a coadd or a subtraction, and you want to have all the stuff for all the things that go into it?  I can see using it for that, but I'm not sure what else I'd use it for.

models/references.py
```diff
@@ -1,29 +1,46 @@
...
+from models.image import Image
+from models.provenance import Provenance
+from models.source_list import SourceList
+from models.psf import PSF
+from models.world_coordinates import WorldCoordinates
+from models.zero_point import ZeroPoint
 
 
-class ReferenceEntry(Base, AutoIDMixin):
+class Reference(Base, AutoIDMixin):
```
rknop, line 13: When I was making comments elsewhere, I realized that with our provenance system we need more information here.

To see the problem, consider a subtraction.  A common use case will be a coadded reference, but a single new image.  So, the reference and the new image will *not* have the same provenance.  That means when loading the reference's source list (or wcs, or zp) you can't just say "give me the one that's the same as the one I'm using for the new image", because they will be different due to the reference and new image having different upstream provenances.  So, we don't know which provenance is the right one to ask for to get the reference image's source list (etc.).

The database normalization way to solve this is to first make zeropoint have an upstream of wcs (rather than soruce list).  Then, in this table, instead of image_id, you have zeropoint_id.  That will then uniqely define a (zeropoint, wcs, source list, image) that you use as a reference.  For convenience, I'd put in image, wcs, and source_list methods that crawl the join chain to get the right object out.

The scary way to solve this is to have four fields here: image_id, source_list_id, wcs_id, and zp_id.  Of course, then the database structure itself isn't enforcing consistency the way it does if we just link to a zeropoint.

models/image.py
```diff
@@ -1242,6 +1077,244 @@ def load(self):
...
+            a list of unique provenances, one for each data type.
+        """
+        image_provs = []
+        sources_provs = []
+        psf_provs = []
+        wcs_provs = []
+        zp_provs = []
+
+        for im in self.upstream_images:
+            image_provs.append(im.provenance)
```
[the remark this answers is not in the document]
rknop: (See comment added to references.py)

models/image.py
```diff
@@ -340,285 +341,66 @@ def mid_mjd(self):
...
-        sa.Text,
+    bkg_rms_estimate = sa.Column(
+        sa.Float,
         nullable=True,
-        doc='Free text comment about this image, e.g., why it is bad. '
+        index=True,
+        doc=(
+            'Background RMS estimate for the image, '
+            'from the first time the image was processed.'
+        )
```
rknop, line 403: These two fields should be set in preprocessing.py (go to line ~265, which is where the values are extracted).

models/image.py
```diff
@@ -1242,6 +1077,244 @@ def load(self):
...
+                if len(zps) > 1:
+                    raise ValueError(
+                        f"Image {im.id} has more than one ZeroPoint matching upstream provenance."
+                    )
+                elif len(zps) == 1:
+                    im.zp = zps[0]
+
+    def get_upstreams(self, session=None):
+        """
+        Get the upstream images and associated products that were used to make this image.
```
rknop, line 1226: This description isn't really right.  The sources, psf, wcs, etc. will only be loaded for the upstream images if the upstream images are already in the session and you've called something to fill in the .sources, .zp, etc. fields.

models/image.py
```diff
@@ -17,37 +14,43 @@
...
-              sa.ForeignKey('images.id', ondelete="CASCADE", name='image_sources_source_id_fkey'),
+              sa.ForeignKey('images.id', ondelete="CASCADE", name='image_upstreams_association_upstream_id_fkey'),
               primary_key=True),
-    sa.Column('combined_id',
+    sa.Column('downstream_id',
               sa.Integer,
-              sa.ForeignKey('images.id',ondelete="CASCADE", name='image_sources_combined_id_fkey'),
+              sa.ForeignKey('images.id', ondelete="CASCADE", name='image_upstreams_association_downstream_id_fkey'),
               primary_key=True),
 )
```
rknop, line 50: So, rather than having this assocate back to images, it should associate back to something else.

That something else is "zeropoint".  Because the coadd process uses source lists, and WCSes and zeropoints from at least some of the images, we have to know all of (image, source_list, wcs, zeropoint) to figure out what actually went into making this coadded (or subtracted) image.

Because it sounds weird to link to a zeropoint, rename Zeropoint to CalibratedImage.  It will have database fields wordcoordinates_id, provenance_id, and the actual zeropoint stuff (zp, apercor, etc).  Give it convenience method "source_list" that gets the source list from the worldcoordinates the CalibratedImage points to, and "image" that gets the image from that source list.

Then, in this table, upstream_id is actually a foreign key into CalibratedImage.  That way, we've got everything we need to figure out what went into the image combination (be it coadd or subtraction).

rknop (commented):
I haven't actually looked at everything yet; I wanted to get these comments in before I had to run off and do something else.  There are a number of small comments, but a bunch of other more substantive comments.  As I was writing them, I realized that a lot of them were related.

I think the fundamental issue is what the upstream_id of the image_upstreams_assocation_table points to.  See the comment we made on that table; if we make that change, it solves a lot of problems of having to figure out what goes with what.  It makes loading all of the upstream's associated products easier, without any need to parse provenances.

(In another only slightly related change that spans several comments, I think we should store the nature of the assocation in the image_usptreams_assocation_table, rather than having specific ref_image_id and new_image_id fields in the image table.)

models/image.py
guynir42: Why don't we rename it ImageProducts?

guynir42:
I'm going to attempt to fix this based on the Provenance model. I don't think it will be a big problem. 

I am not sure why it is so bad to have multiple foreign key paths connecting the same objects if the paths add information. But I think I can avoid all of them using Provenance. 

The problem with your suggestion of linking through the ZP is that (a) the PSF is not directly on the link path and (b) there could be other objects related to an object that have nothing to do with ZP but you'd have to link them through the ZP (or CalibtrateImage) and you may as well just save links to all the data products you want and save it as an ImageProducts. 

Let's jump on a zoom as early as you can today, and try to pin this down.

models/image.py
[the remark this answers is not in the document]
guynir42: No longer enforcing a single provenance for all the objects in the upstreams list.

models/references.py
```diff
@@ -85,8 +106,137 @@ class ReferenceEntry(Base, AutoIDMixin):
...
+    provenance_id = sa.Column(
+        sa.ForeignKey('provenances.id', ondelete="CASCADE", name='references_provenance_id_fkey'),
+        nullable=False,
+        index=True,
+        doc=(
+            "ID of the provenance of this reference. "
+            "The provenance will contain a record of the code version "
+            "and the parameters used to produce this reference. "
+        )
+    )
```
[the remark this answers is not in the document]
guynir42: The Reference needs a provenance so that it can keep track of the upstream provenances so it can match the Image it is linking to with the other products. 
The other option is to have it link to something other than an Image (like to the ZeroPoint or to the ImageProducts). 
If we are basing the ability of an object to find the image products on the Provenance then the reference must keep at least a list of upstream provenances (might as well save a whole Provenance at that point). 

Another thing that could be useful is that for example the reference producing pipeline might have some parameters (such as "collect all images from this date to this date") and then a Provenance will help you choose which "reference building strategy" is the one you want. Same for "custom references" that people may want to build on the fly.

pipeline/detection.py
```diff
@@ -418,7 +420,7 @@ def _run_sextractor_once( self, image, apers=[5, ], psffile=None, tempname=None,
             psfargs = []
             paramfilebase = astromatic_dir / "sourcelist_sextractor.param"
 
-        # SExtractor reads the measurements it produes from a parmeters
+        # SExtractor reads the measurements it produces from a parameters
```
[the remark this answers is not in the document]
guynir42: I have tried to (not so) gently suggest that using a modern IDE could help with this.

pipeline/photo_cal.py
```diff
@@ -273,5 +273,8 @@ def run(self, *args, **kwargs):
             ds.zp = ZeroPoint( source_list=ds.sources, provenance=prov, zp=zpval, dzp=dzpval,
                                aper_cor_radii=sources.aper_rads, aper_cors=apercors )
 
+            ds.image.zero_point_estimate = ds.zp.zp  # TODO: should we only write if the property is None?
+            # TODO: we should also add a limiting magnitude calculation here.
```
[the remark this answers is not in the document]
guynir42: I have strategically decided not to add this in this PR. I have made an issue and we can return to this at any point when we decide the details.

models/base.py
```diff
@@ -225,9 +232,18 @@ def recursive_merge(self, session, done_list=None):
         done_list.add(obj)
 
         # only do the sub-properties if the object was already added to the session
```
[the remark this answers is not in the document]
guynir42: I don't understand this comment either :) removed it

models/image.py
guynir42: The project really doesn't have to be the same, that's easy to remove. 
The problem is that if you are on a grid of fields, then `target` will tell you the field ID and `section_id` will tell you where you are on that grid. This is the only thing we have right now that prevents coadding/subtracting the wrong sky coordinates. 

I suggest that we make a separate PR where we add the functionality of checking if images are overlapping or not, before combining them, and then use that instead of these restrictions. 

in the meantime I will try to see if we actually need to have values for the `section_id` and `project` and `target` in the coadded / subtracted image, and how critical is it if we just pick the values from one image (risking that it will be different from the others).

models/image.py
guynir42: I think there's very little cost to not assuming that the images don't overlap each other (as per our discussion of getting reference images from different instruments), so I will leave it here. 

The images are sorted by `mjd` a few lines above that.

models/image.py
guynir42: I will update the docstring. This should be used for coadds but this also could include a master bias or flat combination.

models/references.py
guynir42: Discussion below.

models/image.py
guynir42: I will do the actual filling in of these fields in a separate PR.

rknop:
So then if you want an imageproduts table, remove all of the foreign keys from source list, wcs, psf, and zeropoint.  Use just the imageproducts table as a way of linking them all together.  Also move the provenances to the imageproducts table.  In fact, then we don't need to keep upstream provenances in the provenances any more, because all of the individual provenances will be in the imageproducts table.

models/image.py
```diff
@@ -91,98 +94,101 @@ def format(self, value):
...
-    @classmethod
-    def is_sub(cls):
-        return cls.ref_image_id.isnot(None) & cls.new_image_id.isnot(None)
+    is_coadd = sa.Column(
+        sa.Boolean,
+        nullable=False,
+        default=False,
+        index=True,
+        doc='Is this image made by stacking multiple images.'
+    )
```
rknop, line 183: It may well be that sub and coadd are the only compsite images we'll deal with, but if we're going to store flatfields, and flatfied inputs, as images, then there are others.

I wonder, though, if instead of a pair of booleans, it might be worth having a 8-bit or 16-bit integer here that's a "compsite type", where 0 means not a composite.

sub and coadd are two composites.

But, a combined flatfield is a different sort of composite from sub and coadd, as is a combined bias.  (Whether a combined bias and a combined flatfield are the same kind of composite is a longer combination.)  I can't think right now of any other sorts of composites, but we might come up with some in the future, and if we have a byte here for composite type rather than a pair of flags, it's more future proof.

models/image.py
rknop: I think this comment still stands; I really won't put target here, as that will come back to bite us.  (Consider sky flats, if nothing else-- that's something that doesn't match RA/Dec either!)  We've also discussed earlier about how section_id will need to be different in combinations we make.  For the most general consideration of image combaintion, we should just be checking instrument, telescope, filter, and perhaps type.  In this function, only those general considerations should be checked.  More specific considerations will come from the code that implements the actual combination.

As for setting the attribute in the combined image, if the corresponding attributes aren't the same in all the input images, they should be set to some placeholder value that indicates it comes from something heterogeneous.  Or, it might be different for different combinations.

models/image.py
```diff
@@ -877,72 +702,110 @@ def from_images(cls, images):
...
 
-        output.source_images = images
         base_type = images[0].type
         if not base_type.startswith('Com'):
             output.type = 'Com' + base_type
 
+        output.upstream_images = images
+
+        # mark the first image as the reference (this can be overriden later when actually warping into one image)
+        output.ref_image_index = 0
```
rknop, line 719: (Again associated with comment above about using image id rather than index to identify the reference.)

models/image.py
rknop: Unresolved this -- I still think this should be only checking instrument, telescope, and filter.  This is the highest-level, most general thing that associates images.  Since we know there are cases where projects, sections, and targets will be mixed, we shouldn't be putting checks on that here.  If there are places in the more specific implementation functions where those checks are appropriate, put them there, but putting them here limits us a priori in ways we know we don't want to be limited.

models/image.py
```diff
@@ -877,72 +702,110 @@ def from_images(cls, images):
...
             if att == 'section_id':
                 ref_value = str(ref_value)
                 new_value = str(new_value)
-            if ref_value != new_value:
+
+            # TODO: should replace this condition with a check that RA and Dec are overlapping?
+            #  in that case: what do we consider close enough? how much overlap is reasonable?
+            #  another issue: what happens if the section_id is different, what would be the
+            #  value for the subtracted image? can it live without a value entirely?
+            #  the same goes for target. what about coadded images? can they have no section_id??
```
rknop, line 783: coadded images may well be coadded from different section ids, so the section id for a coadded image is not well defined.  We need to make sure that the image infrastructure is able to handle that, which may require some work.  So, we should only check that the filter is the same.

Re: RA and Dec overlapping, I'd say if they have at least (say) 5%, or even 1%, overlap, go ahead and do the subtraction.  Somebody might be trying to find something specific they know is in that overlap region.  The actual subtraction algorithm itself may fail if there's not enough overlap to psf match, for instance, but at this high-level definition of the subtraction, we shouldn't worry about that.  That will be the problem of the implementation of the subtraction algorithm.

models/image.py
```diff
@@ -1242,6 +1105,287 @@ def load(self):
...
+
+        IMPORTANT RESTRICTION: to maintain the ability of a downstream to recover its upstreams
+        using the provenance (which is the definition of why we need a provenance) it is not
+        allowed for images with the same provenance to have related products (e.g., a SourceList)
+        that have different provenances.  This is because the downstream would not know which
+        SourceList to use.  Images from different instruments, or a coadded reference vs.
+        a new image, would have different provenances, so their products could (and indeed must)
+        have different provenances. But images from the same instrument with the same provenance
+        should all be produced using the same code and parameters, otherwise it will be impossible
+        to know which product was processed in which way.
```
rknop, line 1133: ...I'm not sure this restirction is a good idea.  If we do have this restriction, then a lot of the provenance stuff we have in place isn't actually necessary. Once you've got the image id and the code version, the provence is uniquely defined for all downstream products.

This restriction will make it impossible for us to do tests by trying different paremeters for downstream processes (catalogs, zeropoints, whatever) off of the same image.  This is something I suspect we're going to want to have to do.  (In the past, I've done it ad hoc, which of course becomes a pain to keep track of, and difficult to do in bulk.  The provenace system you've defined alows us to keep track of this, which is great, but adding this restriction totally undermines the flexibility of it.

If we DO go this way with this restriction, then I'd strongly recommend we think about other places where we can simplify the code, because I think we have a lot of complexity that becomes unnecessary if this restriction is in place.

I think there's still a fundamental design problem here.  To make a coadd, you need to have the source list, and perhaps the WCS, of the images that go into this coadd.  Because right now for a coadd, you list the images themselves as the upstreams, you now are introducing this requirement that there can only be one downstream of any time for an image.  But I don't think that's the right solution.

The right solution is, I think the one I proposed a couple of months ago.  List the things that actually define what goes into a coadd, even if it's not entirely intuitive. For sky images, that could be a zeropoint object, as a zeropoint will point uniquely to upstream WCS, sourcelist, and image.  That doesn't work for flatfields, though, which won't have any of those, only images.  Maybe the solution is an ImageProducts table, but as I said when we were talking about that, if we add that, then we should also remove the direct foreign keys from things like source lists, WCS, etc., so that the imageproducts table becomes the way to find those linkages.  (What I don't want is to have the linkages multiply defined.)

rknop (commented):
I haven't read everything, because I hit a wall at "get_upstream_provenances" in image.py.

We don't want to add the restriction that you add in the function's docstring.  See the comment I put there. As I say in that comment, I think there's a design flaw in here in listing images as the upstreams of a coadd, because the coadd depends not just on images, but also on (at least) sourcelists.

models/image.py
```diff
@@ -17,37 +14,43 @@
...
-    image_preprocessing_dict,
-    string_to_bitflag,
-    bitflag_to_string,
 )
 
 import util.config as config
 
-image_source_self_association_table = sa.Table(
-    'image_sources',
+# links many-to-many Image to all the Images used to create it
```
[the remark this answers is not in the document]
guynir42: This is indeed many to many. Images A, B, and C can be used to create Image 1, but they can also be used to create image 2 (or a subset of them).

models/image.py
```diff
@@ -91,98 +94,101 @@ def format(self, value):
...
-    ref_image_id = sa.Column(
-        sa.ForeignKey('images.id', ondelete="CASCADE", name='images_ref_image_id_fkey'),
+    ref_image_index = sa.Column(
+        sa.Integer,
         nullable=True,
-        index=True,
         doc=(
-            "ID of the reference image used to produce a difference image. "
-            "Only set for difference images. This usually refers to a coadd image. "
+            "Index of the reference image used to produce this image, in the upstream_images list. "
```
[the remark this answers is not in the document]
guynir42: The upstream_images is guaranteed to be sorted by MJD. This list does not exist on the DB at all (sorted or not), so it is up to SQLA to sort the list as it loads the relationship and also I've added this to the getter or setter (I think). 

We could refactor this to use image ID but as you say, duplicate data...

models/image.py
```diff
@@ -91,98 +94,101 @@ def format(self, value):
...
-        primaryjoin='images.c.new_image_id == images.c.id',
-        uselist=False,
-        remote_side='Image.id',
+    new_image_index = sa.Column(
+        sa.Integer,
+        nullable=True,
         doc=(
-            "New image used to produce a difference image. "
-            "Only set for difference images. This usually refers to a regular image. "
+            "Index of the new image used to produce a difference image, in the upstream_images list. "
```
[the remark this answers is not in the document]
guynir42: That could actually work...

models/base.py
```diff
@@ -1248,5 +1273,134 @@ def cone_search( self, ra, dec, rad, radunit='arcsec' ):
...
+            The session to use for the update. If None, will open a new session,
+            which will also close at the end of the call. In that case, must
+            provide a commit=True to commit the changes.
+        commit: bool (default True)
+            Whether to commit the changes to the database.
+        """
+        # make sure this object is current:
+        with SmartSession(session) as session:
+            new_bitflag = 0  # start from scratch, in case some upstreams have lost badness
+            for upstream in self.get_upstreams(session):
```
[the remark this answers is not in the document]
guynir42: I did implement the get upstream/downstream for the SourceList.

models/image.py
```diff
@@ -91,98 +94,101 @@ def format(self, value):
...
-                return True
-        except DetachedInstanceError:
-            if not self.is_sub and self.exposure_id is None:
-                return True
+    def new_image(self):
+        if self.new_image_index is None:
+            return None
+        if len(self.upstream_images) <= self.new_image_index:
+            raise RuntimeError(f'Index {self.new_image_index} is out of range for upstream images!')
+        return self.upstream_images[self.new_image_index]
```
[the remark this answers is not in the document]
guynir42: makes sense.

models/image.py
guynir42: I am generally against future proofing.

models/image.py
guynir42: I added this as Issue #179

models/image.py
```diff
@@ -91,98 +94,101 @@ def format(self, value):
...
-        nullable=True,
-        index=True,
-        doc=(
-            "ID of the new image used to produce a difference image. "
-            "Only set for difference images. This usually refers to a regular image. "
-        )
-    )
+            # make sure the upstream_images list is sorted by mjd:
+            self.upstream_images.sort(key=lambda x: x.mjd)
+            self.ref_image_index = self.upstream_images.index(value)
```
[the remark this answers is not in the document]
guynir42: I'm going to defer this point to a separate issue as it would be much easier to fix after we merge all the remaining changes (Issue #178 )

models/image.py
guynir42: Will fix this in issue #179

models/image.py
guynir42: Issue #179

models/image.py
guynir42: I think the docstring is not clear. If image has provenance A, then it can have SourceList provenance 1 and 2 associated with it, just not *relative to this specific coadd*. I will have to fix the text to be more clear. 

So the way it should be: 
Any Image can have any number of different products with different provenances. 
If you put two images with the same provenance *into a coadd* then the associated products (e.g., SourceList) that you pick to use along with those upstream images, must share a provenance as well. They could have other SourceLists but those are not involved in making this specific coadd image.

models/image.py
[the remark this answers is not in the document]
guynir42: I will do this in Issue #178

models/image.py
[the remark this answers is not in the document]
guynir42: Issue #178

models/image.py
rknop: Oh!  OK, that's not nearly as bad as what I was thinking... in fact, it's not bad at all.  Yeah, the docstring sounds like things are far more restrictd than they really are.  But, yeah, makes sense that all the things going into a coadd need to have the same provenance, and that that provenance is what contributes to the coadd's provenance.

OK, I will keep reading further down on the PR.

models/image.py
guynir42: I am also inclined to say we should stop thinking about calibration objects (such as flats) as Image objects. They have much less in common to our definition of Image and I don't think we gain much from asking Image to also represent calibration data. 
I think flats should go as CalibratorFile that are FileOnDiskMixin that happen to be a 2D dataset. But this is a discussion for another time (see Issue #162 )

guynir42:
@rknop I agree that we should loosen the restriction on coadding images with the same target or section_id. I want to leave it for now because it works and it would be easier to fix on top of the other PRs, so I put it into an Issue #179. 
Same goes for the indexing of the ref/new in the upstream images list (Issue #178)

As for the provenances, I will have to fix the docstring because it is unclear and indeed managed to confuse you. I hope I cleared it up in the comment. I will also change it in the code. Please continue to review and let me know if there are more changes needed. 

I think we will have to defer a lot of changes (other than minor bugs and incorrect docstrings) to fix after we merge this sequence of PRs. I promise to go back to fixing things on the spot once we are synchronized!

models/image.py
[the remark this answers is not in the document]
rknop: The list *does* exist in the database, in the association table.

models/base.py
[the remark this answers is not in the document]
rknop: Huh... grep has failed me.

models/image.py
rknop: Yeah, I'd agree that we shouldn't think about flatfields as Images.

models/image.py
rknop: The restriction is really about what's loaded into the the associated fields in self.upstream_images (which, as documented,must already be loaded for this routine to work).

models/references.py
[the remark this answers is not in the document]
rknop: Revisiting this... as it exists right now, won't the underlying image have in its provenance all the provenances of all the derived products?  (Assuming that the code did it right and set the image's provenance from get_upstream_provenances().)  As such, I don't think the references table needs its own provenance.

pipeline/data_store.py
```diff
@@ -540,6 +548,20 @@ def get_image(self, provenance=None, session=None):
 
         return self.image  # could return none if no image was found
 
+    def apped_image_products(self):
```
rknop, line 551:
```suggestion
    def append_image_products(self):
```

rknop (approved):
OK, once I understood the comment in get_upstream_provenances, this all looks fine, except for a bunch of mostly small stuff that I *think* you've already made into other Issues. 

I do think that, with your latest provenance upstreaming/downstreaming, Reference no longer needs its own provenance.  Unless I'm wrong about that, create an issue to remove it.

Otherwise, perhaps calrify that one comment before merging.

pipeline/data_store.py
guynir42: Yes, this was fixed in one of the later PRs.

guynir42:
I'm leaving the Reference provenance. The reason is that it is linked to the coadd image in a 1-to-1 relationship using the image_id, but the coadd image could have multiple product versions (e.g., it can have more than one SourceList). To allow the Reference to choose the products of the coadd, it needs a Provenance. 
The products of the images that *went into the coadd* can be recovered using the coadd's provenance, but not the products that are downstream of the coadd. 

In addition, we may find it useful at some point to put parameters on the Reference, for example if we make some policy about how references are generated and want to capture those in the provenance (as opposed to manually making the references each year or at the start of the survey).

Issue #140, opened 2023-11-23
Test data folder


guynir42:
We should separate the DECam_examples and other such folders in `data` from a dedicated `test_data` folder that is 100% ok to wipe out at end of tests. 
Probably want to put all the cached data that is downloaded once into folders like DECam_examples or something like that. 
There should be a clear distinction between:
- Data that comes with the repo and so is not in the gitignore file. 
- Data that get cached on the first time tests are run, but should be wiped if you want to test the fetching of those files. 
- Data that gets wiped at the end of the test suite run.

Issue #132, opened 2023-11-20
Cleanup fixtures


guynir42:
The `conftest.py` file is becoming large and disorganized. It is very hard to find what you are looking for. 

I don't know if splitting it up is the way to go, but at least dividing it into sections where it is clear where the DECam images are, where the ZTF/PTF images are, where the generic fixtures are, etc. 

Also, wouldn't hurt to make the naming of fixtures a little more consistent.

guynir42:
Addressed by #144

PR #144, opened 2023-11-26
Fixture cleanup

guynir42:
The fixtures file has been growing large and too complicated. 

- [x] split it off into files in the "fixtures" folder. 
- [x] rename all the "demo" images/exposures to "sim_image" and so on. 
- [x] make a consistent naming convention for all fixtures from instruments (DECam, ZTF, PTF, etc). 
- [x] make sure all temporary data is saved in a "temp_data" folder under "tests" that is nuked at end of tests.
- [x] make sure all cached data is saved in a "data_cache" folder under "data". Make sure to cache all the downloaded images. 
- [x] replace some of the fixtures that return a datastore with fixtures that return an Image with some other objects attached (e.g., sources, psf). 
- [x] consider separating fixtures that are already committed with "uncommitted" fixtures that are used to test the process of saving to DB. This will shorten a lot of tests that have to save, load a provenance, recursive merge, and commit the data, all before starting the actual test. 
- [x] Make sure all the provenances of fixtures make sense in terms of upstreams. 
- [x] Supply datastores that can load their data on demand from the cache. 
- [x] Create pipeline object fixtures with a "test_parameter" so we can re-do calculations using a new Provenance inside the tests. 
- [x] Add a "was calculated" flag to pipeline objects to know if they did the work or if they just pulled data from the datastore. 
- [x] Make sure tests clean up the database, and not using the "wipe all provenances" method in the conftest file, as that could be catastrophic if we ever accidentally run a test on production.


guynir42:
Addresses #132 and #140

rknop:
Haven't looked at this yet, but --

Why are you opposed to using a Datastore as a way of storing an image and related things?  That's *exactly* what it is!

guynir42:
@rknop this is a really long one. It took me over a week. The reason is that while cleaning up the fixtures I uncovered many bugs and things that needed fixing along the way. Good luck!

guynir42:
@rknop this is ready for a review after merging the previous PR.

models/base.py
```diff
@@ -268,6 +326,215 @@ def get_downstreams(self, session=None):
...
+        for key in sa.inspect(self).mapper.columns.keys():
+            value = getattr(self, key)
+            if key == 'md5sum' and value is not None:
+                if isinstance(value, UUID):
+                    value = value.hex
+            if key == 'md5sum_extensions' and value is not None:
+                if isinstance(value, list):
+                    value = [v.hex if isinstance(v, UUID) else v for v in value]
+
+            if key == 'aper_rads' and isinstance(value, np.ndarray):
```
rknop, line 360: I wonder if it might be worth keeping a list somewhere that lists all the keys that are array keys, and another list that are date/time keys.  Then you could just iterate through those lists here, and if we add new keys that fit those definitions, we just add them to the lists rather than adding to the if block.

Either way, we need to think about how we're going to make sure we (and anybody else who expands the code) remembers that something needs to get edited somewhere whenever defining a new database column that's either an array or a datetime.

models/base.py
```diff
@@ -1018,18 +1297,32 @@ def delete_from_disk_and_database(self, session=None, commit=True, remove_folder
...
-        self.md5sum_extensions = None
-        self.filepath_extensions = None
-        self.filepath = None
+        try:
+            self.md5sum = None
+            self.md5sum_extensions = None
+            self.filepath_extensions = None
+            self.filepath = None
+        except:
+            pass
```
rknop, line 1306: ?? why the try/except here?  Setting a bunch of fields to None isn't going to throw an exception.

models/image.py
```diff
@@ -1120,22 +1124,18 @@ def get_upstream_provenances(self):
...
+        IMPORTANT RESTRICTION: to maintain the ability of a downstream to recover its upstreams
+        using the provenance (which is the definition of why we need a provenance) it is not
+        allowed for different images with the same provenance to have related products
+        (e.g., a SourceList) that have different provenances.  This is because the downstream
+        would not know which SourceList to use.  Images from different instruments, or a
+        coadded reference vs. a new image, would have different provenances,
+        so their products could (and indeed must) have different provenances.
+        But images from the same instrument with the same provenance  should all
+        be produced using the same code and parameters, otherwise it will be
+        impossible to know which product was processed in which way.
```
rknop, line 1138: Ak!  Looks like this PR reverts the comment fix you put in.  Might be worth making sure that any other changes you made in review of the last PR didn't get reverted as well.

pipeline/data_store.py
```diff
@@ -823,10 +871,22 @@ def get_reference(self, provenance=None, session=None):
...
+            image = self.get_image(session=session)
+
+            if self.reference is not None:
+                if not (
+                        (self.reference.validity_start is not None or
+                         self.reference.validity_start <= image.observation_time) and
+                        (self.reference.validity_end is not None or
+                         self.reference.validity_end >= image.observation_time) and
+                        self.reference.filter == image.filter and
+                        self.reference.target == image.target and
```
rknop, line 884: Usual comment: we should remove this, as there may well be different "targets" defined for the same point on space, and that will use the same reference.

tests/conftest.py
```diff
@@ -1,80 +1,111 @@
...
+    FileOnDiskMixin.configure_paths()
 
-    yield
-    # Will be executed after the last test
-    # print('Final teardown fixture executed! ')
 
+# This will be executed after the last test (session is the pytest session, not the SQLAlchemy session)
+def pytest_sessionfinish(session, exitstatus):
+    # print('Final teardown fixture executed! ')
     with SmartSession() as session:
```
rknop, line 44: For clarity, might be worth naming the parameter pytest_session, rather than naming it session and then overwriting that value with something that's a completely different type.

(For instance, if in some future we come back and need to do something with the pytest session, looking back at the function definition might give us the idea that "session" is the variable to use for that.)

tests/conftest.py
```diff
@@ -121,8 +152,8 @@ def rnd_str(n):
     return ''.join(np.random.choice(list('abcdefghijklmnopqrstuvwxyz'), n))
 
 
-@pytest.fixture
-def config_test():
+@pytest.fixture(autouse=True)
+def test_config():
```
rknop, line 156: I think the reason I named it config_test rather than test_config was so that it wouldn't be mistaken by the reader as an actual test that gets run.  (Not a big deal.)

tests/conftest.py
```diff
@@ -830,188 +263,17 @@ def archive():
...
-def example_psfex_psf_files():
-    psfpath = ( pathlib.Path( FileOnDiskMixin.local_path )
-                / "test_data/ztf_20190317307639_000712_zg_io.083_sources.psf" )
-    psfxmlpath = ( pathlib.Path( FileOnDiskMixin.local_path )
-                   / "test_data/ztf_20190317307639_000712_zg_io.083_sources.psf.xml" )
-    if not ( psfpath.is_file() and psfxmlpath.is_file() ):
-        raise FileNotFoundError( f"Can't read at least one of {psfpath}, {psfxmlpath}" )
-    return psfpath, psfxmlpath
+@pytest.fixture( scope="module" )
+def catexp(data_dir, persistent_dir):
```
rknop, line 267: Is this not something that makes sense to move into one of the fixture subfiles?

tests/pipeline/test_extraction.py
```diff
@@ -325,26 +333,34 @@ def test_run_detection_sextractor( decam_example_reduced_image_ds ):
     # wrong.
 
     assert ds.sources.psffluxadu()[0].min() == 0.0
-    assert ds.sources.psffluxadu()[0].max() == pytest.approx( 1726249.0, rel=1e-5 )
-    assert ds.sources.psffluxadu()[0].mean() == pytest.approx( 48067.805, rel=1e-5 )
-    assert ds.sources.psffluxadu()[0].std() == pytest.approx( 169444.77, rel=1e-5 )
+    assert ds.sources.psffluxadu()[0].max() == pytest.approx( 1726249.0, rel=1e-3 )
+    assert ds.sources.psffluxadu()[0].mean() == pytest.approx( 48067.805, rel=1e-3 )
+    assert ds.sources.psffluxadu()[0].std() == pytest.approx( 169444.77, rel=1e-3 )
```
rknop, line 338: Were these failing with the tighter constraints?  That surprises me, and alarms me a little bit.  (I'm not sure what would be different-- though if you were running it in your own environment rather than in the Dockerfile, it might be a different version of SExtractor.)

tests/seechange_config_test.yaml
```diff
@@ -5,15 +5,19 @@ overrides:
...
 db:
   host: seechange_postgres
 
 archive:
   archive_url: http://archive:8080/
   verify_cert: false
   path_base: test/
-  local_read_dir: null
-  local_write_dir: null
+  local_read_dir: /storage/base/
```
rknop, line 19: ...hmm, this one alarms me a little bit.  I will have to look at how archive is implemented, but the idea the archive is contacted via url XOR the archive is read/written from local files.  Having a local read/write dir *and* a archive_url may lead the logic to not doing what we think it's doing.

I would prefer to document that local_read_dir and local_write_dir can be non-null if and only if archive_url is null. (And, then, perhaps, put a check and exception if it finds it otherwise.)  Right now, it's not obvious how the archive would be accessed.

I'd rather have something ad-hoc that's explicitly for the test environment than reusing the archive configuration itself, so we can be sure that the archive code isn't misusing the configuration here when we test it.

tests/conftest.py
```diff
@@ -820,7 +253,7 @@ def archive():
         # To tear down, we need to blow away the archive server's directory.
         # For the test suite, we've also mounted that directory locally, so
         # we can do that
-        archivebase = f"{os.getenv('SEECHANGE_TEST_ARCHIVE_DIR')}/{cfg.value('archive.path_base')}"
+        archivebase = f"{test_config.value('archive.local_read_dir')}/{test_config.value('archive.path_base')}"
```
rknop, line 256: See my comment on seechange_config_test.yaml -- I think it's hazardous to reuse the archive configuration options for something that the tests need internally, but that's not really correct for the way the archive itself works.

rknop (changes requested):
This one does a whole lot more than the title would suggest....

I'm very happy to have the fixtures broken out into subfiles rather than in one gigantic conftest file.  (I've solved this in the past by just putting fixtures that are local to one set of tests in the test file itself, but that doesn't help when there are some fixtures you might want in a couple of different test files.)

Lots of minor comments, but only one that I'd really like to see changed.  See the comment in seechange_config_test.yaml -- I'd really rather not use the archive configuration as the place where the tests figure out where they do their own internal testing of whether things went to the archive, and whether the archive got cleaned up.  (So, I'd recommend bringing back the SEECHANGE_TEST_ARCHIVE_DIR environment variable as a way of configuring the tests without setting configuration variables that might have implications outside of what we're really using them for.)

models/base.py
guynir42: Maybe add an error in case a column has a forbidden type at the end of this call.

models/base.py
guynir42: Oh, with SQLA you'd be surprised what sorts of things throw exceptions...

guynir42: I can re-run the tests without this part and see if anything shakes out

models/image.py
guynir42: This was the only change, and indeed I had some trouble with merging. I will bring back the text from the previous PR.

pipeline/data_store.py
guynir42: yes. I will make sure to round these up in a separate PR.

tests/conftest.py
guynir42: I did that, but pytest is really picky about the names of variables (think about fixtures as parameter names). 
I can call our session dbsession or something like that.

tests/conftest.py
guynir42: Makes sense. But it seems like the fixture tag took care of that.

tests/conftest.py
guynir42: Don't we want to have a different folder for the archive when running tests? What's the difference if I put it into the config vs. if I use an environmental variable?

tests/conftest.py
guynir42: I split the fixtures based on survey (mostly) so this Gaia catalog doesn't really belong with any of them. 

That said, it may belong because it targets a field which corresponds to one of the surveys' test images. I'll check.

tests/pipeline/test_extraction.py
guynir42: Yes, I think that was it. But I was unable to confirm the version on the dockerized build.

tests/seechange_config_test.yaml
guynir42: I agree that we should used URL or local storage. I can add a check for that. 

I am not sure why configuring the archive differently for the tests is not good.

guynir42:
@rknop I don't know why it would be a problem to configure the archive for the test via the config file. I'm also not sure if there's a reason to use an environmental when you could just as well use the config file override.

tests/seechange_config_test.yaml
rknop: The intention I had was to have the archive set up to use the HTTP interface for the tests (that's how it's set up in the docker file).  The reason for this is because it's the more general interface (i.e. the one that would work anywhere, whereas the "local file" interface will only work if you're on the machine where the archive has local files).  (Both interfaces are tested separately in the submodule's own tests.)

I'm not sure what the logic is in the archive code, but if it sees a non-null local_read_dir and local_write_dir, it may just use that and not use the url.  In any event, by setting both a url and local dirs, we're ambiguously configuring the archive, and we don't want to do that.

What we really need for the tests is a way to find the archive's directories to make sure that the http transfers completed correctly, and to clean it up when we're done.  This is outside of the normal usage of the archive, and so should be configured outside of the configuration of the normal usage of the archive.

rknop: Looking at the code, it's actually documented that if both archive_url and local_write_dir are set, it with write the file twice, once to local_write_dir, and once via the HTTP interface.  If local_read_dir is not None, then it will read from there rather than using the http interface.

So, you are changing the behavior of the archive generally by configuring both.  Unless we want to change how we interface with the archive in the tests, we need to leave local_read_dir and local_write_dir as None

Issue #154, opened 2023-12-13
SourceList _is_star


guynir42:
There's a `_is_star` property of the source list that is lazy loaded from the data table's `CLASS_STAR` but it is not very reliable. 
Right now we are mostly depending on the catalog itself to provide only stars and then the photometric/astrometric calibration can go on without using `is_star` but maybe we can do better by saving the information from the catalog into the source list? 

Right now the data doesn't live in the database but in the fits table, so I don't know how we would add this.

guynir42:
Maybe add a bunch of columns to the source list table right when we calculate it. 

This may even be a shortcut to allow doing astro_cal and photo_cal without having to re-fetch Gaia data.

guynir42:
In `calc_aper_cor` in SourceList must come back to using some sort of star classifier, but not necessarily the is_star property.

guynir42:
Confirmed that this indeed fails to find sufficient stars when calculating aperture corrections on the sample of PTF images I use in the fixtures (as of PR #155 ).

guynir42:
Probably best thing to do is to cross match to Gaia (or additional catalogs) at the extraction phase, and then use one of those catalogs to tell if a source is a star, a galaxy, or something unknown (e.g., a supernova). 

This will then be propagated through the pipeline and help choose the stars to use in PSF finding or in ZP estimation. 

We will probably need to also check if there is already the correct catalog matched to the sources when doing astrometric/photometric calibration and only do a cross match to the required catalog if that hasn't happened yet.

guynir42:
Make sure to integrate the new star/galaxy catalog from Adam Miller's group.

rknop:
Related to Issue #381

rknop:
We didn't have to iterate!  psfex does it's own (presumably robust) star/galaxy separator.  So, we can just use the FWHM we get from psfx before the second run.  Closing this issue because #381 has been moved to Done.

PR #155, opened 2023-12-13
Fix PTF fixtures

guynir42:
Making some improvements to the way we download and process PTF images. 

This is essential on the path to testing the subtraction algorithms. 

- [x] Build a PTF instrument class to allow easy processing of PTF exposures. 
- [x] Download PTF images (in this case, preprocessed Exposures)
- [x] Make a datastore with all the SourceList, WorldCoordinates and ZeroPoint
- [x] Download a batch of images and process each of them to build a reference image. 
- [ ] Build a coadd image and a Reference object for our PTF example. (I'm deferring the coadd construction to a new PR)
- [x] Make it easy to pull down a bunch of PTF images after the reference is made, where we can extract a lightcurve.


models/ptf.py
```diff
@@ -38,10 +58,111 @@ def get_filename_regex(cls):
...
+    @classmethod
+    def _get_fits_hdu_index_from_section_id(cls, section_id):
+        cls.check_section_id(section_id)
+        return 0  # TODO: improve this if we ever get multiple HDUs per file
+
+    def _make_new_section(self, section_id):
         """
-        Get a short name used for e.g., making filenames.
+        Make a single section for the PTF instrument.
+        The section_id must be a valid section identifier ([0, 11]).
```
rknop, line 82: Perhaps put in a check that it's an int between 0 and 11?

models/ptf.py
```diff
@@ -38,10 +58,111 @@ def get_filename_regex(cls):
...
+        #     'r': np.array([-0.28526417, 0.65444024, -0.25415955, -0.00204337]),
+        #     'i': np.array([-0.2491122, 0.51709843, 0.02919352, -0.02097517]),
+        #     'z': np.array([-0.38939061, 0.70406435, 0.04190059, -0.01617815])
+        # }
+        # TODO: must find better transformations!
+        transformations = {
+            'R': np.array([0]),
+            'V': np.array([0]),
+            'I': np.array([0]),
+        }
```
rknop, line 148: ...I might say "must find transformations" rather than "must find better transformations" in the TODO, because it's very obviously just a placeholder right now, and will not have any resemblance to reality.

(Somebody somewhere must have done this already.)

models/source_list.py
```diff
@@ -444,7 +430,7 @@ def calc_aper_cor( self, aper_num=0, inf_aper_num=None, min_stars=20 ):
 
         bigflux, bigfluxerr = self.apfluxadu( apnum=inf_aper_num )
         smallflux, smallfluxerr = self.apfluxadu( apnum=aper_num )
-        wgood = self.is_star & self.good & ( bigflux > 5.*bigfluxerr ) & ( smallflux > 5.*smallfluxerr )
+        wgood = self.good & ( bigflux > 5.*bigfluxerr ) & ( smallflux > 5.*smallfluxerr )
```
rknop, line 433: Need to leave is_star in here -- we only want to be doing aperture corrections on PSFs.  It's better to not use all the stars (because some got omitted) than include things that aren't stars.  In the former case, it will (ideally) just increase the uncertainty.  In the latter case, it will systematically bias the aperture correction.

pipeline/detection.py
```diff
@@ -505,7 +505,7 @@ def _run_sextractor_once( self, image, apers=[5, ], psffile=None, tempname=None,
                      "-FILTER", "Y",
                      "-FILTER_NAME", str(conv),
                      "-WEIGHT_TYPE", "MAP_WEIGHT",
-                     "-RESCALE_WEIGHTS", "N",
+                     "-RESCALE_WEIGHTS", "Y",
```
rknop, line 508: I'm pretty sure we want RESCALE_WEIGHTS to be N here.  (See comment above in the code.)  Since we believe our weight images actually reflect the real variances, we don't want to be doing any rescaling of them.

tests/conftest.py
```diff
@@ -152,7 +152,7 @@ def rnd_str(n):
     return ''.join(np.random.choice(list('abcdefghijklmnopqrstuvwxyz'), n))
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture(scope='session', autouse=True)
```
rknop, line 155: I thought scope='session' was the default...

tests/fixtures/pipeline_objects.py
```diff
@@ -191,10 +313,24 @@ def make_datastore(*args, cache_dir=None, cache_base_name=None, session=None):
             if ds.image is None:  # make the preprocessed image
                 _logger.debug('making preprocessed image. ')
                 ds = preprocessor.run(ds)
+                if bad_pixel_map is not None:
+                    ds.image.flags |= bad_pixel_map
+                    if ds.image.weight is not None:
+                        ds.image.weight[bad_pixel_map.astype(bool)] = 0.0
```
rknop, line 319: You want to use ds.image.flags here rather than bad_pixel_map, as the former is the one that's bitwise ORed both of them together.

tests/models/test_source_list.py
```diff
@@ -233,13 +233,21 @@ def test_write_sextractor():
 def test_calc_apercor( decam_datastore ):
     sources = decam_datastore.get_sources()
 
+    # These numbers are when you don't use is_star at all:
```
rknop, line 236: ...but we do want to use is_star (see comment there).

tests/pipeline/test_extraction.py
```diff
@@ -147,7 +147,7 @@ def test_sextractor_extract_once( decam_datastore, extractor ):
         extractor.pars.test_parameter = uuid.uuid4().hex
         sourcelist, sourcefile = run_sextractor(decam_datastore.image, extractor)
 
-        assert sourcelist.num_sources == 5611
+        assert sourcelist.num_sources == 5623
```
rknop, line 150: I suspect these changes (and the ones below) came in because you set the RESCALE_WEIGHTS parmeter in sextractor to Y instead of N.  Unless you know why we want to use that, we should keep that at N, which would mean reverting these changes.  (That will probably also allow us to put the tolerances in test_astro_cal.py back to 0.01/3600. degrees instead of 0.05/3600.  In any event, I think we want to leave that tolerance very small, because it's checking to make sure that we got the WCS solution we expected to.)

tests/pipeline/test_extraction.py
```diff
@@ -189,11 +189,11 @@ def test_sextractor_extract_once( decam_datastore, extractor ):
         assert sourcelist.info['SEXBKGND'] == pytest.approx( 179.8, abs=0.1 )
         assert sourcelist.x.min() == pytest.approx( 16.0, abs=0.1 )
         assert sourcelist.x.max() == pytest.approx( 2039.6, abs=0.1 )
-        assert sourcelist.y.min() == pytest.approx( 16.3, abs=0.1 )
+        assert sourcelist.y.min() == pytest.approx( 16.0, abs=0.1 )
         assert sourcelist.y.max() == pytest.approx( 4087.9, abs=0.1 )
         assert sourcelist.apfluxadu(apnum=1)[0].min() == pytest.approx( -656.8731, rel=1e-5 )
         assert sourcelist.apfluxadu(apnum=1)[0].max() == pytest.approx( 2850920.0, rel=1e-5 )
-        assert sourcelist.apfluxadu(apnum=0)[0].min() == pytest.approx( 89.445114, rel=1e-5 )
+        assert sourcelist.apfluxadu(apnum=0)[0].min() == pytest.approx( 84.74182, rel=1e-5 )
```
rknop, line 196: This change (and some of the others, including aperture sizes) will be a change from both RESCALE_WEIGHTS and using is_star in the aperture flux correction (both changes I think should be reverted).

rknop (changes requested):
Mostly small stuff, but two of the small comments I think are very important:

* We do want to have the is_star qualifier in the aperture correction.  Even though our star/galaxy separator isn't great, at least in a star-heavy field it's missing stars rather than admitting too many things.  As such, the sample used to calculate aperture corrections is far more polluted with non-PSFs without the is_star qualifier than with it, so we're going to get a more accurate aperture correction measurement using is_star.

* Unless you've figured out what RESCALE_WEIGHTS IN sextractor does (in which place, please replace the comment I have about how I'm not sure what it is), we should leave it to N rather than Y.  The reason is, we believe our weight images do really reflect the true variances in the images, and aren't just relative weight images.  As such, we don't want to be doing any rescaling of them (which we would if they were just relative weight images).

Reverting both of these changes will revert a whole bunch of the test result changes you had to make.

tests/conftest.py
guynir42: I think the default is scoped to a function. Session scope is the "biggest" one.

models/source_list.py
guynir42: I am leaving this out and will fix it when addressing issue #154

PR #157, opened 2023-12-15
Alignment as part of Image

guynir42:
Alignment is carried out inside the Image model, instead of as a pipeline process step. 

This means that when you are about to create a coadd or a subtraction, you'd implicitly call the ImageAligner inside the Image model to align all the images (either to the "new" or one of the images marked as "ref"). 

This allows us to recreate the aligned images from the coadd/subtraction when needed, without having to save them in the database or on disk.


guynir42:
@rknop this is now merged and ready for review.

models/image.py
```diff
@@ -107,7 +108,7 @@ def format(self, value):
     )
 
     ref_image_index = sa.Column(
-        sa.Integer,
+        sa.SMALLINT,
```
rknop, line 111: I think the reason I'd rather have the image id here instead of the image index is so that we could figure out the reference using pure SQL, without having to rely on intermediate processing in python to do it.

(We probably could do it in pure SQL by sorting by MJD, but it makes the SQL more complicated and more fraught.)

(No need to change this here; you  may have already made an Issue for this.  Something to think about for the future.)

models/image.py
```diff
@@ -635,6 +642,69 @@ def from_exposure(cls, exposure, section_id):
...
+            'bkg_rms_estimate',
+            'ref_image_index',
+            'new_image_index',
+            'is_coadd',
+            'is_sub',
+            '_bitflag',
+            '_upstream_bitflag',
+            '_format',
+            '_type',
+        ]
```
rknop, line 690: Probably add a "not copied attributes" array here, and then raise an exception if there are any attributes left over after copy_, simple_, and not_copied_.  (Reason: if we add/remove fields to the Image model later, we also have to edit this function.  Without something that raises an exception in this function, we could easily not remember to do that, and then this function will silently be doing the wrong thing without us realizing it.)

models/image.py
```diff
@@ -635,6 +642,69 @@ def from_exposure(cls, exposure, section_id):
 
         return new
 
+    @classmethod
+    def copy_image(cls, image):
+        """Make a new Image object with the same data as an existing Image object.
+
+        This new object does not have a provenance or any relationships to other objects.
+        It should be used only as a working copy, not to be saved back into the database.
+        The filepath should also be set to a new (unique) value so as not to overwrite the original.
```
rknop, line 651: If I read the function correctly, the filepath will be empty.  Perhaps document that in the docstring.

models/image.py
```diff
@@ -813,6 +883,54 @@ def from_new_and_ref(cls, new_image, ref_image):
...
         return output
 
+    def align_images(self):
+        """Align the upstream_images to one of the images pointed to by image_index.
+
+        The parameters of the alignment must be given in the parameters attribute
+        of this Image's Provenance.
+
+        The index to which the images are aligned is given by the parameters, using
+        the "to_index" key, which can be "first" or "last".
```
rknop, line 893:
```suggestion
        The index to which the images are aligned is given by the "to_index" key in the
        "alignment" dictionary in the parameters of the image provenance; the value can
        be "first" or "last".
```

models/image.py
```diff
@@ -813,6 +883,54 @@ def from_new_and_ref(cls, new_image, ref_image):
...
+
+        The resulting images are saved in _aligned_images, which are not saved
+        to the database. Note that each aligned image is also referred to by
+        a global variable under the ImageAligner.temp_images list.
+        """
+        from pipeline.alignment import ImageAligner  # avoid circular import
```
rknop, line 899: If the point of this PR is to move image alignment out of the pipeline and into the image model, then the ImageAligner probably really belongs in the util subdirectory rather than the pipeline directory.

(I suspect we have other cases of things in model importing things in pipeline.)

This is the sort of thing that could go to a new Issue rather than being modified in this PR.

models/image.py
```diff
@@ -813,6 +883,54 @@ def from_new_and_ref(cls, new_image, ref_image):
...
+            image_index = -1
+
+        if self._aligner is None:
+            self._aligner = ImageAligner(**self.provenance.parameters['alignment'])
+        else:
+            self._aligner.pars.override(**self.provenance.parameters['alignment'])
+
+        # verify all products are loaded
+        for im in self.upstream_images:
+            if im.sources is None or im.wcs is None or im.zp is None:
```
rknop, line 917: Does alignment actually need the zp?  I believe it only really uses sources.  (It may not even use the wcs.)

pipeline/data_store.py
```diff
@@ -1382,8 +1382,8 @@ def delete_everything(self, session=None):
             autoflush_state = session.autoflush
             try:
                 obj_list = self.get_all_data_products(output='list', omit_exposure=True)
-                for obj in obj_list:  # first make sure all are merged
-                    obj = obj.recursive_merge(session)
+                for i, obj in enumerate(obj_list):  # first make sure all are merged
+                    obj_list[i] = obj.recursive_merge(session)
```
rknop, line 1386: Aha!  This is one of those fixes I see and think "duh, of course"... but python sometimes lulls us into a false sense of security with all its high-level syntactic sugar.  Teaching computational physics to students who aren't experienced programmers, and haven't done anything in a lower-level language like C before, it can be a little bit tricky to explain what's going on here.

tests/pipeline/test_alignment.py
```diff
@@ -19,11 +26,19 @@ def test_warp_decam( decam_datastore, decam_reference ):
 
         # Check a couple of spots on the image
         # First, around a star:
+# <<<<<<< HEAD
```
rknop, line 29: Looks like you have a merge conflict baked in here....

...or is this here to remember that at one point we used an abs cutoff rather than a rel, and we're not 100% sure which is what we really want?  If so, probably make the comments reflect that.

models/base.py
```diff
@@ -1243,23 +1247,39 @@ def remove_data_from_disk(self, remove_folders=True):
...
+                downstreams = self.get_downstreams()
+                for d in downstreams:
+                    if hasattr(d, 'remove_data_from_disk'):
+                        d.remove_data_from_disk(remove_folders=remove_folders, remove_downstream_data=True)
 
-    def delete_from_disk_and_database(self, session=None, commit=True, remove_folders=True):
+            except NotImplementedError as e:
+                pass  # if this object does not implement get downstreams, it is ok
+
+    def delete_from_disk_and_database(
```
rknop, line 1280: Is there a test that verifies that when this is called with remove_downstream_data=True, the downstreams get deleted from disk, database, and archive?  (If not, we should write one; make an Issue.)

rknop (approved):
Mostly minor comments.  A couple of things where we might want to create an Issue, a few suggested changes to comments.  Go ahead and merge when you've done what should be done with those.

One thing that might be worth addressing is the copy_image class method of the Image class.  It's another place where if we make changes to a model, we also have to make changes to code to keep it synced with the model.  It would probably be worth putting in a check that will throw an exception if the Image model doesn't look like what the code in copy_images is assuming.

models/image.py
guynir42: I'm not sure how to do that. I don't want to simply go over all attributes in `__dict__` so I don't know if there's an automated way to verify we are copying all the stuff we mean to copy. I will add an issue to deal with this.

models/base.py
guynir42: I made a new issue.

PR #158, opened 2023-12-16
Spalipy as an option for aligning images

guynir42:
This is part of the effort to replace command line executables with python scripts. 

Spalipy should be able to align images based on the source lists in memory.

Right now it doesn't succeed in matching the stars from the reference to the image (for reasons I'm not sure about). 

More work is needed to see if we can get it to replace swarp.


guynir42:
This doesn't work and I don't have time to keep fighting with it. Closing for now.

PR #159, opened 2023-12-18
Coaddition object

guynir42:
Add the object in charge of actually coadding images in a proper way. 
Currently I'm going to implement ZOGY (well, Zackay & Ofek is the correct term but I will call it ZOGY for simplicity). 
The method involves coadding the PSF matched-filtered images and then dividing by a coadd of the PSFs. 
reference: https://ui.adsabs.harvard.edu/abs/2017ApJ...836..188Z/abstract

This PR leans heavily on the alignment PR #157

What we need for this to work:
- [x] Use the `Image.from_images` and the Image's ability to align its `upstream_images` to get the source material. 
- [x] Run the ZOGY coaddition on the images themselves, keeping track of the image and PSF stacks. 
- [x] Run ZOGY on the weight image (with the understanding that NaNs or bad pixels in the image have weight zero). The final image must come with this weighted image that tells user which pixels have lower depth (higher noise). 
- [x] Allow a tunable parameter that tells what fraction of the FWHM of the PSF to consider the "splash zone" for bad pixel flags. This is used to expand (or dilate) each bit in the bitmask. 
- [x] Test against PTF images. Make sure the final PSF is sharper than using the naive method. If we implement a naive method, it would simply add together all the different datasets like the image, weight, flags, etc. 
- [ ] Make some adjustments to the image's raw header that is written to disk. This should include things like the MJD start and finish and exposure time, but also the filepaths of the upstream images and the index that was used to align them. 
- [x] Naming convention for coadded images (would that be the same for a subtracted image? probably not, it would get the name of the new image and a hash of the upstream_images which could be the same as the coadd case).


guynir42:
@rknop this is ready for review.

models/image.py
```diff
@@ -776,17 +789,17 @@ def from_images(cls, images):
         output.end_mjd = max([image.end_mjd for image in images])  # exposure ends are not necessarily sorted
 
         # TODO: what about the header? should we combine them somehow?
```
rknop, line 791: Maybe add a header HISTORY or COMMENT that lists all the images that were combined?

Or, a series of keywords COMBIMn, where n is an increasing integer (and corresponds directly to the indexes of the combined images), and the value of each is the image ID.  (With a comment indicating that.)

models/psf.py
```diff
@@ -150,6 +150,11 @@ def info( self ):
     def info( self, value ):
         self._info = value
 
+    @property
+    def image_shape(self):
+        """The shape of the image this PSF is for."""
+        return self.header['NAXIS2'], self.header['NAXIS1']
```
rknop, line 156: This isn't right.  For PSFs from PSFex, the header is the header of the .psf FITS file written by PSFex (or so is documented in the "header" property), and the NAXIS1/NAXIS2 there are the size of the FITS data in the PSF, which is very different from the image that the PSF is measured for.

`self.image.header['NAXIS2'], self.image.header['NAXIS1']` should do it.

models/psf.py
```diff
@@ -387,6 +387,13 @@ def get_clip( self, x, y, flux, norm=True, noisy=False, gain=1., rng=None, dtype
         if self.format != 'psfex':
             raise NotImplementedError( "Only know how to get clip for psfex PSF files" )
 
+        if x is None:
+            # x = self.image_shape[1] / 2.
+            x = 0  # I think the x/y are centered on the center of the image
```
rknop, line 392: No... the x and y are the actual image x and y coordinates.  (See docstring on `get_resampled_psf`.) PSFex stores a couple of values `POLZERO1` and `POLZERO2` that are the zero offsets of the polynomial it fits (in FITS coordinates, so 1-offset), and `get_resampled_psf` uses this.

models/ptf.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, **kwargs):
         self.read_noise = 4.5
         self.dark_current = 0.1
         self.gain = 1.5
-        self.saturation_limit = 50000
+        self.saturation_limit = 20000
         self.non_linearity_limit = 50000
```
rknop, line 23: ...a nonlinearity limit higher than the saturation limit is... weird.

pipeline/coaddition.py
```diff
@@ -0,0 +1,423 @@
...
+
+class ParsCoadd(Parameters):
+    def __init__(self, **kwargs):
+        super().__init__()
+
+        self.method = self.add_par(
+            'method',
+            'zogy',
+            str,
+            'Coaddition method.  Currently only "naive" and "zogy" are supported. ',
```
rknop, line 25: We should add an issue to add the swarp coaddition method.  (In that case, alignment and coaddition can be done as a single step.)

pipeline/coaddition.py
```diff
@@ -0,0 +1,423 @@
...
+    that share a provenance hash, their respective downstreams (e.g., PSF) should also have a single provenance
+    hash for each type of downstream.  This makes it possible to identify the upstream images' associated products
+    based solely on the provenance's upstream hashes.
+
+    Areas on the edges where the images are not overlapping (or bad pixels, etc.) are coadded but will
+    contribute zero weight, so the total weight of that pixel will be zero (if all input images have a bad pixel)
+    or they would have lower weight if only some images had bad pixels there.
+
+    Remember that coaddition uses convolution with the PSF, so any effects of individual pixels could affect nearby
+    pixels, depending on the size of the PSF.
```
rknop, line 88: ...this may be true for the zogy coaddition method, but won't be true for all of them.  (For isntance, the naive coaddition method doesn't do this.)

pipeline/coaddition.py
```diff
@@ -0,0 +1,423 @@
...
+        """Perform the core Zackay & Ofek proper image coaddition on the input data cube.
+
+        Parameters
+        ----------
+        datacube: ndarray
+            The data cube to coadd. Can be images or weight maps (or anything else).
+        psfcube: ndarray
+            The PSF cube to use for coaddition.
+        sigmas: ndarray
+            The noise estimate for each image in the data cube.
```
rknop, line 185:
```suggestion
            The sky noise estimate for each image in the data cube.
```

pipeline/coaddition.py
```diff
@@ -0,0 +1,423 @@
...
+            raise ValueError('The flux_zps array must have the same length as the first axis of the data cube. ')
+
+        if np.sum(np.isnan(datacube)) > 0:
+            raise ValueError('There are NaNs values in the data cube! Use inpainting to remove them... ')
+
+        # calculations:
+        datacube_f = fft2(datacube)
+        psfcube_f = fft2(psfcube)
+
+        score_f = np.sum(flux_zps / sigmas ** 2 * np.conj(psfcube_f) * datacube_f, axis=0)  # eq 7
```
rknop, line 225: eq 7 of what?  (Probably put the paper reference in the function docstring.)

rknop (commented):
Lots of good work here.  Bunch of comments, none of them terribly earth-shattering.

improc/alignment.py
```diff
@@ -279,62 +334,66 @@ def run( self, source_image, target_image ):
...
-        #         source_zp.provenance,
-        #         target_image.provenance,
-        #         target_sources.provenance,
-        #         target_wcs.provenance,
-        #     ],  # this does not really matter since we are not going to save this to DB!
-        # )
+            warped_image.type = 'Warped'
+            warped_image.psf = source_image.psf
+            warped_image.zp = source_image.zp
+            # TODO: are SourceList and WorldCoordinates also included? Are they valid for the warped image?
```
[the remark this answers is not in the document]
guynir42: So I should make an issue about re-measuring the PSF after alignment?

improc/alignment.py
```diff
@@ -279,62 +334,66 @@ def run( self, source_image, target_image ):
...
+                source_sources.provenance,
+                source_wcs.provenance,
+                source_zp.provenance,
+                target_image.provenance,
+                target_sources.provenance,
+                target_wcs.provenance,
+            ],  # this does not really matter since we are not going to save this to DB!
+        )
+        warped_image.provenance.update_id()
+        warped_image.header['original_image_id'] = source_image.id  # add this only for aligned images for verification
```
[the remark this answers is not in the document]
guynir42: This is not a fits header, it is the JSONB where we dump all the stuff we want to have accessible on the DB level but not an independent column. I deliberately use lower case keywords here, to tell it apart from the raw header.

improc/inpainting.py
```diff
@@ -0,0 +1,256 @@
...
+            'median',
+            str,
+            'Method to rescale images before interpolating between them. '
+            'Can be "median" (default) or "sigma_clipping" or "none". ',
+            critical=True
+        )
+
+        self.single_image_method = self.add_par(
+            'single_image_method',
+            'biharmonic',
```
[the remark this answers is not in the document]
guynir42: This doesn't really matter. I'm adding a flag to all interpolated pixels. The reason I want to put "something reasonable" is that the Fourier methods used in the coaddition/subtraction can cause big artifacts if we leave sharp drops like "replace NaNs with zeros" would do.

models/image.py
guynir42: I've made an issue about this (#172 ) we should discuss.

models/psf.py
guynir42: I wanted to allow the PSF object to independently know the size of the image it is related to, without having to load it from the database (and indeed, load it from file to read the header). 
I'm not sure if this is necessary anymore as I get a fairly good PSF when just requesting 0,0 as the coordinates.

models/psf.py
guynir42: So what you are saying is that, for the "representative" PSF of the image, I should default the x,y coordinates to `float( self.header['POLZERO1'] ) - 1` and `POLZERO2`?

pipeline/coaddition.py
guynir42: Issue #184

pipeline/coaddition.py
```diff
@@ -0,0 +1,423 @@
...
+
+        return outim, outwt, outfl
+
+    def _zogy_core(self, datacube, psfcube, sigmas, flux_zps):
+        """Perform the core Zackay & Ofek proper image coaddition on the input data cube.
+
+        Parameters
+        ----------
+        datacube: ndarray
+            The data cube to coadd. Can be images or weight maps (or anything else).
```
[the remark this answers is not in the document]
guynir42: I'm not sure that is important. If you coadd with the background you just get a high background on the output image.

pipeline/coaddition.py
```diff
@@ -0,0 +1,423 @@
...
+        psf_fwhms: list of floats
+            The FWHM of the PSF for each image.
+            If images is given as Image objects, can be left as None.
+        flux_zps: list of floats
+            The flux zero points for each image.
+            If images is given as Image objects, can be left as None.
+        bkg_means: list of floats
+            The mean background for each image.
+            If images is given as Image objects, can be left as None.
+            This variable can be used to override the background estimation.
```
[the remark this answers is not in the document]
guynir42: I think there are already some issues dealing with background subtraction. 
This code lets you do the quick and dirty thing if you want, or just supply the images after background is removed.

pipeline/coaddition.py
```diff
@@ -0,0 +1,423 @@
...
+        psfcube = np.array(psfs)
+        bkg_means = np.reshape(np.array(bkg_means), (len(bkg_means), 1, 1))
+        bkg_sigmas = np.reshape(np.array(bkg_sigmas), (len(bkg_sigmas), 1, 1))
+        flux_zps = np.reshape(np.array(flux_zps), (len(flux_zps), 1, 1))
+
+        # subtract the background
+        imcube -= bkg_means
+
+        # make sure to inpaint missing data
+        imcube = self.inpainter.run(imcube, flcube, wtcube)
```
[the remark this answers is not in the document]
guynir42: I have this vague memory of doing something like that. Maybe I added it in the next PR or something. I will have to check.

pipeline/coaddition.py
```diff
@@ -0,0 +1,423 @@
...
+
+        # This is where the magic happens
+        outim, psf, score = self._zogy_core(imcube, psfcube, bkg_sigmas, flux_zps)
+
+        # coadd the variance as well
+        varflag = wtcube == 0
+        wtcube2 = wtcube ** 2
+        wtcube2[varflag] = np.nan
+        varmap = 1 / wtcube2
+        varmap = self.inpainter.run(varmap, varflag, wtcube)  # wtcube doesn't do anything, maybe put something else?
```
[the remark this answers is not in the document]
guynir42: I'm leaving the weights because I don't know what kind of inpainting we will want to do in the future, and I expect images to always have weights anyway. To allow weights=None means figuring out all the ways the inpainting will have to deal with that missing data. I don't think it is worth it.

guynir42:
@rknop I've addressed all comments. Use "see changes from your last review" if you want to focus on the new things. 
If the tests all pass and you are happy, I will merge this into the next PR.

improc/alignment.py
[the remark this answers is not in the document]
rknop: Sooner or later we're probably going to regret having both "header" and "raw_header" as we become confused....

pipeline/coaddition.py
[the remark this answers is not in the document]
rknop: Let me know if you find it.  If not, at the very least, add a TODO in comments here to indicate that it's implicitly assuming the images are all scaled the same, and that we still need to do something about different zeropoints.

models/psf.py
rknop: It's not immediately obvious, but I would probably use the center of the image.  The POLZERO1 nad POLZERO2 aren't necessarily the positions of the "most average" estimate, it's just what it decided to use in its fit.  The commented-out image.shape[n] / 2. are what I'd put in.

pipeline/coaddition.py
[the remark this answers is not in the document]
rknop: But the inpainting will get screwed up if the images aren't background subtracted (unless they all happen to have exactly the same sky background).

rknop (changes requested):
Added a few comments.  This is what I'd change:

* the defaults for position in `get_clip` in `models/psf.py` should be `self.image_shape[n]/2.`, the center of the image.  (See comment.)
* Either remove the function `image_shape` in `psf.py`, or fix it so it does what it says.  Right now, what it's doing isn't what it claims to do.
* In `coaddition.py` function `_coadd_zogy`, see comment on line 353 re: adding a TODO for handling image scaling if you didn't actually already do that in a later PR.

improc/alignment.py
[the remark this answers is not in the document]
guynir42: Should we replace "header" with "additional_data" or something like that?

models/psf.py
guynir42: I assumed we want the center of the image. This is why I've added the size of the image in PSF to begin with. 
But when I set the x,y to the center of the image I got really bad PSFs (sort of donut shaped) and everything failed. 

Now that it is set to zero, I get nice PSFs. I think that those coordinates are not what you think they are.

guynir42: I'm leaving it as-is and making an issue to investigate this.

pipeline/coaddition.py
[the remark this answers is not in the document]
guynir42: Yes I am thinking about that, too. You can't rescale images to both the flux zero point and to the background mean. I guess we will need to enforce background subtraction for coadds as well as for subtractions. I'm leaving it like this and making an issue to deal with this later.

guynir42:
@rknop I've made issues for what I'm not sure about and fixed some more stuff. 
The PSF `image_size` used the wrong header keyword, I think that may be part of the problem I've had, but I've put that to another issue. 
Let me know if we can merge.

models/psf.py
rknop: There must be something wrong with the psf solution to your image.

I'm very sure that the coordinates in `get_resampled_psf` are what I think they are.  I've been using versions of that function for a long time.  Look at the code....

But, also, look at `test_psf_rendering` in `tests/models/test_psf.py`.  Uncomment the things that create images (which we should probably replace with an if on whatever conditional variable you've created for interactive tests).  Run the test, and then look first at `/seechange/data/resamp.fits`, which was created with `resamp = psf.get_resampled_psf( 512., 512., dtype=np.float32 )` on a 1024×1024 image.  Next look at `model.fits`, `resid.fits`, and `data.fits`.  data is fake data, with stretched gaussians dropped down. model was built using `PSF.add_psf_to_image`, feeding it the image positions of the psfs that were dropped down, which in turn calls `get_clip` using those image positions.

The problem is somewhere else, not in get_clip.  The best thing to put if x and y are non is the center of the image, not the corner of the image.

What image, and what sequence of functions/commands, did you do that gave you a bad-looking PSF?

rknop:
Ah!  That is exactly what the problem was.  The NAXIS1 of the PSF imge is 15000 (with NAXIS2 of 1), so you were getting the PSF at y=7500, extrapolating the fit way off of the edge of the image.

I just verified that with one of the images you get from the `ptf_reference_images` fixture, running `get_clip(2048,1024)` gives a resonable-looking psf, not a donut.

models/psf.py
```diff
@@ -387,6 +387,13 @@ def get_clip( self, x, y, flux, norm=True, noisy=False, gain=1., rng=None, dtype
         if self.format != 'psfex':
             raise NotImplementedError( "Only know how to get clip for psfex PSF files" )
 
+        if x is None:
+            x = self.image_shape[1] / 2.
+            # x = 0  # I think the x/y are centered on the center of the image
```
rknop, line 392:
```suggestion
```

models/psf.py
```diff
@@ -387,6 +387,13 @@ def get_clip( self, x, y, flux, norm=True, noisy=False, gain=1., rng=None, dtype
...
+        if y is None:
+            y = self.image_shape[0] / 2.
+            # y = 0  # I think the x/y are centered on the center of the image
```
rknop, line 395:
```suggestion
```

rknop (approved):
I'm 100% confident that (0,0) is not the center of the image in `PSF.get_clip`, so we may as well remove the commented lines.  After that, feel free to merge.

PR #171, opened 2023-12-27
Coaddition Pipeline

guynir42:
The coaddition pipeline is similar to the top-level pipeline. 
It can be initialized from the config file or using overrides from `kwargs`. 
It then accepts either a list of images or a time range (with both start and end date optional, we can set reasonable defaults for them) + a filter, section ID and coordinates (or target).
If using time range/filter/coordinates it will look for all the images within that range and pull them up to make the coadd. 

The pipeline will output a coadded image that has source extraction, PSF extraction, astro/photo calibration all done on it. 

It can then be saved either as a combined science image (e.g., nightly/weekly science coadds) or as a reference image.


guynir42:
@rknop this is ready for review.

default_config.yaml
```diff
@@ -123,6 +123,37 @@ astromatic:
...
+      multi_image_method: median
+      feather_width: 2
+      rescale_method: median
+      single_image_method: biharmonic
+      ignore_flags: 0
+  extraction:
+    measure_psf: true
+    threshold: 3.0
+    method: sextractor
+  astro_cal:
```
rknop, line 144: For the things that also have general settings (extraciton, astro cal, photo cal), it'd be good if in the config we had a way of saying "just use the general settings for coadd".  It's nice to have the ability to customize it and use something different for coadd, but it'll be a pain to have to change things in two places if you do want to use the same parameters.

improc/alignment.py
```diff
@@ -392,8 +392,10 @@ def run( self, source_image, target_image ):
                 target_wcs.provenance,
             ],  # this does not really matter since we are not going to save this to DB!
         )
-        warped_image.provenance.update_id()
-        warped_image.header['original_image_id'] = source_image.id  # add this only for aligned images for verification
+        warped_image.provenance_id = warped_image.provenance.id  # make sure this is filled even if not saved to DB
+        warped_image.header['original_image_id'] = source_image.id
```
rknop, line 396: I liked the "additional_data" suggestion you had on the previous PR, or some other name besides "header" (to avoid confusion with raw_header).  Make this an Issue if you haven't already, because it's not worth trying to find and fix all of the instances of header in this current PR.

pipeline/astro_cal.py
```diff
@@ -234,7 +238,7 @@ def _run_scamp( self, ds, prov, session=None ):
                 _logger.info( f"Failed to solve for WCS with maxmag {maxmag}, trying the next one." )
 
         if not success:
-            raise RuntimeError( "_run_scamp failed to find a match." )
+            raise RuntimeError( f"_run_scamp failed to find a match. Exceptions that were raised: {exceptions}" )
```
rknop, line 241: This could potentially be a very long error message....

pipeline/coaddition.py
```diff
@@ -404,13 +422,18 @@ def run(self, images):
...
+
         output.is_coadd = True
         output.new_image = None
 
+        if aligned_images is not None:
+            output.aligned_images = aligned_images
+
         if self.pars.method == 'naive':
-            outim, outwt, outfl = self._coadd_naive(images)
+            outim, outwt, outfl = self._coadd_naive(output.aligned_images)
```
rknop, line 434: If you leave the aligned_images parameter to run as None, how does output.aligned_images get filled?  It looks to me like it doesn't.

pipeline/coaddition.py
```diff
@@ -424,3 +447,171 @@ def run(self, images):
...
+        self.coadder = Coadder(**coadd_config)
+
+        # source detection ("extraction" for the regular image!)
+        extraction_config = self.config.value('coaddition.extraction', {})
+        extraction_config.update(kwargs.get('extraction', {'measure_psf': True}))
+        self.pars.add_defaults_to_dict(extraction_config)
+        self.extractor = Detector(**extraction_config)
+
+        # astrometric fit using a first pass of sextractor and then astrometric fit to Gaia
+        astro_cal_config = self.config.value('coaddition.astro_cal', {})
```
rknop, line 487: The comment I made above in the config yaml file.  You could put a line here that if astro_cal_config is (say) the string "default", then run astro_cal_config = self.config.value( 'astro_cal', {} )

pipeline/coaddition.py
```diff
@@ -424,3 +447,171 @@ def run(self, images):
...
+
+        self.datastore = None  # use this datastore to save the coadd image and all the products
+
+        self.aligned_images = None  # use this to pass in already aligned images
+
+    def parse_inputs(self, *args, **kwargs):
+        """Parse the possible inputs to the run method.
+
+        The possible input types are:
+        - a list of Image objects
```
rknop, line 506:
```suggestion
        - a list of Image objects
        - two lists of Image objects (where the second list are aligned images corresponding to the first list)
```

pipeline/coaddition.py
```diff
@@ -424,3 +447,171 @@ def run(self, images):
...
+            args = args[0]
+        # check if any remaining args are not Image objects
+        for arg in args:
+            if not isinstance(arg, Image):
+                raise ValueError('All unnamed arguments must be Image objects. ')
+
+        images = args
+
+        if len(images) > 0:
+            return images  # TODO: maybe instead allow the kwargs to filter down the list of Images?
```
rknop, line 551: It seems a little weird to store the aligned_images in self, but then return images and require the caller to store in self.  Why not just store the images in self.images instead of returning the images, so that the two are treated the same way?

pipeline/coaddition.py
```diff
@@ -424,3 +447,171 @@ def run(self, images):
...
+
+        # TODO: this feels like it could be a useful tool, maybe need to move it Image class?
```
rknop, line 553: Agreed.

rknop (approved):
This one seems pretty straightforward (given the pipeline structure you put together long ago).

A few smallish comments; feel free to either address them here before merging, or make an Issue as appropriate.

The only comment I'm a little worired about is the one on line 434 of coaddition.py.  (It's not clear how output.aligned_images gets set if you don't pass aligned_images.)

default_config.yaml
guynir42: I've made the coaddition pipeline load the generic parameters first, and override them with the "coaddition" keys.

improc/alignment.py
guynir42: Created an issue.

pipeline/astro_cal.py
guynir42: Better to have a long error than nothing.

pipeline/coaddition.py
guynir42: The `output` image is generated from a list of upstream images, and has, in its provenance, an "alignment" parameter dictionary. Using those the output image can lazy-calculate aligned images. 
The ability of the pipeline to accept aligned images as an optional input is to skip this calculation and save runtime if you already have the aligned images (e.g., when you cache them in tests).

pipeline/coaddition.py
guynir42: added an issue for this.

pipeline/coaddition.py
guynir42: Agreed.

PR #174, opened 2023-12-31
Subtraction Pipeline

guynir42:
Integrate the subtraction algorithm (currently ZOGY) with the pipeline in general. 

This includes: 
- [x] Make sure the reference image gets loaded automatically for a new image. 
- [x] Create a subtraction image from the reference and new, including the alignment parameters. 
- [x] Run the chosen subtraction algorithm and make sure the outputs are saved in the subtraction image. 
- [ ] Provide a datastore with the subtraction image already created, for the next part of the algorithm.


guynir42:
@rknop there's more work to do on subtraction, and cleanup of some of the other issues we've found along the way, but I think this stands on its own for review. 

We are officially back to the same timeline and can return to normal working procedures...

models/provenance.py
```diff
@@ -294,7 +294,7 @@ def update_id(self):
         superdict = dict(
             process=self.process,
             parameters=self.parameters,
-            upstream_hashes=self.upstream_hashes,
+            upstream_hashes=self.upstream_hashes,  # this list is ordered by upstream ID
```
rknop, line 297: Do we want to order it by upstream ID, or by the order they're stored in the index fields?

pipeline/data_store.py
```diff
@@ -136,6 +136,11 @@ def parse_args(self, *args, **kwargs):
         Can initialize based on exposure and section ids,
         or give a specific image id or coadd id.
 
+        If given an Image that is already loaded with related products
+        (SourceList, PSF, etc.) then these will also be added to the
+        datastore's attributes, to be checked against Provenance in the
+        usual way when the relevant getter is called (e.g., get_sources).
```
rknop, line 142: With all the related products being stored in the Image object, I'm less clear on what the data_store does for us.  I know it has the various methods for saving things and adding them to the database, but at this point might it make more sense just to put all of those in Image?

pipeline/subtraction.py
```diff
@@ -36,6 +44,79 @@ def __init__(self, **kwargs):
...
+        outwt: np.ndarray
+            The weight image for the difference
+        outfl: np.ndarray
+            The flag image for the difference
+        score: np.ndarray
+            The ZOGY score image (the matched-filter result)
+        zogy_psf: np.ndarray
+            The ZOGY PSF image (the matched-filter PSF)
+        """
+        pass
```
rknop, line 118: I thought I remembered seeing a ZOGY implementation in an earlier PR; where was that?  Why not use it here?

rknop: (But, in any event, probably a good idea to throw a NotImplementedError rather than just pass.)

pipeline/subtraction.py
```diff
@@ -47,39 +128,61 @@ def run(self, *args, **kwargs):
...
+                    raise NotImplementedError('Hotpants not implemented yet')
+                elif self.pars.method.lower() == 'zogy':
+                    outim, outwt, outfl, score, psf = self._subtract_zogy(new_image, ref_image)
+                else:
+                    raise ValueError(f'Unknown subtraction method {self.pars.method}')
+
+                sub_image.data = outim
+                sub_image.weight = outwt
+                sub_image.flags = outfl
+                if 'score' in locals():
```
rknop, line 182: This kind of thing always scares me; if later we're editing code above in this function, we could well create a local variable named 'score'.  I'd feel better of the `self._subtract_*` methods returned a dictionary rather than an unnamed tuple, and then we could look for keys in that dictionary.

rknop (approved):
This mostly looks fine.  I think we would be better served to have the subtraction methods return a dictionary rather than an unnamed tuple; see the comment in subtraction.py.

models/provenance.py
guynir42: I don't know what index fields is, in this context. 

I think the Provenance's ID is the only reasonable way to sort these. There could be many provenances (of the same type or different types) that go into this hash, and we must make sure the order of the provenances is consistent, even if they were produced in a different order (as long as they are the same set).

pipeline/data_store.py
guynir42: I thought about this, too. I think the datastore still does some complicated work for us. 
So here's what the Image can do: it can, if it has upstream_images and a provenance, load the correct products for each of the upstream images. It can also align them. 

The datastore can fetch not only upstream images, it can fetch the image for an exposure, data products for a single image (if they exist), the reference for an image. These all require non-trivial provenance logic. I am not sure it is a good idea to swap it for the Image object. I will have to think more on this. 

I am also going to add some failure modes to the data store, adding to the complexity of pushing all of this into the Image.

pipeline/subtraction.py
guynir42: Ahh yes, I was half way into this PR when we started working on the backlog. I will put a not-implemented here and we will have to continue this in a new PR that is less about the data flow and more about the actual subtraction.

pipeline/subtraction.py
guynir42: good idea.

Issue #187, opened 2024-01-11
Rename header attribute


guynir42:
Change the Image's JSONB column from `header` to something else (maybe `additional_info` or just `info`). 

Consider also changing the `raw_header` attribute to `header`. 

Should also change the Exposure naming convention. 

This should be done very carefully as we use both `header` and `raw_header` all over the place.

Issue #179, opened 2024-01-03
What to do about coadding images with different target/section ID


guynir42:
Right now we require that some attributes of the images used as upstreams of a coadded image are all the same. For example: filter, instrument, section_id and target (maybe a few others). 

When making the coadded image (in the `Image.from_images()` method) there's a check that they are all the same and then we can use them to fill the attributes of the coadd image. 
If we allow them to be different, we need to decide:
- what would be put as the value in the coadded image? 
- how do we verify that the coadded images are in fact overlapping in coordinates?

guynir42:
Also we should make `section_id` nullable to allow coadd images that have multiple sections in the mix.

guynir42:
Addressed by #190

Issue #178, opened 2024-01-03
Reference index in the upstream_images list


guynir42:
Consider reverting the identification mode for which image is the reference, from the list index (current design) to using the image ID (which requires data multiplication but is safer as it doesn't depend on list ordering). 

Also, remove the new_image_index as that is only relevant for subtractions, and those only have two upstream images, so we can infer the new image as it is not the ref image.

guynir42:
Addressed by #190

Issue #160, opened 2023-12-18
Unique constraint on (image_id, provenance_id) for products


guynir42:
For Image products like SourceList or ZeroPoint there should only ever be one instance with the same image_id and provenance_id. 

We should add this as a unique constraint but maybe also as an index?

guynir42:
Addressed by #190

PR #190, opened 2024-01-14
Roundup of issues from recent PRs

guynir42:
Rounding up a bunch of semi-related issues into a single PR, to cleanup a bunch of open issues we collected when working on a set of recent PRs. 

- Replace indexing of `upstream_images` for `new_image` and `ref_image` with a single ID of the ref image. Issue #178 
- Replace the `Image.header` with `Image.info` and `Image.raw_header` with `Image.header`. Issue #187 
- Remove limitations when making combined images, such as having different section ID or target. Issue #179 
- Allow images to not have a `section_id`. Also part of Issue #179 
- Add constraint on products that are unique with regards to `image_id` and `provenance_id`. Issue #160 
- Make sure products have their own `provenance_id` as part of their filenames (to make them unique). Issue #146 
- `downstream_images` as a lazy-loaded relationship on Image. Issue #151 
- Remove `recursive_merge()` in favor of regular merge and the new `merge_all()` that runs on Image, SourceList and Reference, because it needs to take special care of the products (which are not relationships!).


guynir42:
In addition, and I don't know why this happened, I started to make some minor adjustments to how we merge things in the data store. That caused a whole bunch of bugs that I have been running after for three days, touching half of our code.

guynir42:
@rknop There are still a couple of test failures, related to object cleanup (not the actual logic of the tests). 
If you have time to review this, you can start, and I will fix the remaining minor issues tomorrow morning.

models/cutouts.py
```diff
@@ -44,6 +52,12 @@ def format(self, value):
         doc="The source list this cutout is associated with. "
     )
 
+    index_in_sources = sa.Column(
```
rknop, line 55: It occurs to me that there's a possible ambiguity here.  cutouts is associate with a source list, but the source lists on the new, ref, and sub will all have different indexes.  (And, it will often happen that there is no source in the ref's source list at this coordinate.)

I believe what's intended here is that this is the index in the sub's source list, and that that is the same thing that the sources_id points to.  We should make that clear in the documentation here.

(Assuming we're doing to do cutouts from warped ref, `and` that the ref was warped to the new, then the positions in all three images (ref, new, sum) will be the same.)

models/image.py
```diff
@@ -111,65 +111,46 @@ def format(self, value):
...
+            "Reference image used to produce this image, in the upstream_images list. "
         )
     )
 
     @property
     def new_image(self):
-        if self.new_image_index is None:
+        """Get the image that is NOT the reference image. This only works on subtractions (with ref+new upstreams)"""
+        image = [im for im in self.upstream_images if im.id != self.ref_image_id]
+        if len(image) == 0 or len(image) > 1:
```
rknop, line 151: If len(image) > 1, we should raise an exception here, rather than return None ; new_image() was called on a coadd rather than a subtraction, so new_image is meaningless.

models/image.py
```diff
@@ -487,8 +471,44 @@ def init_on_load(self):
...
+        new_image = self.safe_merge(session=session)
+        session.flush()
+        if self.sources is not None:
+            self.sources.image = new_image
+            self.sources.image_id = new_image.id
+            self.sources.provenance_id = self.sources.provenance.id if self.sources.provenance is not None else None
+            new_image.sources = self.sources.merge_all(session=session)
+            new_image.wcs = new_image.sources.wcs
+            new_image.zp = new_image.sources.zp
+            new_image.cutouts = new_image.sources.cutouts
```
rknop, line 494: I think this subverts lazy loading.  

I think if self.cutouts is None, I don't think you want to set new_image.cutouts here.  (Same for other fields.)

models/image.py
```diff
@@ -788,8 +806,8 @@ def from_images(cls, images, index=0):
         output.end_mjd = max([image.end_mjd for image in images])  # exposure ends are not necessarily sorted
 
         # TODO: what about the header? should we combine them somehow?
+        output.info = images[index].info
```
rknop, line 809: We should probably do some filtering on info and header just like you did with the attributes.

Practically speaking, when we swarp the images together, it will strip all of the header information out; you can pass an argument to swarp telling it which header keywords to copy from the first image in the list.  We should update the image's header and info attributes after the swarp is done.  (Not sure if it does that right now, will have to look.)

models/image.py
```diff
@@ -1071,7 +1099,7 @@ def invent_filepath(self):
             try:
                 section_id_int = int(self.section_id)
             except ValueError:
-                section_id_int = 0
+                section_id_int = 0  # TODO: maybe replace with a placeholder like 99?
```
rknop, line 1102: If only there were an Int equivalent of NaN, that would be the right thing to use here.

When do we use the section_id_int?  Perhaps we should just embrace that section_id needs to be a string.

models/instrument.py
```diff
@@ -1600,7 +1600,7 @@ def overscan_and_data_sections( self, header ):
         Parameters
         ----------
         header: fits.Header
-          The header of the image in question; use Image.raw_header not Image.header.
+          The header of the image in question; use Image.header not Image.info.
```
rknop, line 1603: With the names .header and .info instead of .header and .raw_header, the comment after the semicolon probably isn't needed any more.  (No harm leaving it in.)

pipeline/catalog_tools.py
```diff
@@ -384,7 +385,13 @@ def fetch_GaiaDR3_excerpt( image, minstars=50, maxmags=22, magrange=None, sessio
...
                     catexp.filepath = dbfile
                     catexp.save( localfile )
-                    session.add( catexp )
+                    existing_catexp = session.scalars(
+                        sa.select(CatalogExcerpt).where(CatalogExcerpt.filepath == dbfile)
+                    ).first()
+                    if existing_catexp is None:
+                        session.add( catexp )  # add if it doesn't exist
+                    else:
+                        catexp = existing_catexp  # use the existing one
```
rknop, line 394: Hmm... I'm a little disturbed by this.  We're saving a catexp, but then returning an existing database entry.  It's entirely possible that the disk file will have changed, but we don't want to do that.

Looking further up in the code, I think that this should never happen; if there's already a db entry with the same filename, it should have been caught by the query above that searches for existing catalogs.

This should really be an exception; if the query on the corners, mag limits, and number of items didn't return anything, but there's an entry in the database that already has the filename of the catalog excerpt we made here, then there's a problem with our filename scheme not being sufficiently unique.

pipeline/data_store.py
```diff
@@ -515,11 +515,11 @@ def get_image(self, provenance=None, session=None):
             # (If all the ids are None, it'll match even if the actual
             # objects are wrong, but, oh well.)
             if (self.exposure_id is not None) and (self.section_id is not None):
-                if ( (self.image.exposure_id != self.exposure_id) or
+                if ( (self.image.exposure_id is not None and self.image.exposure_id != self.exposure_id) or
```
rknop, line 518: What's the reason for this change?  If self.image.exposure_id is None, then shouldn't self.exposure_id also be None?

pipeline/data_store.py
```diff
@@ -1306,65 +1310,100 @@ def save_and_commit( self, exists_ok=False, overwrite=True, no_archive=False,
...
+            'image',
+            'sources',
+            'psf',
+            'wcs',
+            'zp',
+            'sub_image',
+            'detections',
+            'cutouts',
+            'measurements'
+        ]
```
rknop, line 1324: This is another one of those cases where if we change the code somewhere else to add a new thing to the datastore, we have to remember to add an element to this list.

At least, let's put a comment in __init__ that warns of this.  It might also be good to make a class atribute _attributes_to_save or some such that holds this list, which sits at the top of the whole class, and which the comment in __init__ points to.

tests/fixtures/decam.py
```diff
@@ -209,7 +214,8 @@ def decam_raw_image( decam_exposure ):
 def decam_small_image(decam_raw_image):
     image = decam_raw_image
     image.data = image.data[256:256+512, 256:256+512].copy()  # make it C-contiguous
-    return image
+
+    yield image
```
rknop, line 218: Is there any benefit to using `yield` rather than `return` when there's nothing in the function after the `yield`?  (I'd think if anything there's a benefit in using `return`, as it keeps it simpler; you don't have issues of calling it inside other things.)

rknop (changes requested):
Most of these cleanups look good.  I have a few questions and concerns, which are in the comments on the various lines of the code.

models/cutouts.py
guynir42: Agreed. I will add some thing in the docstring to make it clear that the sources refer to the subtraction image. 
Maybe it would even be better to call it "detections"?

models/image.py
guynir42: The problem is that `new_image` is not a function, it is a property. I think calling properties that routinely raise errors is going to be a headache (consider listing all properties of an  object). When it returns none you just know it is not relevant.

models/image.py
guynir42: This is supposed to only set the attributes that were already there. 
You cannot lazy load cutouts or sources without specifying the provenance (which means it never gets lazy loaded, only explicitly loaded). 
In this case you have to do this for the merge to work. It essentially replaces the "recursive merge".

models/image.py
guynir42: I have issue #172 to think about and sort this out.

models/image.py
guynir42: The problem is that if you are using actual integer values (like CCD number) you'd want the (customizable) naming convention to allow you to do things like 01, 02... in the filename, and that only works with integers. 

The only problem is that same naming convention will be incorrect when `section_id=None`. I am not sure what to do about this, hence the TODO.

models/instrument.py
guynir42: yeah, I agree.

pipeline/catalog_tools.py
guynir42: I think I had some issues with this during tests (when the CatalogExcerpt survived the test restart but the files didn't, or the other way around). I'll try to replace it with an exception and see if the tests pass.

pipeline/data_store.py
guynir42: There are some situations (e.g., where Image is loaded from the test cache without the proper links) where it would fail before having a chance to assign all the new relationships.

tests/fixtures/decam.py
guynir42: Some of the fixtures are nested, and if you have a return instead of a yield, it messes up the cleanup of the containing fixture.

pipeline/data_store.py
guynir42: I've refactored it into two lists at the beginning of the class definition. I still chose to explicitly initialize them in `__init__` because then my IDE doesn't complain that these attributes are defined outside the initialization.

models/cutouts.py
guynir42: I'm leaving it as `sources` because it is a `SourceList`, and using `detections` would make things much more confusing. I'll just add some emphasis in the docstrings.

guynir42:
@rknop I've addressed all comments. Please have another look and see if it is ok to merge (use the "see changes since last review" to save time). Thanks!

rknop (approved):

PR #191, opened 2024-01-21
Archive Cleanup

guynir42:
This PR makes two different setups for communicating with the archive during tests. 

In the local config method, we use `local_read_dir` and `local_write_dir` to define direct folder access to the archive, as you'd do when you were using a local (or mounted) folder to contain the archive storage. 

In the dockerized config method, we use the archive as a server, where `archive_url: http://archive:8080/` is used to communicate with the archive, as you would when, e.g., launching the pipeline on a cloud server or on NERSC. In this case we will re-introduce the environmental variable `SEECHANGE_TEST_ARCHIVE_DIR` to the docker compose file, so that tests can go to the actual storage and see what is going on in the archive when stuff gets added/deleted there. 

In addition, I would also like to verify that 
- [x] the `delete_from_disk_and_database(remove_downstream_data=True)` actually removes the downstream data from the archive, not just from local storage, and 
- [x] that the archive is empty at the end of the test run, just like the temp folder should be.


tests/conftest.py
```diff
@@ -237,26 +259,43 @@ def provenance_preprocessing(code_version):
...
+def archive_path(test_config):
+    if test_config.value('archive.local_read_dir', None) is not None:
+        archivebase = test_config.value('archive.local_read_dir')
+    elif os.getenv('SEECHANGE_TEST_ARCHIVE_DIR') is not None:
+        archivebase = os.getenv('SEECHANGE_TEST_ARCHIVE_DIR')
+    else:
+        raise ValueError('No archive.local_read_dir in config, and no SEECHANGE_TEST_ARCHIVE_DIR env variable set')
+
+    # archive.path_base is usually /test
+    archivebase = pathlib.Path(os.path.join(archivebase, test_config.value('archive.path_base')))
```
rknop, line 272:
```suggestion
    archivebase = pathlib.Path(archivebase) / test_config.value('archive.path_base')
```

pathlib.Path overloads the / operator to append subdirectories and files.  (In general, when using pathlib, you nearly never need to use any of the os.path functions.)

tests/pipeline/test_alignment.py
```diff
@@ -1,127 +0,0 @@
...
-import sqlalchemy as sa
-
-from models.base import SmartSession
-from models.provenance import Provenance
-from models.image import Image
-
-from improc.alignment import ImageAligner
-
-
-def test_warp_decam( decam_datastore, decam_reference ):
```
rknop, line 16: Why are these tests deleted?  Or is this an artifact of the "cleanup" PR not being properly merged?

rknop (commented):
This looks good.  The only question I have is something I suspect will go away when you merge the latest version of the main branch into this branch.

tests/conftest.py
guynir42: I know, we have very different ways of working with paths. I will merge it and then make sure this part makes sense.

rknop: I like the pathlib library, and have never internalized the os.path functions.

I think it's fine that we have both ways of doing things in the code, but there's no need to mix the two in a single line...

rknop (approved):
Looks good.  I'm distressed by the hardcoded dropbox URLs.

What do you mean by the mirror going back up on NERSC?  If we need to archive specific files for the tests, perhaps they should just be committed into the git archive.  If they're too huge for that... hmm, not sure what the best way to do it really is.

guynir42:
@rknop These files used to exist on NERSC. They were also hard-coded but that didn't help. 

Here's the folder: https://portal.nersc.gov/cfs/m2218/decat/decat/templatecache/

It contains several other files, but the specific one we used as a reference for DECam is gone. 

Please check if you can re-upload it there. The files are in the tens-of-megabytes and I don't see a reason to add them to the repo when we are already downloading so many other things on the fly.

PR #195, opened 2024-01-26
Source Detection

guynir42:
Implementing detection of transient sources on the difference image.


guynir42:
@rknop There's one failing test that needs to be updated (the test is wrong, not the code). You can go ahead and review it...

models/image.py
```diff
@@ -152,6 +155,22 @@ def new_image(self):
...
+            raise ValueError('Do not assign None to new_image. Simply clear the upstream_images list.')
+        if not isinstance(value, Image):
+            raise ValueError("The new_image must be an Image object.")
+        if len(self.upstream_images) != 2:
+            raise ValueError("This only works for subtractions with exactly two upstream images.")
+
+        if self.upstream_images[0].id == self.ref_image_id:
+            self.upstream_images[1] = value
+        elif self.upstream_images[1].id == self.ref_image_id:
+            self.upstream_images[0] = value
```
rknop, line 170: Did we pick a convention where if there are two upstream images, the first one is the ref and the second one is the new (or vice versa)?  Or did we decide that the upstream images are unordered, because we have ref_image_id and new_image_id?  Based on this code, I'm assuming we did the latter.

models/image.py
```diff
@@ -405,25 +424,41 @@ def _get_inverse_badness(self):
...
+        # additional data products that could go with the Image
         self.sources = None  # the sources extracted from this Image (optionally loaded)
         self.psf = None  # the point-spread-function object (optionally loaded)
         self.wcs = None  # the WorldCoordinates object (optionally loaded)
         self.zp = None  # the zero-point object (optionally loaded)
 
         self._aligner = None  # an ImageAligner object (lazy loaded using the provenance parameters)
         self._aligned_images = None  # a list of Images that are aligned to one image (lazy calculated, not committed)
-        self._combined_filepath = None  # a filepath built from invent_filepath and a hash of the upstream images
+        self._nandata = None
```
rknop, line 461: What is nandata?

(Update: I saw below.)

models/image.py
```diff
@@ -1720,6 +1758,65 @@ def score(self):
...
+    def psffluxerr(self, value):
+        self._psffluxerr = value
+
+    @property
+    def nandata(self):
+        """The image data, only masked with NaNs wherever the flag is not zero. """
+        if self._nandata is None:
+            self._nandata = self.data.copy()
+            if self.flags is not None:
+                self._nandata[self.flags != 0] = np.nan
```
rknop, line 1793: Ah, OK, now I see what nandata is.

rknop: If we ever feel the need to preserve memory, we can just do imageobj.nandata = None, but I wonder if it's worth making a less "look at the internals of the object" interface for that.  (Not worth thinking about too hard right now.)

(A perlmutter CPU node has 128 cores (256 threads) and 512GB of memory.  I wouldn't run more than 128 processes at once, so each thread has 4GB of memory, so for the most part we shouldn't be in too much trouble.  A 2048x4096 image uses 32MB of memory if float, 64MB if double, so we can store 64 or 128 images before we've used up our memory.  Each image has something like 4 images (more if it's a subtraction), so we should be reasonably comfortable *except* if we try to keep everything in memory when doing a coadd.)

models/image.py
```diff
@@ -1720,6 +1758,65 @@ def score(self):
...
+        """
+        Display the image using the matplotlib imshow function.
+
+        Parameters
+        ----------
+        **kwargs: passed on to matplotlib.pyplot.imshow()
+            Additional keyword arguments to pass to imshow.
+        """
+        import matplotlib.pyplot as plt
+        mu, sigma = sigma_clipping(self.data)
```
rknop, line 1810: I'll have to look at what you did with sigma-clipping.  I've used astropy.visualization.ZScaleInterval for this kind of thing in the past.

On other fronts: part of me thinks that this doesn't belong in the image class.  It's a view thing, not a model thing.

models/source_list.py
```diff
@@ -516,7 +524,7 @@ def load(self, filepath=None):
         if filepath is None:
             filepath = self.get_fullpath()
 
-        if self.format == 'sepnpy':
+        if self.format in ['sepnpy', 'filter']:
```
rknop, line 527: Is there any reason to define sepnpy and filter as two different formats?  (I'll have to read below, but so far it looks like they're synonyms.)

Update: yes, I saw below.  Ignore this.

models/source_list.py
```diff
@@ -731,6 +739,20 @@ def get_downstreams(self, session=None):
 
         return wcs + zps
 
+    def show(self, **kwargs):
```
rknop, line 742: Same comment about model/view separation as for the image show method.

A utility worth having (though I probably wouldn't do it here) is converting source positions to a ds9 region file.  It's easy to do (so I guess make an image), though you do have to remember that ds9 region files assume 1-center-offset pixels, not 0-center-offset pixels.

improc/tools.py
```diff
@@ -78,4 +78,131 @@ def sigma_clipping(values, nsigma=3.0, iterations=5, axis=None, median=False):
...
+
+    if norm == 1:
+        output /= np.sum(output)
+    elif norm == 2:
+        output /= np.sqrt(np.sum(output ** 2))
+
+    return output
+
+
+def make_cutouts(data, x, y, size=15):
```
rknop, line 151: The default size is something that should be configured in the yaml file.

pipeline/detection.py
```diff
@@ -99,7 +117,39 @@ def get_process_name(self):
...
 
 class Detector:
-    """Extract sources (and possibly a psf) from images or subtraction images."""
+    """Extract sources (and possibly a psf) from images or subtraction images.
+
+    There are a few different ways to get the sources from an image.
+    The most common is to use SExtractor, which is the default we use for direct images.
+    There is also "sep", which is a python-based equivalent (sort of) of SExtractor,
+    but it is not fully compatible and not fully supported.
+    Finally, you can use a matched-filter approach, meaning that you cross-correlate
```
rknop, line 126: Ah, OK, yes, sep and filter are different things.

pipeline/detection.py
```diff
@@ -624,6 +674,7 @@ def _run_psfex( self, tempname, image, psf_size=None, do_not_cleanup=False ):
                                 '-WRITE_XML', 'Y',
                                 '-XML_NAME', psfxmlfile,
                                 '-XML_URL', 'file:///usr/share/psfex/psfex.xsl',
```
rknop, line 676: ...we might want to make this one configurable.  The default will work inside the Docker image, so I don't think this is very urgent.  But, if somebody is trying to run on their own installation without using the Docker image, they may  not have the freedom to put stuff in /usr/share.

pipeline/detection.py
```diff
@@ -706,6 +757,121 @@ def extract_sources_sep(self, image):
...
+        score = image.score.copy()
+
+        if score is None:
+            raise NotImplementedError('Still need to add the matched-filter cross correlation! ')
+
+        psf_image = None
+        if image.psf is not None:
+            psf_image = image.psf.get_clip()
+        elif getattr(image, 'zogy_psf', None) is not None:
+            psf_image = image.zogy_psf
```
rknop, line 802: Is there ever a case where you'd want to use zogy_psf even if image.psf is not None?

pipeline/detection.py
```diff
@@ -706,6 +757,121 @@ def extract_sources_sep(self, image):
...
+
+        # remove the edges
+        border = fwhm_pixels * 2
+        score[:border, :] = np.nan
+        score[-border:, :] = np.nan
+        score[:, :border] = np.nan
+        score[:, -border:] = np.nan
+
+        # normalize the score based on sigma_clipping
+        # TODO: we should check if we still need this after b/g subtraction on the input images
```
rknop, line 828: (Is there an issue?)

pipeline/detection.py
```diff
@@ -706,6 +757,121 @@ def extract_sources_sep(self, image):
...
+        mu, sigma = sigma_clipping(score)
+        score = (score - mu) / sigma
+        det_map = abs(score) > self.pars.threshold  # catch negative peaks too (can get rid of them later)
```
rknop, line 831: I don't think you use it here, but it occurs to me that we might want two different thresholds.

The first one is the one where you find a single pixel that's above the threshold to preliminarily find objects.

The second one is the S/N inside the whole PSF (or inside an aperture, or whatever) for the objects that you keep.

(The extraction method we used in the SCP back in the late 90s did this two-step kind of thing.   I'm not sure in detail what sextractor does.)

pipeline/detection.py
```diff
@@ -706,6 +757,121 @@ def extract_sources_sep(self, image):
...
+
+        # dilate the map to merge nearby peaks
+        struct = np.ones((1 + 2 * fwhm_pixels, 1 + 2 * fwhm_pixels))
```
rknop, line 834: Do we want to do a circular rather than square dilation?  More effort, of course (though only a couple of lines).  But, doing it as squares means that the confusion length scale is going to be ~√2 different between 0° and 45°

(And, of course, we can't do *exactly* circular, so there's always going to be some direction quantization, but we can do the best possible binary pixel approximation.)

models/image.py
```diff
@@ -1720,6 +1758,65 @@ def score(self):
     def score(self, value):
         self._score = value
 
+    @property
+    def psfflux(self):
+        """
+        The flux of the PSF (2D float array).
```
rknop, line 1764: It would be worth documenting more exactly what this is supposed to be.

I'm assuming it is : this must be an odd-sized 2d array where the PSF is centered on the center pixel.  The entire array must sum to 1.  It represents the PSF at the same resolution as the image resolution, evaluated at the center of the image.  

Storing the PSF at image resolution isn't enough to really reconstruct the PSF for sub-pixel shifts, unless you're sufficiently oversampled (which we usually won't be in LS4).  psfex actually stores the PSF at a different resolution so it can do sub-pixel shifts without spreading the PSF out.

Becuase of this, and because this doesn't capture a variable psf, the property psfflux isn't really the right thing to use for everything you'd want to use a psf for.  It's worth putting in the documentation what it *is* good for.

pipeline/detection.py
```diff
@@ -706,6 +757,121 @@ def extract_sources_sep(self, image):
...
+            return arr[np.unravel_index(np.nanargmax(abs(arr)), arr.shape)]
+
+        scores = ndimage.labeled_comprehension(score, labels, all_idx, peak_score, float, np.nan)
+
+        def count_nans(arr):
+            return np.sum(np.isnan(arr))
+
+        num_flagged = ndimage.labeled_comprehension(score, labels, all_idx, count_nans, int, 0)
+
+        # TODO: add a correct aperture photometry, instead of the label_fluxes which only sums the labeled pixels
```
rknop, line 862: (Issue?)

pipeline/subtraction.py
```diff
@@ -116,8 +130,57 @@ def _subtract_zogy(self, new_image, ref_image):
...
+        ref_image_psf = ref_image.psf.get_clip()
+        new_image_noise = new_image.bkg_rms_estimate  # TOOD: improve this by using a Background object?
+        ref_image_noise = 1.0  # proper coaddition images have noise=1.0 by construction
+        new_image_flux_zp = 10 ** (0.4 * new_image.zp.zp)
+        ref_image_flux_zp = 10 ** (0.4 * ref_image.zp.zp)
+        # TODO: consider adding an estimate for the astrometric uncertainty dx, dy
+
+        # do some additional processing of the new image
+        mu, sigma = sigma_clipping(new_image_data)
+        new_image_data = (new_image_data - mu) / sigma  # TODO: skip this if we already background subtracted
```
rknop, line 150: Also TODO would be a way to specify different background subtraction methods.  (Though perhaps in that case you should just background subtract before throwing it into this function.  We'd need to document that clearly though, indicating which background subtraction method it will use if it doesn't get background-subtracted images.)

(I'll have to look at sigma clipping; it might be about the same thing as the sextractor background algorithm.  But, there's also things like the box size for smoothing when using a variable background.  And, for some fields, we might want to use a more complicated background subtraction method.)

rknop (approved):
This looks good.  Lots of little comments, but nothing really major.

There are a couple of places where I'd like to add a bit more meat to the comments and/or documentation.  Feel free to merge after you've done that.

improc/tools.py
```diff
@@ -78,4 +78,131 @@ def sigma_clipping(values, nsigma=3.0, iterations=5, axis=None, median=False):
...
+        The y positions of the cutouts.
+    size: int
+        The size of the cutouts. Default is 15.
+
+    Returns
+    -------
+    cutouts: 3D np.ndarray
+        The cutouts, with shape (len(x), size, size).
+    """
+    cutouts = np.full((len(x), size, size), np.nan)  # preallocate!
```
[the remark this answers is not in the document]
guynir42: I tend to agree about odd numbered cutouts but I'm not sure if we should enforce it or just let the sys admin decide.

models/image.py
guynir42: The `upstream_images` are sorted by MJD so you cannot assume anything on the order of ref vs. new.

models/image.py
guynir42: We can also use `np.where()` to produce this on the fly. I'm not sure if the extra memory or the extra CPU cycles is more expensive for us.

models/image.py
guynir42: I am thinking object-oriented rather than database mapping... having each object able to draw itself on the figure is good OOP but I'm not against having a more serious plot function somewhere else. Then it would still need to be object-specific (draw an Image is not the same as draw Cutouts).

models/source_list.py
guynir42: That seems like something that could live down at the end of our issues list.

pipeline/detection.py
guynir42: Maybe this should point to the local directory?

pipeline/detection.py
guynir42: Yeah, there are a few issues on background subtraction, I'm going to deal with this after Cutouts and Measurements.

pipeline/detection.py
guynir42: I think that if the image has a working PSF it should probably take precedence but I don't know how that would happen because measuring the PSF from the subtraction is not easy.

pipeline/detection.py
guynir42: I'm not sure what the other threshold does. When match-filtering, each pixel already contains the aggregated S/N for the area of the PSF around that pixel. So looking at a single pixel is the right thing to do for source detection on a matched-filtered image.

pipeline/detection.py
guynir42: Yes, I think this may be worth doing. I'm not sure if I want to put it into this PR, but as a small issue to improve precision it sounds like a good idea.

guynir42: Issue created.

pipeline/detection.py
guynir42: Added new issue.

pipeline/subtraction.py
guynir42: This is here only as a placeholder. Once we get organized on the background subtraction it will no longer be necessary to do any of this at the subtraction point. You'd just be guaranteed to be able to produce a background subtracted image on the fly because you'd have a Background object, the same way you have a PSF object.

pipeline/detection.py
guynir42: Ok I see, this is a configuration file. I opened an issue to deal with this someday.

PR #206, opened 2024-02-22
Cutouts Model

guynir42:
Create the actual code that saves and loads cutouts into HDF5 files.


rknop (commented):
Not a lot of coments, but one is big.  I'd like to think about having a CutoutsFile (or some such) model that is the one with the FileOnDiskMixin.  Then, each Cutout points to a CutoutsFile.  The CutoutsFile is what points to the subtraction.

This would allow us, for instance, to keep the "unique" constraint on the filename in FileOnDiskMixin.  It also narrows the Cutouts table, which is going to be one of the longest tables we have.

Cutouts wouldn't need a provenance itself, since it can just use the provenance of the CutoutsFile.

guynir42:
I agree that this is a good idea. 

But since I'm half way done with Measurements, I don't think this is the best time to make the change. 

I suggest I open an issue and start working on it after we are done with Measurements. I think I will also need some time to figure out the interfaces (e.g., would the datastore still hold a list of Cutouts objects or just the CutoutsFile? Would the CutoutsFile know all the Cutouts pointing to it?).

guynir42:
If you agree to defer please approve this PR so I can merge it into Measurements. Thanks!

PR #211, opened 2024-03-04
Test Data URL update

guynir42:
I moved all the data we need to one location on project `m4616` and replaced all the URLs for auto-downloading the data.


guynir42:
@rknop This is now synced with main and ready for review.

tests/fixtures/decam.py
```diff
@@ -149,8 +158,8 @@ def decam_raw_origin_exposures():
 
 
 @pytest.fixture(scope="session")
-def decam_filename(data_dir, cache_dir):
-    """Pull a DECam exposure down from the NOIRLab archives.
+def decam_filename(download_url, data_dir, decam_cache_dir):
+    """Pull a DECam exposure down from the NOIRLab archives.  TODO: should we download from NERSC or from NOIRLab?
```
rknop, line 162: Re: the TODO, that depends what we want to test.  If we want to test that we know how to pull from the NOIRLab archive, leave it that way.  If we don't care, and just want to get the exposure, it will almost certainly perform better where we will be running most of the time if we pull it from NERSC.

rknop (approved):
This one is pretty straightforward.  I like the use of the download_url fixture, and that we've consolodated things under the LS4 project.  Depending on external data not in the git archive means test will break in 10 years when somebody comes back and tries to use this, but I bet there will be worse breakages because of API changes in various python libraries.  In any event, for the duration of LS4 I expect that NERSC directory and URL to be stable.

tests/fixtures/decam.py
guynir42: I will make the fixture pull from our URL and add a "slow" test for downloading from NOIRLab.

Issue #194, opened 2024-01-23
PSF format for ZOGY


guynir42:
When running sextractor / PSFex we get a PSF object with a certain format. 
Other algorithms like ZOGY provide a simpler PSF result (often, this is just a single image of the PSF). 
We should expand the PSF object to be able to use this information. 
The use case (which is not urgent right now), is that a difference image from ZOGY will have a PSF that can be used downstream. Right now we don't use the PSF in the downstream of the difference image, so it doesn't really matter.

guynir42:
Related to #242

rknop:
Now that ImagePSF exists, zogy should use that.

PR #213, opened 2024-03-06
Measurements

guynir42:
The data model for Measurements and the associated pipeline step "measuring" are implemented.


guynir42:
@rknop there are still some issues with the tests that I will figure out next week, but I think the new code is not directly responsible for that. You can already start reviewing.

models/image.py
```diff
@@ -161,11 +157,16 @@ def new_image(self, value):
             raise ValueError('Do not assign None to new_image. Simply clear the upstream_images list.')
         if not isinstance(value, Image):
             raise ValueError("The new_image must be an Image object.")
-        if len(self.upstream_images) != 2:
-            raise ValueError("This only works for subtractions with exactly two upstream images.")
+        if len(self.upstream_images) not in [1, 2]:
+            raise ValueError("This only works for subtractions that have one or two upstream images.")
```
rknop, line 161: How can a subtraction only have one upstream image?

models/image.py
```diff
@@ -1728,6 +1735,17 @@ def get_downstreams(self, session=None):
 
             return downstreams
 
+    def get_psf(self):
+        """Load the PSF object for this image.
+
+        If it is a sub image, it will load the PSF from the new image.
```
rknop, line 1741: Note that this does hardcode the assumption that it's only the reference image we're convolving if the sub doesn't have its own PSF.  Generally that's what I do in alard-lupton subtractions, but I don't know if we want to bake that in.

models/measurements.py
```diff
@@ -40,9 +47,257 @@ class Measurements(Base, AutoIDMixin, SpatiallyIndexed):
...
+
+    mjd = association_proxy('cutouts', 'sources.image.mjd')
+
+    exp_time = association_proxy('cutouts', 'sources.image.exp_time')
+
+    filter = association_proxy('cutouts', 'sources.image.filter')
+
+    @property
+    def mag_psf(self):
+        return -2.5 * np.log10(self.flux_psf) + self.zp.zp  # what about aperture correction?
```
rknop, line 96: psf photometry shouldn't need aperture correction.  (Assuming the zp is a full-psf zp, not an aperture zp... I'm pretty sure we did it that way.  It's the most sane thing to do, but is *not* what the lensgridner pipeline does.)

models/measurements.py
```diff
@@ -40,9 +47,257 @@ class Measurements(Base, AutoIDMixin, SpatiallyIndexed):
...
+        return self.cutouts.sources.image.get_psf().get_clip(x=self.cutouts.x, y=self.cutouts.y)
+
+    @property
+    def pixel_scale(self):
+        return self.cutouts.sources.image.new_image.wcs.get_pixel_scale()
+
+    background = sa.Column(
+        sa.Float,
+        nullable=False,
+        doc="Background of the measurement, from a local annulus. Given as counts per pixel. "
```
rknop, line 148: Going back to my comment on measurements where we want to have the ability to do photometry without using an annulus, you could either make the background Null in that case, or make the background 0 and the background_err 0.  We'd have to think about whether setting both to 0 were enough, or if that might cause issues somewhere down the road.

pipeline/data_store.py
```diff
@@ -16,15 +16,15 @@
...
+    'exposure': [],  # no upstreams
     'preprocessing': ['exposure'],
     'extraction': ['preprocessing'],
     'astro_cal': ['extraction'],
     'photo_cal': ['extraction', 'astro_cal'],
     'subtraction': ['reference', 'preprocessing', 'extraction', 'astro_cal', 'photo_cal'],
     'detection': ['subtraction'],
     'cutting': ['detection'],
-    'measurement': ['detection', 'photo_cal'],
+    'measuring': ['cutting'],
```
rknop, line 27: This isn't important, but... we're mixing different kinds of nouns here (gerunds and non-gerunds).  We have subtraction instead of subtracting, but then measuring instead of measurement.

improc/photometry.py
```diff
@@ -0,0 +1,332 @@
...
+
+        # Create the circle
+        xgrid, ygrid = np.meshgrid(np.arange(self.imsize), np.arange(self.imsize))
+        xgrid = xgrid - self.imsize // 2 - x
+        ygrid = ygrid - self.imsize // 2 - y
+        r = np.sqrt(xgrid ** 2 + ygrid ** 2)
+        im = 1 + self.radius - r
+        im[r <= self.radius] = 1
+        im[r > self.radius + 1] = 0
+        # TODO: improve this with a better soft-edge function
```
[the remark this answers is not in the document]
rknop: (Issue #228)

improc/photometry.py
```diff
@@ -0,0 +1,332 @@
...
+            # get an offset annulus to get a local background estimate
+            inner = get_circle(radius=annulus[0], imsize=nandata.shape[0]).get_image(reposition_cx, reposition_cy)
+            outer = get_circle(radius=annulus[1], imsize=nandata.shape[0]).get_image(reposition_cx, reposition_cy)
+            annulus_map = outer - inner
+
+            # background and variance only need to be calculated once (they are the same for all apertures)
+            # but moments/centroids can be calculated for each aperture, but we will only want to save one
+            # so how about we use the smallest one?
+            if j == 0:  # smallest aperture only
+                # TODO: consider replacing this with a hard-edge annulus and do median or sigma clipping on the pixels
```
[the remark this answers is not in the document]
rknop: I made Issue #229.

default_config.yaml
```diff
@@ -90,12 +90,19 @@ detection:
   threshold: 5.0
 
 cutting:
-  cutout_size: 20
+  cutout_size: 25
+
+measuring:
+  annulus_radii: [10, 15]
+  annulus_units: pixels
```
rknop, line 97: There should be (and we should use) a aperture radius as well.  The default should be something like 1 FWHM -- meaning the units should be FWHM rather than pixels.  I'd then make the default annulus something like 10-12 FWHMs.  (That gives an area 44x the area of the aperture, so the nosie contribution should be small.)

pipeline/measuring.py
```diff
@@ -0,0 +1,307 @@
...
+        # the object did any work or just loaded from DB or datastore
+        self.has_recalculated = False
+
+        self._filter_bank = None  # store an array of filters to match against the cutouts
+        self._filter_psf_fwhm = None  # recall the FWHM used to produce this filter bank, recalculate if it changes
+
+    def run(self, *args, **kwargs):
+        """
+        Go over the cutouts from an image and measure all sorts of things
+        for each cutout: photometry (flux, centroids), real/bogus, etc.
```
rknop, line 118: real/bogus doesn't belong in this class; it should be done externally, as it's really a separate process.

(The analytical things like the elongation etc. do make sense to have here, as they really are properties of the photometry itself.  r/b is a property of photometry plus decisions about training, etc.  And, we want to have the abillity to add a new r/b method at any point without having to redo all photometry.)

pipeline/measuring.py
```diff
@@ -0,0 +1,307 @@
...
+            'pixels',
+            str,
+            'Units for the annulus radii. Can be pixels or fwhm. '
+            'This describes the units of the annulus_radii input. '
+            'Use "pixels" to make a constant annulus, or "fwhm" to '
+            'adjust the annulus size for each image based on the PSF width. '
+        )
+
+        # TODO: should we choose the "best aperture" using the config, or should each Image have its own aperture?
+        self.chosen_aperture = self.add_par(
```
rknop, line 41: This is how you specify which aperture to choose... but where is the list of apertures passed in?

OK, looks like it comes out of the zeropoint model.

Hmm.  We might want to have the abillity to do measurements in apertures other than the ones that were previously determined in the zeropoint model.  That does add complication, because we don't have a predetermined aperture correction, so we have to determine it on the fly.

Probably make an issue for this.

rknop (commented):
This mostly looks good.  I see there's already an issue for PSF photometry, and I added an issue for soft-edged circles.

I have two worries.  The first is that we're going to want to have the abillity to do photometry without subtracting an annulus.  The most obvious place for this is on a difference image, and that's also our primary use case.  I think this is worth thinking about now, rather than kicking it down the road with an issue, because the assumption of an annulus is baked into the code and the models in several places.  We should at least think about how the models would be modified to handle a no-annlulus case.  If we convince ourselves that it's a pretty easy modification, then we can just kick it to an issue.  But, we should make sure we aren't putting in structure here that will make that harder later.

I forgot what the second thing was, so it must not have been very important.

improc/photometry.py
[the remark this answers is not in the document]
guynir42: That might do it!

improc/tools.py
```diff
@@ -148,6 +152,101 @@ def make_gaussian(sigma_x=2.0, sigma_y=None, rotation=0.0, norm=1, imsize=None):
...
+    """
+    if imsize is None:
+        imsize = int(radius)
+        if imsize % 2 == 0:
+            imsize += 1
+
+    # round radius to the given resolution:
+    radius = np.round(radius / CACHED_RADIUS_RESOLUTION) * CACHED_RADIUS_RESOLUTION
+
+    if radius in CACHED_CIRCLES:
```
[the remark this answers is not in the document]
guynir42: I think the `make_circle` and `make_annulus` functions are not actually being used by anything. I can remove them.

models/image.py
guynir42: This happens when you are building up the object. If you want to assign the ref and new one after the other, you'd run into a step where it already has a reference but doesn't have the new image yet. This lets you assign the new image.

models/image.py
guynir42: The point of `get_psf` is to make the best effort to get the most relevant PSF. 
Since the ZOGY subtraction gives us a PSF image but without the bells and whistles a full PSF object, we are forced to go back to the new image for the next best thing. 
I have an issue to improve and unify the ZOGY output with our PSF model (#194).

models/measurements.py
guynir42: I think our ZP uses a "big" aperture and then adds corrections for the smaller apertures. But will need to check.

models/measurements.py
guynir42: The way I see it is that it is good to have the background measurements, but you don't have to use it. 
If the image is properly subtracted, then it should measure close to zero anyway, and supposedly be less noisy than the aperture, so subtracting that number should not introduce a lot of noise (and you don't have to subtract it). 

Another thing to think about is the variance in the annulus is a good estimator for the flux error. 

Finally, it sometimes helps to have the background measurement on hand, to rule out bad data. For example if the object seems to brighten on an image or two, but also the background, then it is indicative of a false detection (e.g., a passing asteroid).

pipeline/data_store.py
guynir42: Yeah I know it is annoying. But I don't know if extraction->extracting or detection->detecting is a real improvement. 

I am sure that changing measurement (the verb) to differentiate it from measurements (the noun) is a step in the right direction.

default_config.yaml
guynir42: The apertures are determined by the ZP corrections. I think I remember that we discussed this and decided that we 100% want the apertures to match the ZP. I know this because I had that as a parameter and then ripped it out of the class when we settled on just copying over the aperture sizes from the ZP...

pipeline/measuring.py
guynir42: Yes, you are right, we need to add another model and processing step for R/B.

pipeline/measuring.py
guynir42: See my comment above for the apertures coming from the ZP.

models/image.py
rknop: But isn't the new psf wrong for the ZOGY subtraction?  I thought in ZOGY that both images got convolved with something.

default_config.yaml
rknop: Yeah OK.  We should have those zp apertures configured somewhere on the config file, though.

guynir42: They are pulled from the instrument. If you recall, I was complaining last week that there are too many of them...

models/image.py
guynir42: If the reference has a narrower PSF than the new image, then the ZOGY PSF will look a lot like the new PSF. 

The right thing to do is to always use the PSF from ZOGY (when using that subtraction) but to store it in a way that is compatible with the PSF model, and save it to DB with its own provenance and so on. 

As far as I can tell the PSF photometry in Measurements is the only place where that would be used... and I am not sure how good PSF photometry is going to turn out to be (in the end, everyone just uses aperture photometry).

models/measurements.py
rknop: Right, but we need the ability to not use it even for that.

rknop: I'll make an issue so we can merge this one as it is now.

rknop (approved):

PR #220, opened 2024-03-19
REJECT - Add empty file to test git workflow 

whohensee:
Just testing pushing and automated tests.


whohensee:
usage of print() has been replaced with _logger.debug() from models.base. The following edge-cases remain
- in docs/datastorage.md, print() is used. I have left it untouched because it appears to be trying to be minimalistic for demonstration purposes.
- I have left untouched anything in nersc-upload-connector/
- I also searched using `grep -lr "sys.std"` and found a single case of it being used in models/base.py, but it was not using the write() method, so I don't think it is relevant.

rknop (approved):
This seems straightforward enough.

guynir42:
@whohensee please do `git fetch upstream` and `git merge upstream/main` to update to the newest version, then the "changed files" will only include the things you've actually changed.

PR #224, opened 2024-03-25
Fix decam test fixture

rknop:
I don't know why the tests passed on github, but there was a cache_dir that needed to be renamed to decam_cache_dir in fixtures/decam.py


guynir42 (approved):

guynir42:
It passed because github never uses the cache folder, it recreates/downloads everything when it runs.

PR #233, opened 2024-03-31
Add a __del__ method to FileOnDiskMixin

guynir42:
This cleans up un-committed files (if they don't have a primary key `id` set). 
This prevents orphan files of objects that were created and for some reason saved to disk / archive, but were never persisted. 

With this we can also remove the `ImageAlignment.temp_images` list that was used to help prevent build up of temporary aligned images and their associated files.

### DIFF
--- a/improc/alignment.py
+++ b/improc/alignment.py
@@ -98,17 +98,6 @@ class ParsImageAligner(Parameters):
 
 
 class ImageAligner:
-    temp_images = []
-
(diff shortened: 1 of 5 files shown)


guynir42:
This doesn't seem to work, since the `__del__` gets called as soon as you merge an object and throw away its reference in favor of the new object. 
I will need to come up with a new idea.

PR #241, opened 2024-04-04
Make sure bitflag is propagated correctly

whohensee:
regarding project issue https://github.com/orgs/c3-time-domain/projects/1/views/1?pane=issue&itemId=57122234

### DIFF
--- a/tests/pipeline/test_pipeline.py
+++ b/tests/pipeline/test_pipeline.py
@@ -397,7 +397,6 @@ def test_get_upstreams_and_downstreams(decam_exposure, decam_reference, decam_de
 
             # test get_downstreams
             assert [downstream.id for downstream in ds.exposure.get_downstreams(session)] == [ds.image.id]
-            # image get_downstreams appears to give duplicated results, does not seem too harmful but making note
             assert [downstream.id for downstream in ds.image.get_downstreams(session)] == [ds.psf.id,
                                                                                     ds.sources.id,
                                                                                     ds.wcs.id,
--- a/pipeline/astro_cal.py
+++ b/pipeline/astro_cal.py
@@ -251,12 +251,6 @@ class AstroCalibrator:
 
         ds.wcs = WorldCoordinates( sources=sources, provenance=prov )
         ds.wcs.wcs = wcs
-
-        # run calls this method, but perhaps I should still move this into run()?
-        # if ds.wcs._upstream_bitflag is None:
-        #     ds.wcs._upstream_bitflag = 0
-        # ds.wcs._upstream_bitflag |= sources.bitflag
-
         if session is not None:
             ds.wcs = session.merge( ds.wcs )
 


pipeline/detection.py
```diff
@@ -306,6 +306,10 @@ def extract_sources(self, image):
         else:
             raise ValueError(f'Unknown extraction method "{self.pars.method}"')
 
+        # perhaps this spot will end up being better, as it catches all methods and catches both psf and sourcelist 
+        # psf._upstream_bitflag = 0   # perhaps this spot will end up being better, as it catches all methods
+        # psf._upstream_bitflag |= image.bitflag
```
guynir42, line 311: Yes. I don't see much advantages to doing this inside the specific extract methods. 
In general I think the design we use is to do only the hard work inside the sub-routines and do all the housekeeping code on the `run` function (in this case it would be `extract_sources` because of the duality of the `Detector` class).

pipeline/photo_cal.py
```diff
@@ -278,6 +278,9 @@ def run(self, *args, **kwargs):
             # Make the ZeroPoint object
             ds.zp = ZeroPoint( sources=ds.sources, provenance=prov, zp=zpval, dzp=dzpval,
                                aper_cor_radii=sources.aper_rads, aper_cors=apercors )
+            
+            ds.zp._upstream_bitflag = 0
+            ds.zp._upstream_bitflag |= sources.bitflag
```
guynir42, line 283: Photometric calibration depends also on astrometric calibration... so you'd want to add the `wcs` bitflag here as well.

tests/pipeline/test_pipeline.py
```diff
@@ -246,6 +246,45 @@ def test_data_flow(decam_exposure, decam_reference, decam_default_calibrators, a
...
+    Does not check measurements, as they do not have the HasBitflagBadness Mixin.
+    """
+    exposure = decam_exposure
+    ref = decam_reference
+    sec_id = ref.section_id
+
+    try:  # cleanup the file at the end
+        p = Pipeline()
+        assert p.extractor.pars.threshold != 3.14
+        assert p.detector.pars.threshold != 3.14
```
guynir42, line 261:
```suggestion
```

tests/pipeline/test_pipeline.py
```diff
@@ -246,6 +246,45 @@ def test_data_flow(decam_exposure, decam_reference, decam_default_calibrators, a
...
+
+        exposure.badness = 'banding'  # add a bitflag to check for propagation
+        ds = p.run(exposure, sec_id)
+
+        assert ds.image._upstream_bitflag == 2    # 2 is the bitflag for 'banding'
```
guynir42, line 266:
```suggestion
        assert ds.exposure._bitflag == 2    # 2 is the bitflag for 'banding'
        assert ds.image._upstream_bitflag == 2 
```

tests/pipeline/test_pipeline.py
```diff
@@ -246,6 +246,45 @@ def test_data_flow(decam_exposure, decam_reference, decam_default_calibrators, a
...
+        assert ds.wcs._upstream_bitflag == 2
+        assert ds.zp._upstream_bitflag == 2
+        assert ds.sub_image._upstream_bitflag == 2
+        assert ds.detections._upstream_bitflag == 2
+        for cutout in ds.cutouts:   # cutouts is a list of cutout objects
+            assert cutout._upstream_bitflag == 2
+
+        # commit to DB using this session
+        with SmartSession() as session:
+            ds.save_and_commit(session=session)
```
guynir42, line 278: I think that if you are not going to use these objects in the test then there's no reason to commit them... unless you are testing that it is not failing to commit the objects with bitflags, which I don't see as a serious possibility.

tests/pipeline/test_pipeline.py
```diff
@@ -246,6 +246,45 @@ def test_data_flow(decam_exposure, decam_reference, decam_default_calibrators, a
...
+        assert ds.image._upstream_bitflag == 2    # 2 is the bitflag for 'banding'
+        assert ds.sources._upstream_bitflag == 2
+        assert ds.psf._upstream_bitflag == 2
+        assert ds.wcs._upstream_bitflag == 2
+        assert ds.zp._upstream_bitflag == 2
+        assert ds.sub_image._upstream_bitflag == 2
+        assert ds.detections._upstream_bitflag == 2
+        for cutout in ds.cutouts:   # cutouts is a list of cutout objects
+            assert cutout._upstream_bitflag == 2
+
```
guynir42, line 275: Two more things that I'd add to this test:
1) What happens if you add a second bad bit on e.g., the SourceList and see that the `sources.bitflag` is indeed the combination of both, and that the `_upstream_bitflag` of all downstreams is also the combination of all of them. You may need to delete all downstreams and re-run the pipeline to see the change.
2) There's an "update bitflag" command that should automatically propagate bitflags downstream to existing objects. I would try to repeat (1) but add the badness on Image or Exposure and without re-running anything make sure you can both add and remove badness from things and have it propagated correctly.

guynir42 (changes requested):
This is very good. I am glad to have this part functioning, I should have taken care of it months ago. 
I do think you can beef up the test a little bit (see comment below). 
After we merge this I will also implement badness for Measurements.

tests/pipeline/test_pipeline.py
whohensee: below in the parts labeled 'test part 2' and 'test part 3' I started these. I spent some time looking into exactly how the pipeline kept track of objects (memory, archive, database, etc), and I think I have implemented the first one, but I left commented out some of the infrastructure i was trying first. I ended up realizing nothing was committed to db, but had been concerned about copies existing locally somewhere and getting used instead of re-created, thus the delete functions. Some input on the best way to properly 'delete' pipeline objects would be helpful if I've missed anything.

whohensee:
@guynir42 I'm marking you for review again, this PR is definitely somewhat incomplete but I could use a bit of input (mostly in the comments of the test but also in the github comment you left). Let me know if this is 'too incomplete' to be tagging you to look at and I can certainly try to clean it up more for future PRs

tests/pipeline/test_pipeline.py
```diff
@@ -238,6 +238,91 @@ def test_data_flow(decam_exposure, decam_reference, decam_default_calibrators, a
...
+        assert ds.sub_image._upstream_bitflag == desired_bitflag # note my comment in subtraction.py
+                                                                 # only passes if gets bitflag from sources
+
+
+        # test part 3: test update_downstream_badness() function by adding and removing flags
+        # and observing propagation
+        # to Guy: this is partially tested already in test_image.py::test_image_badness,
+        #   but perhaps this is still useful due to testing propagation all the way downstream
+        # ALSO this function will not run on a full ds until get_upstreams has been defined in all the
+        #   downstream classes (currently just sources, image, exposure)
```
whohensee, line 308: just drawing attention to this - if I want to run this function on a datastore full of data products, each of them will need `get_upstreams` to be implemented

whohensee: I consider that perhaps implementing get_upstreams on the various objects could be another PR, and finishing this test could be another PR which depends on that

whohensee: nevermind, I will implement `get_upstreams` as part of this PR, as per the meeting discussion

tests/pipeline/test_pipeline.py
```diff
@@ -246,6 +247,190 @@ def test_data_flow(decam_exposure, decam_reference, decam_default_calibrators, a
...
+            # image get_downstreams appears to give duplicated results, does not seem too harmful but making note
+            assert [downstream.id for downstream in ds.image.get_downstreams()] == [ds.psf.id,
+                                                                                    ds.psf.id,
+                                                                                    ds.sources.id,
+                                                                                    ds.sources.id,
+                                                                                    ds.wcs.id,
+                                                                                    ds.wcs.id,
+                                                                                    ds.zp.id,
+                                                                                    ds.zp.id,
+                                                                                    ds.sub_image.id]
```
guynir42, line 410: This is interesting. Do you have any ideas why the duplicates?

guynir42 (approved):
I have some minor comments. I think if you manage to address them all you should go ahead and merge. 
Good job on this one!

PR #245, opened 2024-04-16
wcs FileOnDiskMixin

whohensee:


tests/models/test_image.py
```diff
@@ -1365,7 +1365,7 @@ def test_image_products_are_deleted(ptf_datastore, data_dir, archive):
     # make sure the files are there
     local_files = []
     archive_files = []
-    for obj in [im, im.psf, im.sources]:  # TODO: add WCS when it becomes a FileOnDiskMixin
+    for obj in [im, im.psf, im.sources, im.wcs]:  # TODO: add WCS when it becomes a FileOnDiskMixin
```
whohensee, line 1368: todo: remove comment if looks good

tests/models/test_world_coordinates.py
```diff
@@ -107,3 +120,40 @@ def test_world_coordinates( ztf_datastore_uncommitted, provenance_base, provenan
...
+
+    with SmartSession() as session:
+        try:
+            wcobj.save()
+
+            txtpath = pathlib.Path( wcobj.local_path ) / f'{wcobj.filepath}'
+            wcobj2 = WorldCoordinates()
+            wcobj2.load( txtpath=txtpath )
+
+            # breakpoint()
```
whohensee, line 150:
```suggestion
```

alembic/versions/2024_05_22_2011-2ea9f6f0b790_make_wcs_a_fileondiskmixin.py
```diff
@@ -0,0 +1,46 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index('cutouts_q3c_ang2ipix_idx', table_name='cutouts')
+    op.drop_index('exposure_q3c_ang2ipix_idx', table_name='exposures')
+    op.drop_index('images_q3c_ang2ipix_idx', table_name='images')
+    op.drop_index('measurements_q3c_ang2ipix_idx', table_name='measurements')
```
guynir42, line 24: Alembic doesn't know about q3c indices so it always tries to remove them on each revision. Make sure you get rid of these (in the downgrade, too).

alembic/versions/2024_05_22_2011-2ea9f6f0b790_make_wcs_a_fileondiskmixin.py
```diff
@@ -0,0 +1,46 @@
...
+Create Date: 2024-05-22 20:11:31.195961
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = '2ea9f6f0b790'
+down_revision = '9a4097979249'
```
guynir42, line 14: Before you merge, make sure the last revision on "main" has a hash that matches the down revision. If some one else has merged a migration you will need to update this.

models/world_coordinates.py
```diff
@@ -152,4 +120,72 @@ def get_downstreams(self, session=None):
...
+
+        # if not, generate one
+        else:
+            if self.image.filepath is not None:
+                self.filepath = self.image.filepath
+            else:
+                self.filepath = self.image.invent_filepath()
+
+            if self.provenance is None:
+                raise RuntimeError("Can't invent a filepath for the WCS without a provenance")
```
guynir42, line 154: Maybe move this check above the invent filepath so it fails without ever making a partial filename

models/world_coordinates.py
```diff
@@ -152,4 +120,72 @@ def get_downstreams(self, session=None):
...
+            self.filepath += f'.wcs_{self.provenance.id[:6]}.txt'
+
+        txtpath = pathlib.Path( self.local_path ) / f'{self.filepath}'
```
guynir42, line 157:
```suggestion
        txtpath = pathlib.Path( self.local_path ) / self.filepath
```

models/world_coordinates.py
```diff
@@ -152,4 +120,72 @@ def get_downstreams(self, session=None):
...
+
+        # ----- Get the header string to save and save ----- #
+        header_txt = self.wcs.to_header().tostring(padding=False, sep='\\n' )
+
+        if txtpath.exists():
+            if ( 'overwrite' in kwargs and not kwargs['overwrite'] ):
```
guynir42, line 163:
```suggestion
            if kwargs.get('overwrite', True):
```

models/world_coordinates.py
```diff
@@ -152,4 +120,72 @@ def get_downstreams(self, session=None):
...
+        ----------
+        txtpath: str, Path, or None
+            File to read. If None, will load the file returned by self.get_fullpath()
+        """
+
+        if txtpath is None:
+            txtpath = self.get_fullpath( download=download, always_verify_md5=always_verify_md5)
+
+        if txtpath is None:
+            raise ValueError("WCS object has no filepath locally or in archive.")
```
guynir42, line 186: This error might be a little misleading. The filepath doesn't come from the archive, it is either defined on the object (i.e., on the database row of this WorldCoordinates) or it is given by the user as an override (this will usually not be the case). 

You can say we didn't find a file with that filepath not locally or on the archive, but that would be an error raised by `get_fullpath` and not here.

guynir42: You could add a check to see if the file exists (e.g., `os.path.isfile(txtpath)`) and then say something like "file is missing". This could happen if you specify `download=False` and it doesn't exist locally (even when the filepath exists).

tests/fixtures/pipeline_objects.py
```diff
@@ -466,18 +466,19 @@ def make_datastore(
...
                         upstreams=[ds.sources.provenance],
                         parameters=p.astro_cal.pars.get_critical_pars(),
                         is_testing=True,
                     )
+                cache_name = f'{cache_base_name}.wcs_{prov.id[:6]}.txt.json'
+                cache_path = os.path.join(cache_dir, cache_name)
+                if os.path.isfile(cache_path):
+                    SCLogger.debug('loading WCS from cache. ')
+                    ds.wcs = WorldCoordinates.copy_from_cache(cache_dir, cache_name)
+                    ds.wcs.load() # TODO COME BACK TO THIS, MAYBE UNNECESSARY
```
guynir42, line 481: I don't think this is needed since the "data" attribute (`wcs` in this case) would be lazy loaded when needed (probably when you call `save()`).

tests/fixtures/pipeline_objects.py
```diff
@@ -500,13 +501,16 @@ def make_datastore(
 
                     ds.wcs.provenance = prov
                     ds.wcs.sources = ds.sources
+                    # make sure this is saved to the archive as well
+                    ds.wcs.save(verify_md5=False, overwrite=True) #copied this from above in PSF
```
guynir42, line 505: I think overwrite is not strictly needed (the default is to allow overwrite, yes?). 
As for `verify_md5`, I think that just saves some runtime but doesn't really matter either way.

tests/models/test_image.py
guynir42: Yes remove this comment. 
In general don't worry too much about removing things, they all show up in the diff so you can always bring them back if you need to.

tests/models/test_world_coordinates.py
```diff
@@ -58,17 +62,17 @@ def test_world_coordinates( ztf_datastore_uncommitted, provenance_base, provenan
                 upstreams=[provenance_extra],
                 is_testing=True,
             )
+            wcobj.save() # This is ok here vs in session, right?
```
guynir42, line 65: Yeah, save should work outside of the session.

tests/models/test_world_coordinates.py
```diff
@@ -78,6 +82,13 @@ def test_world_coordinates( ztf_datastore_uncommitted, provenance_base, provenan
                 session.commit()
             session.rollback()
 
+            # ensure you cannot overwrite when explicitly setting overwrite=False
+            with pytest.raises(
+                OSError,
+                match=".txt already exists"
+            ):
```
guynir42, line 89: You can put these arguments all on one line, it should easily fit into 120 characters...

guynir42 (approved):
I have some minor comments and suggestions. Feel free to go over them and when they are fixed you can merge. 

Don't forget to "squash and merge" and don't include all the commits in the description message. 

Also, make sure the alembic `down_revision` hash is indeed pointing to the last revision on main.

models/world_coordinates.py
whohensee: I wrote this check to raise the error if overwrite exists and is explicitly set to 'False', so I think this suggestion reverses the logic, though I like the syntax used in yours.
I think I'll use
`if kwargs.get('overwrite', True) == False:`
which I like a little better for reading clarity than
`if not kwargs.get('overwrite', True):`

models/world_coordinates.py
whohensee: Would you say I should just remove the error you commented about then, and leave it to be caught somewhere in `get_fullpath` then?

I added the check for file existence and a test for it in `test_save_and_load`.

PR #247, opened 2024-04-17
Update files in devshell to work again, update related docs.

rknop:


devshell/seechange_devshell.yaml
```diff
@@ -1,5 +1,24 @@
...
+archive:
+  archive_url: http://devshell_archive:8080/
+  verify_cert: false
+  path_base: test/
+  local_read_dir: null
+  local_write_dir: null
+  token: insecure
+
+subtraction:
+  method: zogy
```
guynir42, line 24: This seems a rather specific thing to have in this config file. I think the general config already has this... I am guessing you don't actually need it...

docs/setup.md
```diff
@@ -13,59 +13,52 @@ but in the meantime, install Docker Engine instead of Docker Desktop; instructio
...
 #### Development shell -- local database
 
-To create a database on your local machine and get a development shell in which to run code, 
-cd into the `devshell` directory and run
-```
-   USERID=<UID> GROUPID=<GID> docker compose up -d
-```
-replacing `<UID>` with your uid, and `<GID>` with your GID.  
-You can avoid typing this all the time by creating a file `.env` in the `devshell` directory with contents
+The `devshell` directory has a docker compose file that can create a development environment for you.  To set it up, you need to set three environment varaibles.  You can either manually set these with each and every `docker compose` command, you can set them ahead of time with `export` commands, or, recommended, you can create a file `.env` in the `devshell` directory with contents:
```
guynir42, line 16:
```suggestion
The `devshell` directory has a docker compose file that can create a development environment for you.  To set it up, you need to set three environment variables.  You can either manually set these with each and every `docker compose` command, you can set them ahead of time with `export` commands, or, recommended, you can create a file `.env` in the `devshell` directory with contents:
```

docs/setup.md
```diff
@@ -114,6 +109,8 @@ After editing any schema, you have to create new database migrations to apply th
 The comment will go in the filename, so it should really be short.  
 Look out for any warnings, and review the created migration file before applying it (with `alembic upgrade head`).
 
+Note that in the devshell and test docker environments above, database migrations are automatically run when you create the environment.  In those cases, you will only need to run `alembic upgrade head` if you've created a new databse migration since you started up your docker environment.
```
guynir42, line 112: This is not clear: if you've run migrations then it sounds like you don't need to `upgrade head`. Or do you mean that if you left the same database service running and added a migration after it was already created, then you need to `upgrade head`?

guynir42 (approved):
I have a couple of minor comments. I like the improved documentation!

devshell/seechange_devshell.yaml
rknop: I copied this directly from what was in the tests config file.

docs/setup.md
rknop: The latter.

PR #253, opened 2024-04-18
Add free() methods to Image, SourceList, and PSF

rknop:
really only matters for Image.  Gives us the ability to free working memory.  Not currently used anywhere other than tests.

### DIFF
--- a/models/image.py
+++ b/models/image.py
@@ -1468,7 +1468,9 @@ class Image(Base, AutoIDMixin, FileOnDiskMixin, SpatiallyIndexed, FourCorners, H
              psfflux, psffluxerr, nandata, and nanscore.
 
         """
-        tofree = set( only_free ) if only_free is not None else image.saved_extensions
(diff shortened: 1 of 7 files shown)


models/image.py
```diff
@@ -1436,6 +1436,68 @@ def load(self):
...
+          only_free: set or list of strings
+             If you pass this string, it will not free everything, but
+             only the things you specify here.  Members of the string
+             can include raw_data, data, weight, flags, background, score,
+             psfflux, psffluxerr, nandata, and nanscore.
+
+        """
+        allfree = { 'raw_data', 'data', 'weight', 'flags', 'background',
+                    'score', 'psfflux', 'psffluxerr',
+                    'nandata', 'nanscore' }
```
guynir42, line 1473: You may be able to just use `Image.saved_extensions`

guynir42 (approved):
I have one comment, after that feel free to merge.

PR #254, opened 2024-04-18
Add sigma clipping for background estimation

dryczanowski:
Change background estimation to use hard edge annulus with sigma clipping. Add relevant tests.

### DIFF
--- a/improc/photometry.py
+++ b/improc/photometry.py
@@ -23,6 +23,8 @@ def get_circle(radius, imsize=15, oversampling=100, soft=True):
     oversampling: int
         The oversampling factor for the circle.
         Default is 100.
+    soft: bool
+        Toggle the soft edge of the circle. Default is True (soft edge on).
 
     Returns
     -------
@@ -71,10 +73,15 @@ class Circle:
         xgrid = xgrid - self.imsize // 2 - x
         ygrid = ygrid - self.imsize // 2 - y
         r = np.sqrt(xgrid ** 2 + ygrid ** 2)
-        if soft==True:
+        if self.soft==True:
             im = 1 + self.radius - r
             im[r <= self.radius] = 1
             im[r > self.radius + 1] = 0
+        else: 
+            im = r
+            im[r <= self.radius] = 1
+            im[r > self.radius] = 0
+        
         # TODO: improve this with a better soft-edge function
 
         return im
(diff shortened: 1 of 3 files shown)


dryczanowski:
Flaky test fails sometimes on test_measuring, to do with the background error. Sometimes fails on [10], sometimes [11], sometimes passes. Unsure if the variance should also be sigma clipped which might fix the issue but unsure if this actually makes sense.

improc/photometry.py
```diff
@@ -253,18 +262,16 @@ def iterative_photometry(
...
             annulus_map = outer - inner
 
             # background and variance only need to be calculated once (they are the same for all apertures)
             # but moments/centroids can be calculated for each aperture, but we will only want to save one
             # so how about we use the smallest one?
             if j == 0:  # smallest aperture only
-                # TODO: consider replacing this with a hard-edge annulus and do median or sigma clipping on the pixels
-                background = np.nansum(nandata * annulus_map) / np.nansum(annulus_map)  # b/g per pixel
+                background = np.nansum(sigma_clip(nandata * annulus_map,sigma=5)) / np.nansum(annulus_map)  # sigma-clipped b/g per pixel
                 variance = np.nansum((nandata - background) * annulus_map) ** 2 / np.nansum(annulus_map)  # per pixel
```
guynir42, line 274:
```suggestion
                background, standard_dev = sigma_clipping(ndata * annulus_map, sigma=5)
                variance = standard_dev ** 2
```

guynir42: The sigma clipping already gives the mean value in the list of pixels it is given, so we don't need to divide by sum of annulus map (the number of pixels). 
It also gives you the standard deviation of those pixels, so we don't need to calculate the variance manually anymore.

guynir42: Also, notice that astropy's `sigma_clip` is not the same as my `sigma_clipping` code...

guynir42 (commented):
Good! my only comment is regarding the use of sigma clipping. I think because there are two version (the one I wrote and the one from astropy) that are not doing the same thing, then it may be that the code you put in there is doing the same thing. But I'm not sure what astropy does and if it is better.

guynir42 (approved):
Looks good! Feel free to merge this. Use "squash and merge" and don't forget to replace the list of commits with a shorter message.

guynir42:
You have conflicts. Do `git fetch upstream` and the `git merge upstream/main`, resolve the conflicts manually and commit. Then if tests are passing you can squash and merge.

PR #260, opened 2024-04-30
Bkg estimate

rknop:
Use the background mean and sigma produced by sextractor to load that field in the images table.  Done in the extraction step.

### DIFF
--- a/pipeline/detection.py
+++ b/pipeline/detection.py
@@ -235,7 +235,7 @@ class Detector:
                 #  in any event, need a way of passing parameters
                 #  Question: why is it not enough to just define what you need in the Parameters object?
                 #  Related to issue #50
-                detections, _ = self.extract_sources( image )
+                detections, _, _, _ = self.extract_sources( image )
 
                 detections.image = image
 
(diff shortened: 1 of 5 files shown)


pipeline/detection.py
```diff
@@ -588,20 +599,26 @@ def _run_sextractor_once( self, image, apers=[5, ], psffile=None, tempname=None,
                                f"-------\nstdout:\n{res.stdout}" )
                 raise RuntimeError( f"Error return from source-extractor call" )
 
+            # Get the background from the xml file that sextractor wrote
+            sextrstat = votable.parse( tmpxml ).get_table_by_index( 1 )
+            bkg = sextrstat.array['Background_Mean'][0][0]
+            bkgsig = sextrstat.array['Background_StDev'][0][0]
```
guynir42, line 605: ugg I hate these command line utilities so much 🤮

guynir42 (approved):
Great!

pipeline/detection.py
rknop: Just keep repeating to yourself : it's just an interface.  The disk file is a memory buffer.

PR #261, opened 2024-04-30
Zogy improc memory fix

whohensee:
I noticed in a regular run of the pipeline that the method `zogy_subtract` used an enormous amount of memory (about 4.8g during the method) due to creating a large amount of numpy arrays and leaving them around until the method ran out of scope.

Most arrays created were size 8e6 items and of itemsize 8 or 16 bytes (split about half and half), so each array was of size 65/130 e6 bytes.

With this rough proposed fix I change nothing in the logic of the method, but simply delete each array when it is no longer needed.

The result is that in a single run of the method it goes from using 4.8g to peaking around 2.2g, and in a test that runs the full pipeline once, in linux's top the pytest memory usage peaks roughly around 2.5g instead of 5.8.


rknop:
This was subsumed in PR 406

Issue #265, opened 2024-05-07
Measurements module is broken


rknop:
The following code throws an exception:

from models.base import Session
from models.measurements import Measurements
s = Session()
s.query( Measurements )

It really shouldn't.

The problem is that the one of the sqlalchemy constructions in Measurements object refers to an Object object.  Just adding an import of models.objects to measurements.py doesn't work, because you get a circular import.

The workaround is to import objects.py in anything that imports measurements.py.  But, really, we should have something that works without having to use that workaround.

SQLAlchemy is such a disaster.

guynir42:
I don't really understand where this comes from. What was the error?

PR #266, opened 2024-05-07
Import Object whenever importing Measurements

rknop:
See Issue #265

I suspect this never showed up in tests because everything in tests that imported measurements also imported objects.  I ran across this working on the stress test.  I didn't import either, but then had a crash when there was an sqlalchemy query somewhere in there because a sqlalchemy construct in Measurements referenced Object, but Object had never been imported.


guynir42 (approved):
It is surprising that you really need all three imports (I'd expect one would be missing). But fine by me.

PR #267, opened 2024-05-07
Ref table name

rknop:
We can't name a table "references" because that's a reserved postgresql word.  It looks like it was working when calling from python, but trying to do SQL directly didn't work with that table name.  So, name it "refs" instead.


guynir42 (approved):
Ha that is good!

PR #268, opened 2024-05-09
Report model and DataStore failure modes

guynir42:
We add the Report model that tracks the progress of each pipeline run on a single section of an Exposure. 

The report first and foremost tracks the start and (optional) end time of the pipeline. 
If the pipeline didn't finish (because it is still running or hanging) it will have a Null end time. 
If the pipeline had an exception, the end time should also be recorded.

There is also room to track any error that stopped the pipeline, warnings that come up, and the memory/runtime of each step. 

We also use bitflags to record which processing steps were done, which pipeline products were already instantiated when the pipeline was finished, and which were successfully saved to disk/database (this one is a little tricky!). 

To get this to work we also need to change how pipeline objects handle errors and make sure they always return a datastore with appended warnings and errors, so they can be saved to the report.


pipeline/data_store.py
```diff
@@ -62,7 +66,7 @@ class DataStore:
         'measurements'
     ]
 
-    attributes_to_clear = [
+    products_to_clear = [
```
rknop, line 69: If the purpose of this is to save memory, probably we also want to clear detections, measurements, cutouts.  Those aren't as big as the images, but esp. on a very full image, cutouts could be worth cleaning up.  (And if we clean up cutouts, we should clean up the other stuff that goes with them.)

pipeline/data_store.py
```diff
@@ -277,6 +211,122 @@ def parse_args(self, *args, **kwargs):
...
+            return self.reference.image
+        return None
+
+    @ref_image.setter
+    def ref_image( self, value ):
+        if self.reference is None:
+            self.reference = Reference()
+        self.reference.image = value
+
+    def __getattr__(self, key):
```
rknop, line 323: Python question : is this function guaranteed to be called before (say) the ref_image @property function is called when you do `ds.ref_image`?

...looking at docs, it seems that the case is No, that this only gets called if there's an AttributeError.

Did you want `__getattribute__`?

(Based on the comment, I'm assuming you want this to be called on *any* attribute access.)

rknop: Test code:
```
class Thing:
    @property
    def twox( self ):
        return 2. * self.val
    
    def __init__( self, val ):
        self.val = val

    def __getattr__( self, attr ):
        print( "__getattr__ called" )

thing = Thing( 5 )
print( thing.val )
print( thing.twox )
print( thing.puppy )
```

prints:

```
5
10.0
__getattr__ called
None
```

but if I understand your comment correctly, you would have wanted "`__getattr__ called`" printed before the lines with 5 and 10.0.

rknop:
```
class Thing:
    @property
    def twox( self ):
        return 2. * self.val
    
    def __init__( self, val ):
        self.val = val

    def __getattr__( self, attr ):
        print( "__getattr__ called" )

    def __getattribute__( self, attr ):
        print( "__getattribute__ called" )
        return super().__getattribute__( attr )
        
thing = Thing( 5 )
print( thing.val )
print( thing.twox )
print( thing.puppy )
```

writes

```
__getattribute__ called
5
__getattribute__ called
__getattribute__ called
10.0
__getattribute__ called
__getattr__ called
None
```

So I think `__getattribute__` is what you want.  (I'm a little bit sad of the two calls before 10.0, but it makes sense why it happens: one for the `.twox`, and one for the `self.val` inside `twox()`.

pipeline/data_store.py
```diff
@@ -343,6 +393,19 @@ def __getattribute__(self, key):
...
+        """Update the report object with the latest results from a processing step that just finished. """
+        self.report = self.report.scan_datastore(self, process_step=process_step, session=session)
+
+    def finalize_report(self, session=None):
+        """Mark the report as successful and set the finish time."""
+        self.report.success = True
+        self.report.finish_time = datetime.datetime.utcnow()
+        with SmartSession(session) as session:
+            new_report = session.merge(self.report)
+            session.commit()
```
rknop, line 406: Is this the only time the report is sent to the database?

This might be fine -- if it dies, it'll send a report with the exception and status as to how far it got.

This does mean that the reports table isn't usable as a "progress monitor", to see how far along things are.  That kind of thing would be useful in catching processes that have gotten stuck without exiting.

pipeline/data_store.py
```diff
@@ -429,7 +495,7 @@ def get_provenance(self, process, pars_dict, upstream_provs=None, session=None):
```
rknop, line 495: Expanding the code up a bit from this diff -- it looks like it tries (a) to find a provenance in `upstream_provs` (which is initialized from `self.upstream_provs`, (b) to pull a provenance from a object produced by the right step, and (c) if that fails, to just the "latest" one from the database.

Before that last step, shouldn't it do a recursive call to DataStore.get_provenance, passing the upstream step as the process and the appropriate thing from the config file as `pars_dict`?  I'd have to think harder about this, but upon first glance it seems to me that the configured upstream parameters ought to somehow come into the fallback provenance generation.

pipeline/measuring.py
```diff
@@ -139,166 +141,184 @@ def run(self, *args, **kwargs):
...
+
+            self.pars.do_warning_exception_hangup_injection_here()
+
+            # get the provenance for this step:
+            prov = ds.get_provenance(self.pars.get_process_name(), self.pars.get_critical_pars(), session=session)
+
+            # try to find some measurements in memory or in the database:
+            measurements_list = ds.get_measurements(prov, session=session)
+
+            # note that if measurements_list is found, there will not be an all_measurements appended to datastore!
```
rknop, line 178: Thinking ahead -- when we get to the CutoutsFile, all the stuff that would get stored in measurements (or the merged measurements/cutouts table) should be in the hdf5 cutouts file, so they are still accessible somewhere.

pipeline/parameters.py
```diff
@@ -121,6 +123,31 @@ def __init__(self, **kwargs):
             "verbose", 0, int, "Level of verbosity (0=quiet).", critical=False
         )
 
+        self.inject_warnings = self.add_par(
```
rknop, line 126: These, and `do_warning_exception_hangup_injection_here` , are just for testing purposes, yes?

(I have failed to come up with a reason why we might ever want to use this in production.)

pipeline/top_level.py
```diff
@@ -82,45 +112,173 @@ def __init__(self, **kwargs):
...
+            raise RuntimeError(f'Cannot get reference exposure {ds.exposure.filepath}!') from e
+
+        try:  # create (and commit, if not existing) all provenances for the products
+            with SmartSession(session) as dbsession:
+                provs = self.make_provenance_tree(ds.exposure, session=dbsession, commit=True)
+        except Exception as e:
+            raise RuntimeError('Failed to create the provenance tree!') from e
+
+        try:  # must make sure the report is on the DB
+            report = Report(exposure=ds.exposure, section_id=ds.section_id)
```
rknop, line 186: So, if I understand correctly, every time we run the top level pipeline, a new report gets created, yes?

(This sounds like a reasonable way to do it.)

pipeline/top_level.py
```diff
@@ -131,3 +289,107 @@ def run_with_session(self):
         with SmartSession() as session:
             self.run(session=session)
 
+    @staticmethod
+    def get_reference(filter, target, obs_time, session=None):
```
rknop, line 293: I'd probably just do the `*args, **kwargs` thing here, rather than spelling out the arguments.  The docs can refer to the docs on `Reference.get_reference`.  (This saves having to edit the parameter list in multiple places if it ever changes... which it will when you merge in the code for reference finding that I wrote in the stress test.)

pipeline/data_store.py
rknop: (In practice, this will not matter, at least when running the full pipeline, because of ` make_provenance_tree` in `top_level.py`.)

requirements.txt
```diff
@@ -5,7 +5,7 @@ astropy==5.3.4
 astroquery==0.4.6
 beautifulsoup4==4.12.2
 fitsio==0.9.12
-flaky==3.7.0
+flaky==3.8.1
```
rknop, line 8: Seems if it's flaky, we ought to randomly install a different version every time....

rknop (commented):
This looks good.  There are a lot of small comments -- some are questions/musing, a few are things that should be fixed.  There is some merge work to do to make it consistent with the merge from the stress test (esp. with regard to reference finding).

I have a couple of bigger questions/comments, mostly about what we expect the Report class to be for.

models/enums_and_bitflags.py
```diff
@@ -320,7 +320,7 @@ def string_to_bitflag(value, dictionary):
             original_keyword = keyword
             keyword = EnumConverter.c(keyword)
             if keyword not in dictionary:
-                raise ValueError(f'Keyword "{original_keyword}" not recognized in dictionary')
+                raise ValueError(f'Keyword "{original_keyword.strip()}" not recognized in dictionary')
```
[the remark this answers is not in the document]
guynir42: Yeah that's what I was thinking but we also remove all spaces (leading or otherwise) in the `convert` method. So this is actually going to lead you down the wrong path if you leave the spaces (as it did for me).

models/enums_and_bitflags.py
```diff
@@ -409,3 +409,36 @@ class BitFlagConverter( EnumConverter ):
...
+process_steps_dict = {
+    1: 'preprocessing',  # creates an Image from a section of the Exposure
+    2: 'extraction',     # creates a SourceList from an Image, and a PSF
+    3: 'astro_cal',      # creates a WorldCoordinates from a SourceList
+    4: 'photo_cal',       # creates a ZeroPoint from a WorldCoordinates
+    5: 'subtraction',    # creates a subtraction Image
+    6: 'detection',      # creates a SourceList from a subtraction Image
+    7: 'cutting',        # creates Cutouts from a subtraction Image
+    8: 'measuring',      # creates Measurements from Cutouts
+    # TODO: add R/B scores and maybe an extra step for finalizing a report
```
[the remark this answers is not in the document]
guynir42: Yeah we've had this discussion lately. I think users expect to be able to search through alerts on a database, it is just not clear who will be managing it. I don't think it will cost us much to have an alerts database since we already keep all the data on the objects and measurements, and the alerts will just point to those objects.

models/enums_and_bitflags.py
```diff
@@ -409,3 +409,36 @@ class BitFlagConverter( EnumConverter ):
...
+    2: 'sources',
+    3: 'psf',
+    # 4: 'background',  # not yet implemented
+    5: 'wcs',
+    6: 'zp',
+    7: 'sub_image',
+    8: 'detections',
+    9: 'cutouts',
+    10: 'measurements',
+    # 11: 'rb_scores',
```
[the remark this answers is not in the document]
guynir42: I don't think so. I was thinking about this list as a progress bar that fills up with things made by the pipeline. The reference would be a fixture that is needed before you even start.

models/exposure.py
```diff
@@ -735,6 +745,47 @@ def get_downstreams(self, session=None):
...
+                                continue
+                            setattr(found_exp, col, getattr(self, col))
+                        exposure = found_exp
+
+                    break  # if we got here without an exception, we can break out of the loop
+                except IntegrityError as e:
+                    # this could happen if in between the query and the merge(exposure)
+                    # another process added the same exposure to the database
+                    if 'duplicate key value violates unique constraint "ix_exposures_filepath"' in str(e):
+                        print(str(e))
```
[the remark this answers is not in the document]
guynir42: yeah, I think I just accidentally left it like that.

models/provenance.py
```diff
@@ -338,6 +340,32 @@ def get_code_version(cls, session=None):
                 code_version = session.scalars(sa.select(CodeVersion).order_by(CodeVersion.id.desc())).first()
         return code_version
 
+    def merge_concurrent(self, session=None, commit=True):
```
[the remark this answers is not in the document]
guynir42: yes in principle it would be the same, although it might have a different `is_testing` but I don't think that would be different between instances.

models/reference.py
```diff
@@ -273,3 +273,76 @@ def merge_all(self, session):
...
+        """
+        return (
+                (ref.validity_start is None or ref.validity_start <= obs_time) and
+                (ref.validity_end is None or ref.validity_end >= obs_time) and
+                ref.filter == filter and ref.target == target and
+                ref.is_bad is False
+        )
+
+    @staticmethod
+    def get_reference(filter, target, obs_time, session=None):
```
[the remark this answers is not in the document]
guynir42: Yes I need to merge the two methods.

models/report.py
```diff
@@ -0,0 +1,346 @@
...
+    bitflag_to_string,
+    string_to_bitflag,
+    process_steps_dict,
+    process_steps_inverse,
+    pipeline_products_dict,
+    pipeline_products_inverse,
+)
+
+
+class Report(Base, AutoIDMixin):
```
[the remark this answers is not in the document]
guynir42: Yes, the reports are only a way for things like the webapp to know what is going on. We can remove them periodically. 
There could be multiple reports of the same image with the same provenance (if you had an error and restarted the pipeline). 
We might decide we do want to keep a subset of the reports as way to mark which exposures have been "finished" but we would want to periodically prune the reports and leave (a) the correct provenance (b) the latest run that finished successfully. Then we'd have a checklist of what was successfully processed and stored. Or you can do that with a similar, smaller table that says something like StoredInDataRelease and only tells you which exposures were done (and maybe point you to the provenances that are relevant).

models/report.py
```diff
@@ -0,0 +1,346 @@
...
+                self.append_products_exist(prod)
+        
+        self.products_committed = ds.products_committed
+        
+        # store the runtime and memory usage statistics
+        self.process_runtime = ds.runtimes  # update with new dictionary
+        self.process_memory = ds.memory_usages  # update with new dictionary
+
+        # parse the warnings, if they exist
+        if isinstance(ds.warnings_list, list):
```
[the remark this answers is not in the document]
guynir42: yeah I haven't thought of that... 

Maybe something like `***|***|***` and then a new line, would be very readable by people as a line separator...

models/report.py
```diff
@@ -0,0 +1,346 @@
...
+        return new_report
+
+    @staticmethod
+    def read_warnings(process_step, warnings_list):
+        """Convert a list of warnings into a comma separated string. """
+        formatted_warnings = []
+        for w in warnings_list:
+            text = f'{process_step}: {w.category} {w.message} ({w.filename}:{w.lineno})'
+            formatted_warnings.append(text)
+            _logger.warning(text)  # make sure warnings also get printed to the log/on screen.
```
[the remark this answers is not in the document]
guynir42: yes I will do that

pipeline/cutting.py
```diff
@@ -46,75 +49,93 @@ def run(self, *args, **kwargs):
...
-
-        # make sure this is returned to be used in the next step
-        return ds
+                ds, session = DataStore.from_args(*args, **kwargs)
+        except Exception as e:
+            return DataStore.catch_failure_to_parse(e, *args)
+
+        try:
+            t_start = time.perf_counter()
+            if os.getenv('SEECHANGE_TRACEMALLOC') == '1':
```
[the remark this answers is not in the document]
guynir42: I think I had a method like that in skyportal, so yes I agree we should implement it.

pipeline/data_store.py
guynir42: You are correct!

pipeline/data_store.py
guynir42: The report also gets pushed to DB on every call to `scan_datastore` which happens after each processing step.

pipeline/data_store.py
guynir42: yes that would just be pretty complicated as the datastore doesn't have access to all the pipeline objects/configuration to do that sort of thing backwards. 
I don't particularly like this fallback but I couldn't find something less complicated to make it work. We could just make a place to put the whole provenance tree before you start, but that would make it hard to just get back a datastore from a single pipleine object, i.e., `ds = preprocessor(exposure, section_id)` would not work that way.

pipeline/measuring.py
guynir42: I'm not sure I want to store the measurements along with the cutouts file in one process. That would mean that if you want to re-run the measuring part, you'd need to make a new cutouts HDF5 file. Seems a bit wasteful. Maybe it would be better to have a step for making cutouts and putting them into a file, and then making measurements from that file. The HDF5 files are like 10MB so they are not negligible.

pipeline/parameters.py
guynir42: Just for testing.

pipeline/top_level.py
guynir42: yes.

pipeline/top_level.py
guynir42: good idea.

requirements.txt
guynir42: :)

models/enums_and_bitflags.py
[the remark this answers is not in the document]
rknop: I don't see the `.strip()` in `convert`.  (I looked for it before making this comment.)  What am I missing?

pipeline/measuring.py
rknop: Hmm.  OK, here are my goals:

(1) The measurements table in the database only has the measurements that passed the preliminary cuts.  This is to keep that table manageable; at least with the stuff we saw in the stress test, doing that cuts down the length of that table by two orders of magnitude.  This is a table I expect us to be searching fairly frequently (any thing that does a search for n detections with xyz properties is going to search it), there's value in not letting this table bloat.

(2) Having the measurements available *somewhere*, in case we want to think about what happened with the preliminary cuts.  From a "at the end" point of view, having them in the cutouts file seems like the most obvious place.  But, you're right, that would mean redoing cutouts whenever redoing measurements.

So, more thought required.

models/enums_and_bitflags.py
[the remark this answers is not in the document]
guynir42: There is a `replace(' ', '')` command in the `c` function.

.github/workflows/run-model-tests-1.yml
```diff
@@ -59,4 +59,5 @@ jobs:
 
       - name: run test
         run: |
-          TEST_SUBFOLDER=models docker compose run runtests
+          shopt -s nullglob
+          TEST_SUBFOLDER=$(ls tests/models/test_{a..l}*.py) docker compose run runtests
```
rknop, line 63: I'm guessing this split is because the model tests have been slow?

I think the reason is the NOIRLab server; I've been seeing hangs trying to get images from there.  One thing we might want to do is separate out the tests of pulling exposures from that server, and have them run only when an env var is set; then, by default, disable that env var, so the github tests won't hang up when the NOIRLab server is slow.  The exposure(s) we need for the bulk of the decam tests can go to one of our archives where we store our other test images and such.

models/report.py
```diff
@@ -311,7 +319,7 @@ def scan_datastore(self, ds, process_step, session=None):
             if self.warnings is None or self.warnings == '':
                 self.warnings = new_string
             else:
-                self.warnings += ', ' + new_string
+                self.warnings += '\n***|***|***\n' + new_string
```
rknop, line 322: Now I understand what you meant in that comment; I was confused at first, because I thought `"***"` meant a warning, not literal `"***"`.  I was confused as to why three warnings separated by |, then a newline.  This here makes much more sense.

.github/workflows/run-model-tests-1.yml
rknop, line 63: I'm guessing this is because the model tests have been slow?  I think it may be because the NOIRLab server has been having problems.  It's been slow for the last several days; the problems we had last week with the gaia stars may have been all part of the same thing.

I made Issue #289 suggesting that we remove NOIRLab dependence from most of the tests, and hid the ones that still have it behind either an env var that skips them by default, or a mark that we expect it to fail sometimes.

rknop (approved):
This looks good.

.github/workflows/run-model-tests-1.yml
guynir42: I don't think I'm downloading anything from NOIRLab anymore. The tests are slow because there are a lot of them and many of them produce an entire datastore for the test. I might look into shortening that.

models/report.py
guynir42: Yeah I didn't think I made it very clear. But I think this will be good.

PR #269, opened 2024-05-09
Configurable image names in docker-compose.yaml

rknop:
… so that multiple people working on the same machine don't have to share the same images.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -15,13 +15,13 @@ but in the meantime, install Docker Engine instead of Docker Desktop; instructio
 
 The `devshell` directory has a docker compose file that can create a development environment for you.  To set it up, you need to set three environment variables.  You can either manually set these with each and every `docker compose` command, you can set them ahead of time with `export` commands, or, recommended, you can create a file `.env` in the `devshell` directory with contents:
 ```
+  IMGTAG=<yourname>_dev
   COMPOSE_PROJECT_NAME=<yourname>
   USERID=<UID>
   GROUPID=<GID>
 ```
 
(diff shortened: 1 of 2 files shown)


docs/setup.md
```diff
@@ -70,13 +70,19 @@ TBD
...
 ```
-   export GITHUB_REPOSITORY_OWNER=<yourname>
+   export IMGTAG=<yourname>_tests
    export USERID=<uid>
    export GROUPID=<gid>
    docker compose build
    COMPOSE_PROJECT_NAME=<yourname> docker compose run runtests
 ```
-where you replace `<uid>` and `<gid>` with your own userid and groupid; if you don't do this, the tests will run, but various pycache files will get created in your checkout owned by root, which is annoying.  `<yourname>` can be any string you want.  If you are working on a single-user machine, you can omit the `COMPOSE_PROJECT_NAME` variable; the purpose if it is to avoid colliding with other users on the same machine.
+where you replace `<uid>` and `<gid>` with your own userid and groupid; if you don't do this, the tests will run, but various pycache files will get created in your checkout owned by root, which is annoying.  `<yourname>` can be any string you want.  If you are working on a single-user machine, you can omit the `IMGTAG` and `COMPOSE_PROJECT_NAME` variables; the purpose if it is to avoid colliding with other users on the same machine.  To avoid tying this all the time, you can create a file `.env` in the `tests` subdirectory with contents:
```
guynir42, line 79:
```suggestion
where you replace `<uid>` and `<gid>` with your own userid and groupid; 
if you don't do this, the tests will run, but various pycache files will get created in your checkout owned by root, which is annoying.  
`<yourname>` can be any string you want.  If you are working on a single-user machine, you can omit the `IMGTAG` and `COMPOSE_PROJECT_NAME` variables; 
the purpose of it is to avoid colliding with other users on the same machine.  
To avoid typing this all the time, you can create a file called `.env` in the `tests` subdirectory with contents:
```

guynir42 (approved):

PR #278, opened 2024-05-15
Stresstest tweaks

rknop:
Changes include:

* util/logger.py adds a SCLogger class.  Use that class in place of _logger everywhere, remove _logger from models/base.py
* Add a calibfile_downloadlock table to allow locking of calibfiles for a single instrument/section/filter so that two processes won't try to import them at once, and use that table in instruments.py.  This is a temporarily solution, as it's just one examle of a bigger problem that we plan to solve a different way, but it's there for now as I needed it during the stress test.
* Add mpich to the Dockerfile.  (Ended up not using it, but did use MPI for a while during the buildup to the stress test to see if it behaved differently then python multiprocessing for some things.)
* Everything in hacks/rknop ; these are the things I wrote to actually do the stress test.  No tests written for these; the "hacks" directory should tell you that.
* In alignment, futz around a bit with swarp virtual memory (file paths and limits).
* In improc/scamp.py, add the ability for max_sources_to_use to be a list; if it's a list instead of an integer, iterate through it in an attempt to get a good solution.  (This was needed to get successful astrometric solutions for a handful of the chips.)
* In models/image.py add set_corners_from_header_wcs.  Also now call this during the overscan/trim part of preprocessing, and at the end of astro_cal if the image is flagged as not having a first astrometric calibration done.
* A commented-out function in provenance.py that tried to solve the database save race condition.  (Didn't work, because it lead to lazy load failures with objects not properly attached to sessions.)
* In data_store.py, add an _overlap_frac method that is now used in get_reference.  get_reference defaults to looking for the same instrument and filter, with a 85% overlap of the search image.  It can optionally filter on target name and sensor section.
* Some Spin yaml files for my development purposes
*


pipeline/data_store.py
```diff
@@ -926,43 +1006,80 @@ def get_reference(self, provenance=None, session=None):
...
+        includes DataStore.image.observation_time.
+
+        If minovfrac is given, it will return the reference that has the
+        highest ovfrac.  (If, by unlikely chance, more than one have
+        identical overlap fractions, an undeterministically chosen
+        reference will be returned.  Ideally, by construction, you will
+        never have this situation in your database; you will only have a
+        single valid reference image for a given instrument/filter/date
+        that has an appreciable overlap with any possible image from
+        that instrument.  The software does not enforce this, however.)
```
guynir42, line 1020: I think that since we are dealing with floating point, the only way this could happen is if both references use the exact same image. For images taken "of the same coordinates" the edges will be slightly different, if only because of rounding errors.

pipeline/measuring.py
```diff
@@ -195,6 +194,15 @@ def run(self, *args, **kwargs):
...
                 # remove the bad pixels that we want to ignore
+                # NOTE : this was throwing a RuntimeWarning, which was causing tests to
+                #  fail; not sure why they didn't fail before.  (New version of numpy?  Dunno.)
+                #  There were nans present; not sure whether those should be set to
+                #    "bad pixel" or "out of bounds" (could be either), so choosing "bad pixel".
+                #  Put in "casting='unsafe'" to take care of this.
+                # if np.any( np.isnan( c.sub_flags ) ):
+                #     import pdb; pdb.set_trace()
+                #     pass
+                # c.sub_flags[ np.isnan( c.sub_flags ) ] = BitFlagConverter.convert( "bad pixel" )
```
guynir42, line 205: I suspect this is no longer throwing errors, but I think we should investigate these warnings.

requirements.txt
```diff
@@ -10,6 +10,7 @@ GitPython==3.1.40
 h5py==3.10.0
 healpy==1.16.6
 matplotlib==3.8.2
+mpi4py==3.1.6
```
guynir42, line 13: Did we decide we need MPI or not?

util/logger.py
```diff
@@ -0,0 +1,133 @@
...
+
+    @classmethod
+    def set_level( cls, level=_default_log_level ):
+        """Set the log level of the logging.Logger object."""
+        cls.instance()._logger.setLevel( level )
+
+    @classmethod
+    def setLevel( cls, level=_default_log_level ):
+        """Set the log level of the logging.Logger object."""
+        cls.instance()._logger.setLevel( level )
```
guynir42, line 68: Why do we need both these methods?

util/logger.py
```diff
@@ -0,0 +1,133 @@
...
+            date and the log level ("[yyyy-mm-dd HH:MM:SS - {midformat}
+            - INFO]...").  Useful, for instance, in multiprocessing to
+            keep track of which process the message came from.
+
+        datefmt : string, default '%Y-%m-%d %H:%M:%S'
+            The date format to use, using standard logging.Formatter
+            datefmt syntax.
+
+        level : logging level constant, default logging.WARNING
+            This can be changedl lated with set_level().
```
guynir42, line 111:
```suggestion
            This can be changed later with set_level().
```

util/logger.py
```diff
@@ -0,0 +1,133 @@
...
+
+        """
+        # (This lock is probably not necessary, since memory isn't by default shared in multiprocessing.)
+        # This __init__ will be called rarely (typically once at the beginning of a process), so
+        # the brief lock is not a big deal.
+        # lock = multiprocessing.Lock()
+        # with lock:
```
guynir42, line 118: If you are convinced we will never need this, you can remove commented lines.

util/runner.py
```diff
@@ -0,0 +1,167 @@
+import logging
+from mpi4py import MPI
+
+from util.logger import SCLogger
+
+class Runner:
```
guynir42, line 6: General comment/question on this: It looks to me like this is much more complicated than using python's multiprocessing. What are the advantages of doing it this way?

guynir42 (approved):
I have mostly minor corrections. I think that some of this stuff is going to be really useful in production. But I think we should also have some idea of when we will clean up some of the hacks, after we are sure we don't need them anymore.

alembic/versions/2024_05_09_1947-ec64a8fd8cf3_calibfile_downloadlock.py
```diff
@@ -0,0 +1,52 @@
...
+    sa.Column('id', sa.BigInteger(), autoincrement=True, nullable=False),
+    sa.PrimaryKeyConstraint('id')
+    )
+    op.create_index(op.f('ix_calibfile_downloadlock__calibrator_set'), 'calibfile_downloadlock', ['_calibrator_set'], unique=False)
+    op.create_index(op.f('ix_calibfile_downloadlock__flat_type'), 'calibfile_downloadlock', ['_flat_type'], unique=False)
+    op.create_index(op.f('ix_calibfile_downloadlock__type'), 'calibfile_downloadlock', ['_type'], unique=False)
+    op.create_index(op.f('ix_calibfile_downloadlock_created_at'), 'calibfile_downloadlock', ['created_at'], unique=False)
+    op.create_index(op.f('ix_calibfile_downloadlock_id'), 'calibfile_downloadlock', ['id'], unique=False)
+    op.create_index(op.f('ix_calibfile_downloadlock_instrument'), 'calibfile_downloadlock', ['instrument'], unique=False)
+    op.create_index(op.f('ix_calibfile_downloadlock_sensor_section'), 'calibfile_downloadlock', ['sensor_section'], unique=False)
```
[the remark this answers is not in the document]
rknop: Unique together; individually, none of these will be unique.  I'm not sure it's worth the effort to put together a big unique_together for all of these rows.

models/instrument.py
```diff
@@ -1474,14 +1478,51 @@ def preprocessing_calibrator_files( self, calibset, flattype, section, filter, m
...
+                #   don't find it, and both try to download it at the same time.  Just using database
+                #   locks doesn't work here, because the process of downloading and committing the
+                #   images takes long enough that the database server starts whining about deadlocks.
+                #   So, we manually invent our own database lock mechaism here and use that.
+                #   (One advantage is that we can just lock the specific thing being downloaded.)
+                # This has a danger : if the code fully crashes during the try block below,
+                #   it will leave behind this fake lock.  That should be rare, as the finally
+                #   block that deletes the fake lock will always happen as long as python is
+                #   functioning properly.  But a very badly-timed machine crash could leave one
+                #   of these fake locks behind.
```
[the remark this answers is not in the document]
rknop: Yeah, automatically blanket cleaning up is a a fraught thing with multiple machines running at once.  My guess is that this is going to be rare enough that we can just print out an error if it fails to get a lock after a while, and then provide a way to easily clean out locks that are older than n hours.

pipeline/data_store.py
rknop: Ideally, we won't have two different references defined that point to the same image, because then they'd be the same reference.  (In fact, should image_id be a "unique" field in the Reference table?  Hmm, probably not, because we might want to define the same image as a reference for multiple values of "target", for instance.)

requirements.txt
rknop: "Maybe."  It went in there as an experiment (see below).  I left it in for now to facility further experiments as necessary.

util/logger.py
rknop: This was a bit of laziness on my part.  The SCLogger class evolved a bit as I was working on it.  Originally, I had set_level, but then later when I added the info(), debug(), etc. methods, I could do a bulk replace of _logger with SCLogger.  That's when I remember that logging.logger usese setLevel and not set_level, so I just added the method here so that either would work when used in code.

util/runner.py
rknop: We will probably not actually use this functionality, but this works across machines, as MPI can work across machines.  (I imported this from ... curveball? another pipeline.  I used it in `hacks/rknop/process_decam_exposure_mpi.py` when I was exploring the possibility that some dysfunctionalities were arising from using python multiprocessing rather than MPI (which is what I'd always used in pipelines in the past).  The conclusion was that that was not the problem.  I left this in so the hacks mpi thing would still work; eventually, we may throw this out.

(Also, the only reason using MPI is more complicated is that multiprocessing provides the Pool class.  I originally reimplemented this same functionality using multiprocessing, and it was approximately as complicated, until I smacked myself in the face and remembered the Pool class.)

models/image.py
```diff
@@ -604,10 +603,31 @@ def merge_all(self, session):
...
+            wcs = WCS( self._header )
+        # Try to detect a bad WCS
+        if ( wcs.axis_type_names == ['', ''] ):
+            raise ValueError( "Could not find a good WCS" )
+
         ras = []
         decs = []
+        # Is this really what we want?  Or should we prefer data?
+        # (Look at preprocessing. When it trims, does it modify raw_data?)
         data = self.raw_data if self.raw_data is not None else self.data
```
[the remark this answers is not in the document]
rknop: It runs before preprocessing as well as after preprocessing -- see Image.from_exposure, for instance.

rknop:
Re: cleaning up the hacks : I've found it's often useful to have a git-tracked directory for individual users to put their experiments (things like what I did for stress tests).  As long as we're in development mode, we should have these, but we might want to make a "make version 1.0.0" issue that cleans these out.

(On the other hand: realistically, most scientific pipeline code like this never really goes out of development mode, as the people running it are often the same as the people developing it.  The stuff that I'm doing for the Roman space telescope supernova PIT ("project infrastructure team") is different, as eventually we're supposed to deliver stuff to another group for operations, but that's on a whole different scale.)

PR #280, opened 2024-05-16
Replace binary dilate structure from square to circular (no diagonal)

guynir42:


rknop (approved):
Just a comment in one docstring.

PR #281, opened 2024-05-16
Add parameter to control timeout for scamp and swarp

guynir42:


improc/alignment.py
```diff
@@ -350,7 +366,7 @@ def _align_swarp( self, image, target, sources, target_sources ):
                         '-WRITE_XML', 'N' ]
 
             t0 = time.perf_counter()
-            res = subprocess.run( command, capture_output=True, timeout=60 )
+            res = subprocess.run(command, capture_output=True, timeout=self.pars.scamp_timeout)
```
rknop, line 369: This should be `self.pars.swarp_timeout`

Above, line 305, in the call `swarp_fodder_wcs = improc.scamp.solve_wcs_scamp(`, you should add passing self.pars.scamp_timeout as an argument.

rknop (commented):
one small fix needed

improc/alignment.py
```diff
@@ -350,7 +367,7 @@ def _align_swarp( self, image, target, sources, target_sources ):
...
```
rknop, line 370:
```suggestion
            res = subprocess.run(command, capture_output=True, timeout=self.pars.swarp_timeout)
```

rknop (approved):
Still have one `self.pars.scamp_timeout` that should be `self.pars.swarp_timeout` (I put in a suggested change).

PR #286, opened 2024-05-21
Add option to get Gaia DR3 from a custom server we run at Spin that pulls Gaia data from files stored on Nersc

rknop:


tests/seechange_config_test.yaml
```diff
@@ -20,6 +20,9 @@ archive:
   local_write_dir: null
   token: insecure
 
+catalog_gaiadr3:
+  fallback_datalab: False
```
guynir42, line 24: Why not set this to True by default? Could be useful to have our tests not fail when the server is down...

guynir42 (approved):
Very nice. I only have one question but after that you can merge.

tests/seechange_config_test.yaml
rknop: Yeah, will do.  (Obviously, had this there in the first place to make sure I was testing the new server.)

Issue #292, opened 2024-05-26
Measurements and Cutouts


guynir42:
Here's a summary of the debate around measurements going into the cutouts process/file: 

Problem: we want to keep measurements around, even if they failed the cuts, so we can examine them and figure out if our cuts make sense. But that would mean bloating the measurements table, which is likely to be the longest one in the database. 

Option 1: Cutouts will have a single file, single row in the cutouts table, and will be made in a separate process than measurements. We save all cutouts. Then we only save measurements that have passed the cuts. Advantage: we can make one cutouts file and many versions of the measurements. Disadvantage: we don't get to look at the measurements that failed the cuts. 

Option 2: Cutouts will have a single file, single row in the cutouts table, but will be generated in the same process (i.e., will have the same provenance) as the measurements. We still save only the passing measurements to DB but will put a copy of the results in the cutouts file, such that all cutouts and all measurements get saved to disk, regardless of cuts. Advantage: we can look at the bad measurements, inside the cutouts file. Disadvantage: we must make new cutouts files each time we change a parameter on the measurements. We will have to do that anyway when we change code or parameters upstream of cutouts (e.g., on subtraction). 

I have thought about this a little and I have a third proposal that I think answers all our needs: 

Option 3: Do option 1, but add a secondary threshold dictionary. That one decides which measurements are deleted, and can be set to very loose thresholds if you want to see "everything". The regular threshold will no longer delete measurements, but instead mark them as "bad". You'll have to mark the associated Objects as bad, too. We can decide we are keeping all the "bad" objects separate from the good ones, and that gives us the advantage that we can check if we systematically have "bad" objects that have multiple measurements over time (that would be suspicious). 

Then, we have an easy way to look at the bad measurements and periodically clean them up from our database. This gives the sysadmin more control over the bloat of that table, and makes it easier to find the "bad" measurements and inspect a subsample to check our thresholds are not too strict. 

If, at some point, you wish to change this policy and just go back to deleting all the bad measurements you'd just have to adjust the deletion threshold to match the regular one (we can decide that if the delete threshold is None we just use the regular threshold for both cuts). 

We can even make the deletion threshold a non-critical parameter, so changing it would not alter the provenance of the "good" measurements, that would continue to be saved regardless of the save status of "bad" measurements.

whohensee:
Nice, this is a great summary. I will look into the relevant part of the codebase to prepare and get an idea of what implementing it will involve.

PR #295, opened 2024-05-28
measurements and cutouts

whohensee:
See the summary at the top of #292


whohensee (commented):
Some choices I made here that I am uncertain of or could use feedback on:
- I set the is_bad column to have no default with the idea that this would protect us from accidentally saving objects without ensuring they are the right type. However, perhaps a default of `is_bad=True` would be better for convenience of pipeline use.
- I broke the `passes` method, which used to simply decide whether or not an object passed the disqualifying cuts, into two methods. Since there are now three possibilities due to the two thresholds, `compare_to_thresholds` returns a string value of `"ok"`, '"bad"', or '"delete"' in addition to updating the `is_bad` column attribute, and `passes_deletion_thresholds` calls this function and returns a Bool for use in the pipeline to determine whether and object should be saved or not. I used strings here for code clarity, but perhaps 0, 1, -1 or similar could work too. 
- I made it so the `deletion_threshold` dict can be a dict or None, with slightly different logic for `None` compared to `{}`. In the case of None, the default, the regular `threshold` dict will be used as `deletion_threshold`. In the case of explicity setting `{}`, then no deletion will occur and all measurements will pass to be saved. I thought this could maybe save the effort of coming up with unrealistic thresholds manually, but it might be overkill

guynir42 (commented):
I have lots of comments, mostly minor ones. 
But there is one thing that will complicate your life: the requirement that the deletion threshold is not a critical parameter (which I think is important). This will have the unfortunate effect of removing that dictionary from the provenance, which will make you have to rethink the way you check against this (and maybe the other) threshold.

models/measurements.py
```diff
@@ -359,15 +366,63 @@ def find_cutouts_in_list(self, cutouts_list):
...
+
+        for key, value in mark_thresh.items():
+            if deletion_thresh.get(key) is not None and self.disqualifier_scores[key] >= deletion_thresh[key]:
+                passing_status =  "delete"
+                break
+            if mark_thresh[key] is not None and self.disqualifier_scores[key] >= mark_thresh[key]:
+                # I didn't use get here since its required - should fire an error if key isn't in the dict I think
+                passing_status = "bad" # no break because another key could trigger "delete"
+
+        self.is_bad = not passing_status == "ok"
```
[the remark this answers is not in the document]
whohensee: Ah yeah, this is good. I originally wrote `not (x==y)` but my linter got annoyed at me for unnecessary parenthesis, but `!=` just seems better.

models/measurements.py
```diff
@@ -359,15 +366,63 @@ def find_cutouts_in_list(self, cutouts_list):
...
+        deletion_thresh = (mark_thresh if self.provenance.parameters['deletion_thresholds'] is None
+                           else self.provenance.parameters['deletion_thresholds'])
+
+        for key, value in mark_thresh.items():
+            if deletion_thresh.get(key) is not None and self.disqualifier_scores[key] >= deletion_thresh[key]:
+                passing_status =  "delete"
+                break
+            if mark_thresh[key] is not None and self.disqualifier_scores[key] >= mark_thresh[key]:
+                # I didn't use get here since its required - should fire an error if key isn't in the dict I think
+                passing_status = "bad" # no break because another key could trigger "delete"
```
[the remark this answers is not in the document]
whohensee: In the init of `ParsMeasurer`, a value is set for the `thresholds` parameter that is hard coded in and then overwritten by the kwargs from the config file, so I guess this would be possible only if the user overwrote that dictionary with one missing keys. I can loop over both just in case.

tests/fixtures/pipeline_objects.py
```diff
@@ -747,7 +747,7 @@ def make_datastore(
                 ds.all_measurements = Measurements.copy_list_from_cache(cache_dir, cache_name)
                 [setattr(m, 'provenance', prov) for m in ds.all_measurements]
                 [setattr(m, 'cutouts', c) for m, c in zip(ds.all_measurements, ds.cutouts)]
-                ds.measurements = [m for m in ds.all_measurements if m.passes()]
+                ds.measurements = [m for m in ds.all_measurements if m.passes_deletion_threshold()]
```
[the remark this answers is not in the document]
whohensee: Note to self mainly: I *have* changed this segment of code to work with the new changes, but I have not yet looked to see whether or not the other tests can be improved afterwards.

tests/models/test_measurements.py
```diff
@@ -170,6 +170,52 @@ def test_measurements_cannot_be_saved_twice(ptf_datastore):
...
+
+    m.disqualifier_scores['negatives'] = 0.1 # set a value that will pass both
+    assert m.compare_to_thresholds() == "ok"
+
+    m.disqualifier_scores['negatives'] = 0.9 # set a value that will fail
+    assert m.compare_to_thresholds() == "bad"
+
+    # I'd also like to test that deletion_thresholds is a proper non-critical param
+    # so that it can be changed without affecting the provenance
+    # However, it seems like prov.id changes even when varying non-critical params
```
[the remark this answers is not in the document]
whohensee: I went with a version of option 1 (pending discussion/review):

I decided to put a method on `Measurer` `compare_measurement_to_thresholds(self, m)` which uses the non-critical deletion threshold dictionary of that measurer object and the `mark_threshold` dictionary from the Measurements object's provenance.

I wasn't sure the best spot to put it when I first made it a method of `Measurer`, but it could also have been a method of `Measurements` (in which case the deletion dictionary would need to be passed to it instead) or even a utility function (I didn't like this because it felt separated from the relevant code).

I will continue to ponder the best implementation.

pipeline/measuring.py
```diff
@@ -414,3 +425,36 @@ def make_filter_bank(self, imsize, psf_fwhm):
...
+        passing_status = "ok"
+
+        mark_thresh = m.provenance.parameters["thresholds"] # thresholds above which measurement is marked 'bad'
+        deletion_thresh = self.pars.deletion_thresholds
+
+        # I think this accomplishes what we wanted without having to fully loop over both dictionaries
+        combined_keys = np.unique(list(mark_thresh.keys()) + list(deletion_thresh.keys())) # unique keys from both
+        for key in combined_keys:
+            # check if it exists in mark_ but not deletion_. If so, use the value from mark_ for deletion_
+            if deletion_thresh.get(key) is None and mark_thresh.get(key) is not None:
```
whohensee, line 451: Of note with this implementation: You can no longer set the entire `deletion_thresholds` parameter to `None` (it only accepts type `dict` so the equivalent is `{}`), and any missing deletion parameters will default to the value in the `thresholds` dict, regardless of whether they are missing or set to `None`. The only way to change that behavior e.g. if you wanted to turn deletion off would be to explicitly set a value for each disqualifier that would never be exceeded.

whohensee: i *think* this accomplishes what was discussed, with a focus on making it simple to make the deletion behavior just use the normal threshold when desired, and possible but not completely trivial to disable deletion if that is really desired.

tests/models/test_measurements.py
```diff
@@ -170,6 +170,57 @@ def test_measurements_cannot_be_saved_twice(ptf_datastore):
...
+    m.disqualifier_scores['negatives'] = 0.1 # set a value that will pass
+    assert measurer.compare_measurement_to_thresholds(m) == "ok"
+
+    m.disqualifier_scores['negatives'] = 0.4 # a value that would fail mark
+    assert measurer.compare_measurement_to_thresholds(m) == "bad"
+
+    m.disqualifier_scores['negatives'] = 0.9 # a value that would fail both (earlier)
+    assert measurer.compare_measurement_to_thresholds(m) == "bad"
+
+    # test deletion_threshold varies as a non-critical parameter
```
whohensee, line 216: I think this section below might be a bad, trivial test of the non-critical parameter. Perhaps the proper way would be to actually use `measurer.run()` with the deletion dictionary set to two different values and ensure the provenances of resulting measurements objects come out the same. That might deserve moving to a separate test, if we think its a test worth writing.

whohensee: I added what I think may be a better test below for review.

pipeline/measuring.py
```diff
@@ -414,3 +425,36 @@ def make_filter_bank(self, imsize, psf_fwhm):
...
+        Inputs:
+          - m : a Measurements object to be compared
+
+        returns one of three strings to indicate the result
+          - "ok"     : All disqualifiers below both thresholds
+          - "bad"    : Some disqualifiers above mark_thresh but all 
+                       below deletion_thresh
+          - "delete" : Some disqualifiers above deletion_thresh
+        
+        """
```
whohensee, line 441: I will change the logic of this function soon following our Friday meeting, such that 
- you can pass `None` as a deletion_threshold, causing it to use the regular threshold as deletion thresh.
- you can pass an empty dict as deletion_threshold, causing it to do no deletion checking
- any keys missing in either dictionary will simply not be checked in that respective check (deletion or marking)

pipeline/measuring.py
whohensee: Updated as per Friday meeting discussion

default_config.yaml
```diff
@@ -129,7 +129,12 @@ measuring:
     negatives: 0.3
     bad pixels: 1
     offsets: 5.0
-    filter bank: 1.0
+    filter bank: 1
+  deletion_thresholds:
```
whohensee, line 133: I could also set the deletion thresholds to None to mimic current behavior instead of an identical dict like this

pipeline/measuring.py
```diff
@@ -365,8 +364,9 @@ def run(self, *args, **kwargs):
 
                 saved_measurements = []
                 for m in measurements_list:
-                    # passes_deletion_threshold() method sets the is_bad attribute of m rather than doing it explicitly here
-                    if m.passes_deletion_threshold():  # all disqualifiers are below threshold
+                    threshold_comparison = self.compare_measurement_to_thresholds(m)
+                    if threshold_comparison != "delete":  # all disqualifiers are below threshold
+                        m.is_bad = threshold_comparison == "bad"
```
guynir42, line 369: Nice.

pipeline/measuring.py
```diff
@@ -425,3 +425,33 @@ def make_filter_bank(self, imsize, psf_fwhm):
...
+          - "delete" : Some disqualifiers above deletion_thresh
+        
+        """
+        passing_status = "ok"
+
+        mark_thresh = m.provenance.parameters["thresholds"] # thresholds above which measurement is marked 'bad'
+        deletion_thresh = ( mark_thresh if self.pars.deletion_thresholds is None
+                           else self.pars.deletion_thresholds )
+
+        # I think this accomplishes what we wanted without having to fully loop over both dictionaries
```
guynir42, line 448:
```suggestion
```

guynir42: Yes, this is good.

guynir42 (approved):
Looks good! Feel free to squash and merge.

Issue #257, opened 2024-04-26
Merge WCS and ZP into extraction step


guynir42:
To help simplify our versioning / provenance model we should combine all the steps after producing a preprocessed image and before subtraction into a single step. 
This includes:
- Source extraction
- PSF extraction
- Background calculation
- Star vs. galaxy (or other junk) identification using a catalog
- Astrometric calibration
- Photometric calibration

I suggest we just continue to call this "extraction" but I'm open to other ideas for the name. 

The upside is that we simplify things dramatically (see Will's PR on upstreams/downstreams). 
The downside is that when we want to only change the astrometric/photometric calibration then we'd have to (a) recalculate stuff like sextractor and (b) have to re-save another source list, psf and background objects, instead of just re-running the crossmatch and saving just the WCS or ZP. 
I think this is well worth the simplification.

guynir42:
Addressed by #296

Issue #205, opened 2024-02-22
Background object


guynir42:
The thing we need is a background object to go along with the Image object. 
This allows us to do a few things:
1) Hold on to different types of background data, including a map the same size as the image, a 2D polynomial where we only keep the coefficients, a constant value, or even "zero" (which means the background was already subtracted for this Image). 
2) We can save to database and disk the results separately from the Image. 
3) We can use Provenance to track versions of the background subtraction code/parameters. 
4) We can lazy load the background data whenever we want to get the background subtracted image. 

There are some problems with this model: 
a) If the background is a scalar or zero, would we still save a file? If not, how to square this with the FileOnDiskMixin class requiring a filepath. 
b) Where exactly does the background calculation happen? Is it in preprocessing, in extraction, or in a separate process in between those two?
c) Do we add another process with it's own provenance? Or should we append the background as a secondary product of one of the existing processes (preprocessing/extraction)?

I think it will be necessary to save the parameter choices for background estimation, because at some point we will need to allow different, competing algorithms to run. 

However, I am not sure we can separate this from the source extraction. In some cases (e.g., in dense fields where this is the most critical) the background estimation MUST happen along with source extraction, e.g., when using scene modeling. 

I suggest we append the Background product to the extraction process. It already creates two objects that share the same Provenance "process", which are SourceList and PSF. 
The advantage is that we don't need to add another step in the pipeline (that would also require lots of refactoring) but only add an output object. 
The downside is that if you change the background subtraction code you must also re-do extraction and PSF calculation (and vice versa).

guynir42:
Could be a good second issue or third issue.

guynir42:
Addressed by #308

Issue #202, opened 2024-02-09
Datastore upstream provenances


guynir42:
A while back I implemented an `upstream_provs` attribute to allow users to specify which provenance to use as the upstream for the automatically loaded data products (which is needed if you start the analysis from some point in the middle, and not re-run the whole thing). 

I think the interface for this attribute (e.g., in `get_provenance()`) is kinda clumsy and not easy to use. We should maybe replace it with a dictionary and better document what is does.

guynir42:
Addressed by #296

PR #296, opened 2024-05-30
Combine astro_cal and photo_cal into extraction

guynir42:
To simplify the pipeline I've moved the astrometric and photometric calibration steps into the same step as the extraction. 
We already produce both a source list and a psf from this step, we may as well add some more stuff. 
This will include sources, psf, wcs, zp and also a background object. 

I've also made a rehaul of the way we get the provenances from the datastore, and cleaned up some of its getters. 

This PR would hopefully address the following issues: #257, #202, #205, and maybe #154


tests/models/test_objects.py
```diff
@@ -29,7 +29,7 @@ def test_object_creation():
         assert re.match(r'\w+\d{4}\w+', obj2.name)
 
 
-@pytest.mark.flaky(max_runs=3)
+@pytest.mark.flaky(max_runs=5)
```
whohensee, line 32: this seems good, I notice these fail a lot in github

pipeline/data_store.py
whohensee, line 1: I'm a bit unclear on where the "reference" step/product lies in the greater scheme of things, thus a few of these comments in this file.. Is it treated differently because it is created outside of the main pipeline?

pipeline/data_store.py
```diff
@@ -1172,51 +1105,41 @@ def get_subtraction(self, provenance=None, session=None):
...
+                            image_upstreams_association_table.c.downstream_id == Image.id,
+                        )
+                    ).join(
+                        aliased_table,
+                        sa.and_(
+                            aliased_table.c.upstream_id == image.id,
+                            aliased_table.c.downstream_id == Image.id,
+                        )
+                    ).where(Image.provenance.has(id=provenance.id))
+                ).first()
```
whohensee, line 1142: Here was the big query I mentioned in the meeting... I intend to come back to this and look into it some more, its far more complex than anything I have had to do!

whohensee (commented):
This was a big one! I made a bunch of comments of generally minor things, and a lot of them are just questions for my own understanding. Overall I think the structure improvements to the data products are a big improvement!

Let me know if I left too many questions or little comments. I don't want to slow you down with things that are really minor!

models/psf.py
```diff
@@ -527,7 +527,44 @@ def get_upstreams(self, session=None):
         with SmartSession(session) as session:
             return session.scalars(sa.select(Image).where(Image.id == self.image_id)).all()
         
-    def get_downstreams(self, session=None):
-        """Get the downstreams of this PSF (currently none)"""
-        return []
+    def get_downstreams(self, session=None, siblings=False):
+        """Get the downstreams of this PSF.
+
+        If siblings=True (default) then also include the PSFs, WCSes, ZPs and background objects
```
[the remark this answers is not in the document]
guynir42: Thanks!

models/psf.py
```diff
@@ -527,7 +527,44 @@ def get_upstreams(self, session=None):
...
+        that were created at the same time as this source list.
+        """
+        from models.source_list import SourceList
+        from models.world_coordinates import WorldCoordinates
+        from models.zero_point import ZeroPoint
+        from models.provenance import Provenance
```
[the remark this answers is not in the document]
guynir42: I don't think this helps in terms of performance. 

This is the sort of place where if you include all the imports on each one of the classes, you'd get a circular import. 
Usually when I setup the imports there's a clear hierarchy: for example a SourceList knows about an Image but not the other way around (Image is more "basic"), so we can import Image at the start of SourceList. 

But when they need to know about each other, there has to be an import call "when needed" inside a function to avoid the circular import at startup. By the time you call this function, there's a good chance that all the modules are already loaded so it doesn't do anything except put the names of classes into local scope.

models/psf.py
```diff
@@ -527,7 +527,44 @@ def get_upstreams(self, session=None):
...
+            output = subs
+
+            if siblings:
+                # There should be exactly one source list, wcs, and zp per PSF, with the same provenance
+                # as they are created at the same time.
+                sources = session.scalars(
+                    sa.select(SourceList).where(
+                        SourceList.image_id == self.image_id, SourceList.provenance_id == self.provenance_id
+                    )
+                ).first()
```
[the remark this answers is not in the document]
guynir42: I added errors in case it gets not a single object. Can't hurt.

models/source_list.py
```diff
@@ -751,25 +749,39 @@ def get_upstreams(self, session=None):
...
+            ).all()
+            output = subs
+
+            if self.is_sub:
+                cutouts = session.scalars(sa.select(Cutouts).where(Cutouts.sources_id == self.id)).all()
+                output += cutouts
+            elif siblings:  # for "detections" we don't have siblings
+                psfs = session.scalars(
+                    sa.select(PSF).where(PSF.image_id == self.image_id, PSF.provenance_id == self.provenance_id)
+                ).all()
```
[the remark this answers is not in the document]
guynir42: The provenance upstreams contains the upstream provenances, not information about which image was used. The provenance is re-used for all the images and products, so it cannot be sufficient to finding a specific object.

pipeline/coaddition.py
```diff
@@ -601,7 +609,7 @@ def parse_inputs(self, *args, **kwargs):
                 provenance_ids = [prov.id]
             provenance_ids = listify(provenance_ids)
 
-            with SmartSession(session) as session:
+            with SmartSession(session) as dbsession:
```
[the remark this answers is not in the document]
guynir42: yes, exactly.

pipeline/coaddition.py
```diff
@@ -615,19 +623,64 @@ def parse_inputs(self, *args, **kwargs):
...
-        self.datastore = self.photo_cal.run(self.datastore)
+        # TODO: add the warnings/exception capturing, runtime/memory tracking (and Report making) as in top_level.py
+        self.datastore = self.extractor.run(self.datastore)
+        self.datastore = self.astrometor.run(self.datastore)
+        self.datastore = self.photometor.run(self.datastore)
 
         return self.datastore.image
 
+    def make_provenance_tree(self, session=None):
+        """Make a (short) provenance tree to use when fetching the provenances of upstreams. """
```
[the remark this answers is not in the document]
guynir42: Maybe graph would have been a better name for it...

pipeline/data_store.py
```diff
@@ -36,9 +35,7 @@
     'exposure': 'exposure',
     'preprocessing': 'image',
     'coaddition': 'image',
-    'extraction': ['sources', 'psf'],  # TODO: add background, maybe move wcs and zp in here too? 
-    'astro_cal': 'wcs',
-    'photo_cal': 'zp',
+    'extraction': ['sources', 'psf', 'background', 'wcs', 'zp'],
```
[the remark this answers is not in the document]
guynir42: I will be adding it soon, probably in one of the next PRs.

pipeline/data_store.py
```diff
@@ -474,34 +491,34 @@ def get_provenance(self, process, pars_dict, upstream_provs=None, session=None):
...
+                        prov = obj.provenance
 
-                upstreams.append(prov)
+                if prov is not None:  # if we don't find one of the upstreams, it will raise an exception
+                    upstreams.append(prov)
 
             if len(upstreams) != len(UPSTREAM_STEPS[process]):
                 raise ValueError(f'Could not find all upstream provenances for process {process}.')
 
+            for u in upstreams:  # check if "reference" is in the list, if so, replace it with its upstreams
```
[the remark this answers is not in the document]
guynir42: I need the subtraction image's provenance to have access to the provenance of the reference image and its products, so that it can load them when needed from the DB. It is the same as having the provenance of the new image and its products. 
If I just leave the upstream as the provenance of the reference it doesn't help in that regards.

pipeline/data_store.py
```diff
@@ -510,46 +527,34 @@ def get_provenance(self, process, pars_dict, upstream_provs=None, session=None):
...
-            provenance = get_latest_provenance(process, session=session)
+        # see if it is in the prov_tree
+        if self.prov_tree is not None:
+            if process in self.prov_tree:
+                return self.prov_tree[process]
+            else:
+                raise ValueError(f'No provenance found for process "{process}" in prov_tree!')
 
-        return provenance
+        return None  # if not found in prov_tree, just return None
```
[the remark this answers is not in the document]
guynir42: yes, you are right.

pipeline/data_store.py
guynir42: I can barely remember how I got it to work, too...

models/base.py
```diff
@@ -45,6 +45,17 @@
...
+# # printout the list of relevant environmental variables:
+# print("SeeChange environment variables:")
+# for key in [
+#     'INTERACTIVE',
+#     'LIMIT_CACHE_USAGE',
+#     'SKIP_NOIRLAB_DOWNLOADS',
+#     'RUN_SLOW_TESTS',
+#     'SEECHANGE_TRACEMALLOC',
+# ]:
+#     print(f'{key}: {os.getenv(key)}')
```
guynir42, line 58: I'm leaving this comment here as a debug tool.

models/enums_and_bitflags.py
```diff
@@ -418,12 +418,10 @@ class BitFlagConverter( EnumConverter ):
     _dict_inverse = None
 
 
-# the list of possible processing steps from a section of an exposure up to measurments, r/b scores, and report
+# the list of possible processing steps from a section of an exposure up to measurements, r/b scores, and report
 process_steps_dict = {
     1: 'preprocessing',  # creates an Image from a section of the Exposure
     2: 'extraction',     # creates a SourceList from an Image, and a PSF
-    3: 'astro_cal',      # creates a WorldCoordinates from a SourceList
-    4: 'photo_cal',       # creates a ZeroPoint from a WorldCoordinates
```
rknop, line 426: I don't propose actually changing this because it would be painful at this point, but the name 'extraction' really isn't right for this step any more.  Back in the SCP days we would have called this 'reduction', but many astronomers use 'reduction' to mean what we mean by 'preprocessing', so I wouldn't suggest that name.  We're doing more than extraction in this step now, though.  (We were before too, as there was psf estimation.)

models/image.py
```diff
@@ -545,14 +546,14 @@ def merge_all(self, session):
             self.sources.provenance_id = self.sources.provenance.id if self.sources.provenance is not None else None
             new_image.sources = self.sources.merge_all(session=session)
 
-            new_image.wcs = new_image.sources.wcs
-            if new_image.wcs is not None:
+            if new_image.sources.wcs is not None:
```
rknop, line 549: Would there be any value in making image.wcs, image.wcs, image.zp all into `@property` things that point to image.sources.<thing> rather than having them be manually loaded like this?

models/reference.py
```diff
@@ -219,7 +219,7 @@ def load_upstream_products(self, session=None):
 
             sources = session.scalars(
                 sa.select(SourceList).where(
-                    SourceList.image_id == self.image_id,
+                    SourceList.image_id == self.image.id,
```
rknop, line 222: Why not leave this as is?  I guess if you're doing eager loading, then `self.image` will always be loaded, but it seems simpler and more straightforward to just use the `image_id` property.

models/reference.py
```diff
@@ -233,7 +233,7 @@ def load_upstream_products(self, session=None):
 
             psfs = session.scalars(
                 sa.select(PSF).where(
-                    PSF.image_id == self.image_id,
+                    PSF.image_id == self.image.id,
```
rknop, line 236: (here too)

models/source_list.py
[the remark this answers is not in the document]
rknop: On this note... won't 
```
sa.select(Image).where(
                    Image.provenance.has(Provenance.upstreams.any(Provenance.id == self.provenance.id))
                )
```

pull out an ungodly number of images?  It'll be everything that has an upstream provenance including this source list's provenance.  Don't you also need a clause that this image is among the sub images upstreams?

models/world_coordinates.py
```diff
@@ -102,25 +103,56 @@ def get_upstreams(self, session=None):
...
-        downstreams = zps + subs
-        return downstreams
+
+        with (SmartSession(session) as session):
+            subs = session.scalars(
+                sa.select(Image).where(
+                    Image.provenance.has(Provenance.upstreams.any(Provenance.id == self.provenance.id))
+                )
+            ).all()
+            output = subs
```
rknop, line 123: Same question here that I had in source_list -- won't this pull out all images whose provenances includes this world coordiantes' provenance, not just the ones that actually come from this wcs?

models/zero_point.py
```diff
@@ -136,29 +137,57 @@ def get_aper_cor( self, rad ):
...
-                                    .where(Image.provenance
-                                            .has(Provenance.upstreams
-                                                .any(Provenance.id == self.provenance.id)))).all()
-        return subs
+            subs = session.scalars(
+                sa.select(Image).where(
+                    Image.provenance.has(Provenance.upstreams.any(Provenance.id == self.provenance.id))
+                )
+            ).all()
+            output = subs
```
rknop, line 162: Same question here as in source_list and wcs

pipeline/astro_cal.py
```diff
@@ -131,7 +131,7 @@ def __init__(self, **kwargs):
         self.override(kwargs)
 
     def get_process_name(self):
-        return 'astro_cal'
+        return 'extraction'
```
rknop, line 134: I guess I'm not fully clear on all the implications of this.  It's weird to call this 'extraction', given that there's also a 'extraction' process in `detection.py` that really is extraction.

How is `get_process_name()` actually used?  What is this process name for?

It's used in constructing provenances, isn't it?  If so, this has become a little bit tangled... the provenance for extraction now includes all of extraction, wcs, and zp; since the latter two aren't separate steps, they don't get their own provenances, but their parmeters are folded into the extraction provenance.  But, what's the implication of the process_name here not just being astro_cal?

It seems that originally there was a mapping between one of these thingies that had a Parameters subclass and one of the steps of the pipeline, whereas that's not true any more.

pipeline/data_store.py
rknop: Yes.  The subtraction/detection pipeline assumes that all the references it needs exist, and will fail if one doesn't.

We make the references ahead of time.

(I had another pipeline that did just-in-time reference building; if it didn't find a reference, it would go to various surveys and build one.  (In fact, when I inherited the pipeline, it *always* did that, it didn't cache the references, so it was inefficient.)  That led to a mess... where we had the same field that had been subtracted from different references at different times.  You want to keep control of your references, so it makes sense to have the building and registering of those be a separate process rather than something that tries to happen automatically as part of subtraction.)

pipeline/data_store.py
rknop: `curl --data "Please write me sqlalchemy code to join together all my image upstreams correctly" https://chatgpt.org`

rknop (commented):
This mostly looks good, but I do have a couple of questions that are most likely my not understnading something, but that might be a problem.  See the ocmments in source_list.py (plus wcs and zp) about the upstreams search that matches provenance.

models/reference.py
guynir42: I don't remember but there was test where that didn't work. I am not sure what the situation was...

models/source_list.py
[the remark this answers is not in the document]
guynir42: yes, I think you may be right about that!

pipeline/astro_cal.py
guynir42: yes, I guess the get_process_name really was more useful when it was one to one between process and Parameters object. 
I will have another look and maybe we don't actually need this function anymore.

guynir42: I left the function in there, but detached it from the provenances. It is still a useful thing to have for stuff like injecting warnings and having them tagged with an informative "process" name.

rknop (approved):
This looks good.  There's at least one test failiure that's just a change in the warning (your match string has a process name of extraction whereas it should be something else).  There's a second test failure that looked like something in swarp?  I wasn't sure.  As long as those aren't an issue, go ahead and merge.

Issue #297, opened 2024-05-30
Write tests for webap


rknop:

guynir42:
"We'll test it in production"

rknop:
Wrote some.  Selenium is in Dockerfile.  More should be written.  Plus, the webap should do more.

PR #298, opened 2024-05-30
Add the webap used during the stress test

rknop:
No tests written for the webap.  (See Issue #297... pain.)


rknop:
Mostly this code adds the webap.

The other change was I added an env var that tells the tests not to use the cache much.  (It still uses it for things like json and yaml files it creates, so I didn't have to figure out where else to write them.)  As predicted, it slows down the tests a lot.  I also modified the tests that download things from NOIRLab not to keep those in the cache; that turned out to be enough to solve the immediate disk space problem on github, so I haven't enbabled the cache limiting env var by default.

rknop:
(Note that the failed Model Test 1 is one of the test_free things.)

.gitignore
```diff
@@ -12,6 +12,9 @@ data/DECam_default_calibrators
 .pytest.ini
 tests/plots
 tests/temp_data
+webap/seechange_webap_config.py
+webap/static/rkwebutil.js
+webap/install
```
guynir42, line 17: Just curious: don't we need these files on the repo?

tests/docker-compose.yaml
```diff
@@ -89,6 +115,8 @@ services:
         condition: service_completed_successfully
       archive:
         condition: service_healthy
+      # webap:
+      #   condition: service_healthy
```
guynir42, line 119: Can we remove commented lines, or are we bringing them back at some point?

tests/docker-compose.yaml
```diff
@@ -114,6 +142,8 @@ services:
...
```
guynir42, line 146: Same here.

tests/fixtures/cache_utils.py
guynir42, line 1: I'm not sure I am on board with moving these out into the fixtures... I was thinking it may be useful for us to have a way to temporarily dump individual rows to disk and then bring them back up. 

Also, I'm not sure what this code does for objects that are not FileOnDiskMixin, maybe I will see how it works later on.

tests/fixtures/cache_utils.py
```diff
@@ -0,0 +1,254 @@
...
+    which holds the object's column attributes (i.e., only those that are
+    database persistent).
+
+    If caching a non-FileOnDiskMixin object, the filepath argument must be given,
+    because it is used to name the JSON file.
+
+    Parameters
+    ----------
+    FoD: FileOnDiskMixin or None
+        The object (which is an instance of FileOnDiskMixin) to cache.
```
guynir42, line 28: If I'm caching a non-FileOnDiskMixin then you still have to give it the object to turn it into a JSON...

webap/seechange_webap.py
```diff
@@ -0,0 +1,478 @@
...
+            if t0 is not None:
+                q += 'e.mjd >= %(t0)s'
+                subdict['t0'] = t0
+            if t1 is not None:
+                if t0 is not None: q += ' AND '
+                q += 'e.mjd <= %(t1)s'
+                subdict['t1'] = t1
+        q += ( '   GROUP BY e.id ' # ,e.filepath,e.mjd,e.target,e.filter,e.filter_array,e.exp_time '
+               '   ORDER BY e.mjd, e.filter, e.filter_array ' )
+
```
guynir42, line 120: ...so this is what it looks like to work without an ORM...?

webap/static/seechange.js
```diff
@@ -0,0 +1,587 @@
...
+
+    // TODO : make this configurable?  Or at least remember how to
+    //   detect in javascript what the URL you're running from is.  (In
+    //   case the webap is not running as the root ap of the webserver.)
+    this.connector = new rkWebUtil.Connector( "/" );
+}
+
+seechange.Context.prototype.render_page = function()
+{
+    var self = this;
```
guynir42, line 29: I don't think you used `self` anywhere...

webap/static/seechange.js
```diff
@@ -0,0 +1,587 @@
...
+        if ( countdown == 0 ) {
+            countdown = 3;
+            fade = 1 - fade;
+        }
+    }
+}
+
+seechange.ExposureList.prototype.show_exposure = function( id, name, mjd, filter, target, exp_time )
+{
+    let self = this;
```
guynir42, line 198: There are a few of these scattered and I'm not sure if they are used.

tests/fixtures/cache_utils.py
guynir42: I think the functionality is identical to the way it worked on `Base`, but the documentation is a little misleading. 

I would consider putting this back into base, or at least into a util file, not under fixtures, as it could be useful outside of tests. And make the case of the non FileOnDiskMixin a little clearer in the docs.

guynir42 (commented):
I have some issue with moving the caching functions into the fixtures folder. 
Other than that I have minor comments.

.gitignore
rknop: I will change this.  webap/static/rkwebutil.js is made in the Makefile (it's just copied out of a submodule), so it should stay in .gitignore.  The other two are vestigal from before I edited the makefile and how these things were configured, so should generally not exist anyway.

tests/docker-compose.yaml
rknop: I'd like to leave them there for now.  If I do in fact write tests for the webap, we'll need that service.

webap/seechange_webap.py
rknop: I wouldn't even want to think about figuring out the sqlalchemy idiom that would correspond to this query.  It would probably take a me a week to figure out in the first place, and another 3 weeks to debug.

webap/static/seechange.js
rknop: This is a kneejerk thing I'll often do.  I find the javascript scoping rules for "this" nonintuitive (and based on web searches, lots of other people do to).  So often I'm putting in callbacks and stuff that need access to the "this" of the function where I'm building the callback (which will not be the "this" in the fucntion that gets called back) that I often stick this in when I start a function.

(Of course, just as often, I forget to put this in, and get an "undefined self" when I try to run it later.)

tests/fixtures/cache_utils.py
rknop: I'm not sure what the use case for that is.  If we *do* need to temporarily dump things and bring them back, we should do that in temp_path.  I'm not sure outside of test download efficiency we'd need to dump something and then bring it back up in a later run, which is what the cache is for.

I moved these becasue they only place they were used was in the tests, and setting the LIMIT_CACHE_USAGE environment variable was specifically for the tests -- meaning it made more sense that all of the caching be specifically for the tests.

webap/seechange_webap.py
guynir42: I think it would be something like: all the joins are done automatically using relationships, and all the sums you will just do on the python side. Less efficient but probably easier to write.

tests/fixtures/cache_utils.py
guynir42: There is no use case outside of tests because we are only doing tests. 
I can imagine us making mistakes with migrations and things like that, or wanting to fix some bad relationships between objects where one needs to be deleted but we don't want to remove the related objects. There could be lots of hacks like that. Cache is useful in tests but could be useful in general. Why not put it in util?

webap/seechange_webap.py
rknop: ...well, at least we can agree that it would be less efficient.

webap/static/seechange.js
rknop: Looking it, I *do* use self here -- it's in the callbacks to the document elements, e.g.:

`        let button = rkWebUtil.button( p, "Show Exposures", function() { self.show_exposures(); } );`

guynir42 (approved):
Looks good to me!

Issue #272, opened 2024-05-10
Don't save cutouts whose measurements don't pass


rknop:
Right now, we're saving all cutouts to the database, but only saving measurements that pass various analytic cuts.

We should only save cutouts when the measurements pass.

We might want to unify the measurements and cutouts tables, because there's a 1:1 relationship between the two.  Perhaps a CutoutsFile table (that is a FileOnDiskMixin) that we've talked about, which may or may not have cutouts that aren't in the measurements table.

guynir42:
I can see the advantages in making this change, especially in terms of simplifying the pipeline. 
I do see some advantages to the way we do things now: 
We may want to try lots of different measuring algorithms and if we decide to keep them, then we will need to have copies of the cutouts file for each version. If we keep them separate then we can have one file and lots of database rows for measurements with different provenances. 
Also, keeping all cutouts on file while deleting the database entries for those that didn't pass the quality cuts is not so simple, we'd have to make sure all the database information is saved in the file in case we want to recreate them. 
Also, the cutouts table doesn't have a lot of columns, so I'm not sure it will be such a big deal to leave all cutouts on DB. 

That said, we may just split it a bit differently: 
have one provenance step for the Cutouts (or CutoutsFile) that saves all the cutouts and takes up one row in the DB as a FileOnDiskMixin, and the other step for Measurements (or call it Cutouts?) and that one saves only the ones that pass, but include what was saved in individual cutouts before, as part of the measurements row (e.g., the coordinates in the image, and lazy loaded small images for ref/new/sub). This would not be very different from what we have now, but maybe makes more sense with the CutoutsFile.

rknop:
I like the latter solution.  A CutoutsFile table in the database that points to the HDF5 file, and the HDF5 file has all the cutouts (with indexes the same as indexes in the source_list), plus all the meta data (at least x and y; not sure if it needs much else).

Then, a Measurements table with rows only for things that pass the preliminary cuts.  Includes a pointer to the relevant CutoutsFile and to the index needed to find the actual cutout (just like Cutouts does right now).  And, yeah, include all the information that's currently in the Cutouts table.

rknop:
[...]

Issue #217, opened 2024-03-14
CutoutsFile


guynir42:
Right now each Cutouts object is also a FileOnDiskMixin which is a waste because a lot of cutouts (at least 10 if not 100 or 1000) are going to share the same file. That makes the Cutouts table wider than it needs to be. 

Instead, link all the cutouts from the same image to one CutoutsFile (or ListFile??) where we can keep track of things like filepath and MD5sum. This may also make it easier to load/save to file or copy to/from cache.

guynir42:
Related to #272

PR #302, opened 2024-06-04
CutoutsFile and Measurements

whohensee:
Change the structure of the cutouts and measurements such that we have a CutoutsFile that saves all the cutouts and takes up a single row in DB as FileOnDiskMixin, and Measurements which point to the CutoutsFile for relevant information.

See #217 and #272


whohensee:
I made a slight change from what Guy and I discussed about the `co_dict` attribute of cutouts, but I tried to implement it in a way that seemed most intuitive to use. The main `co_dict` (which contains individual subdictionaries for cutout data under keys such as "source_index_0") is actually a class that inherits from dict, and when it is passed a key that it doesn't already have it will search on disk for that key. The end result is that the user of Cutouts can just use the `co_dict` attribute without worrying about whether or not the specific cutouts data is loaded, and a user of Measurements can just use the attributes (such as `m.sub_data`), and things will load if they exist on disk.

If I forgot a useful case, feel free to let me know.

tests/pipeline/test_measuring.py
```diff
@@ -16,85 +16,80 @@ def test_measuring(measurer, decam_cutouts, decam_default_calibrators):
...
 
-    assert len(ds.all_measurements) == len(ds.cutouts)
+    assert len(ds.all_measurements) == len(ds.cutouts.co_dict)
 
     # verify all scores have been assigned
     for score in measurer.pars.analytical_cuts:
         assert score in ds.measurements[0].disqualifier_scores
 
-    m = ds.all_measurements[0]  # delta function
+    m = [m for m in ds.all_measurements if m.index_in_sources == 0][0]  # delta function
```
guynir42, line 92: I'm surprised that the order of `all_measurements` is not the same as the `index_in_sources`. If you'd do `all_measurements[i]` wouldn't you always get `index_in_sources==i`?

tests/pipeline/test_pipeline.py
```diff
@@ -425,14 +425,18 @@ def test_get_upstreams_and_downstreams(decam_exposure, decam_reference, decam_de
                 ds.zp.id,
             ])
             assert [upstream.id for upstream in ds.detections.get_upstreams(session)] == [ds.sub_image.id]
-            for cutout in ds.cutouts:
-                assert [upstream.id for upstream in cutout.get_upstreams(session)] == [ds.detections.id]
+            # for cutout in ds.cutouts:
+                # assert [upstream.id for upstream in cutout.get_upstreams(session)] == [ds.detections.id]
```
guynir42, line 429:
```suggestion
```

tests/pipeline/test_pipeline.py
```diff
@@ -425,14 +425,18 @@ def test_get_upstreams_and_downstreams(decam_exposure, decam_reference, decam_de
...
-            cutout_ids = np.unique([cutout.id for cutout in ds.cutouts])
+            # cutout_ids = np.unique([cutout.id for cutout in ds.cutouts])
+            m_upstream_ids = set([])
             for measurement in ds.measurements:
-                m_upstream_ids =  np.array([upstream.id for upstream in measurement.get_upstreams(session)])
-                assert np.all(np.isin(m_upstream_ids, cutout_ids))
+                # m_upstream_ids = np.array([upstream.id for upstream in measurement.get_upstreams(session)])
+                m_upstream_ids = m_upstream_ids.union(set([upstream.id for upstream in measurement.get_upstreams(session)]))
+                assert m_upstream_ids == set([ds.cutouts.id])
+                # assert np.all(np.isin(m_upstream_ids, cutout_ids))
```
guynir42, line 439: Remove commented lines if you don't need them anymore.

guynir42: and in the next block, too.

guynir42 (approved):
This is very good work. It indeed ended up touching a lot of code but it looks in very good shape. 
I have a bunch of small comments and corrections, but nothing major. 

Make sure your migration hash matches the latest main, and once you fix all the comments you can squash and merge.

guynir42:
Please also try to run the `models/test_image.py::test_image_products_are_deleted` test and see if it is faster now that we've moved from list to single object. If so, mark it not skipped.

tests/pipeline/test_measuring.py
whohensee: I did not get the same order, the test ended up failing a few indices in (at the gaussian), until I made this modification. I didn't look too deeply into why the order was not the same however. I will look a little bit more before I merge since it seems curious.

PR #305, opened 2024-06-09
Preprocessing steps

guynir42:
Refactor the way we request certain preprocessing steps to be done, e.g., applying flat or bias or fringe to certain images. 
- allows instruments to define a list of what steps they can do, and what steps they already done
- user requests a set of steps to be done, either by pipeline or by the thing that provided the exposures. 
- allow skipping certain steps for specific filters.


rknop (approved):
This looks good.

It occurs to me that the code assumes that DECam images always come in raw, and that's how we're usually doing it.  But you *can* pull in reduced DECam images.  When you do that, how do you specify that the steps_done field should be different from what's in the DECam class?

guynir42:
I think the original idea would be to use a separate Instrument class for processed exposures.

PR #314, opened 2024-06-20
143 limiting magnitude

dryczanowski:
Add estimate of 5 sigma limiting magnitude to SourceList


models/source_list.py
```diff
@@ -506,6 +506,34 @@ def calc_aper_cor( self, aper_num=0, inf_aper_num=None, min_stars=20 ):
 
         return -2.5 * np.log10( meanrat )
 
+    def estimate_lim_mag(self, aperture=1, givePlotParams=False):
+        # if aperture = -1: using psf mags, aperture defaults to 1
+        # image must also have zero point
```
guynir42, line 511: Probably better if you replace this with a full docstring.

models/source_list.py
```diff
@@ -506,6 +506,34 @@ def calc_aper_cor( self, aper_num=0, inf_aper_num=None, min_stars=20 ):
...
+            limMagEst = m * np.log(5) + c #limiting magnitude estimate at SNR = 5
+
+            if givePlotParams:
+                return limMagEst, snrMasked, magsMasked, m, c
+            else:
+                return limMagEst
+                
+        else:
+            #Will not provide limiting magnitude estimate if using PSF photometry 
+            #or if no zero point
```
guynir42, line 532: Remind me why we can't have limiting magnitude based on PSF photometry?

pipeline/photo_cal.py
```diff
@@ -303,12 +303,14 @@ def run(self, *args, **kwargs):
 
                 ds.image.zero_point_estimate = ds.zp.zp  # TODO: should we only write if the property is None?
                 # TODO: I'm putting a stupid placeholder instead of actual limiting magnitude, please fix this!
-                ds.image.lim_mag_estimate = ds.zp.zp - 2.5 * np.log10(5.0 * ds.image.bkg_rms_estimate)
+                # ds.image.lim_mag_estimate = ds.zp.zp - 2.5 * np.log10(5.0 * ds.image.bkg_rms_estimate)
```
guynir42, line 306: You can remove these two commented lines.

tests/models/test_image_querying.py
```diff
@@ -538,8 +538,8 @@ def test_image_query(ptf_ref, decam_reference, decam_datastore, decam_default_ca
         stmt = Image.query_images(max_exp_time=60, order_by='quality', seeing_quality_factor=factor)
         results3 = session.scalars(stmt.limit(2)).all()
 
-        # quality will be a higher, but also a different image will now have the second-best quality
-        assert results3 != results1
+        # images stay the same but quality will be higher
+        assert results3 == results1
```
guynir42, line 542: The only reason I wanted this second test is to show that we can find a `seeing_quality_factor` that shuffles the order of the images. Maybe if you increase it above `3.0` it would do that?

tests/models/test_image_querying.py
```diff
@@ -581,8 +581,10 @@ def test_image_query(ptf_ref, decam_reference, decam_datastore, decam_default_ca
         assert len(results4) == 2
         assert results4[0].mjd == results4[1].mjd  # same time, as one is a coadd of the other images
         assert results4[0].instrument == 'PTF'
-        assert results4[0].type == 'ComSci'  # the first one out is the high quality coadd
-        assert results4[1].type == 'Sci'  # the second one is the regular image
+        # assert results4[0].type == 'ComSci'  # the first one out is the high quality coadd
+        # assert results4[1].type == 'Sci'  # the second one is the regular image
```
guynir42, line 585: This seems fine. This highlights the fact that if you are filtering by "quality" you will definitely get the coadds firsts, as they are almost always going to be much deeper that the regular images.

guynir42: and also you can remove the commented lines.

tests/models/test_measurements.py
```diff
@@ -83,7 +84,40 @@ def test_measurements_attributes(measurer, ptf_datastore, test_config):
     m.flux_apertures[m.best_aperture] = original_flux
     new_im.zp.dzp = original_zp_err
 
-    # TODO: add test for limiting magnitude (issue #143)
+    #test limiting magnitude estimation
+    srcList = ptf_datastore.sources
+    #if interactive testing enabled, make and save a Magnitude vs SNR plot (for debugging)
```
guynir42, line 89: This is good!

tests/models/test_measurements.py
```diff
@@ -83,7 +84,40 @@ def test_measurements_attributes(measurer, ptf_datastore, test_config):
...
+            plt.xlabel('log SNR')
+            plt.ylabel('magnitude')
+            plt.title('Limiting magntiude = {:.2f} mag'.format(limMagEst))
+            ymin,ymax = plt.gca().get_ylim()
+            plt.vlines(x=np.log(5),ymin=ymin,ymax=ymax)
+            plt.hlines(y=limMagEst,xmin=np.log(3),xmax=np.log(20))
+            plt.xlim(np.log(3),np.log(20))
+            plt.ylim(ymin,ymax)
+            plt.savefig('plots/snr_mag_plot.png')
+            plt.show()
```
guynir42, line 116: I would argue that what would be really useful is to add the plotting option to the `estimate_lim_mag` function. That means you can always get feedback on that fit, regardless of tests. 

I would also use the fixture `blocking_plots` in the `show()` function that follows after the call to `estimate_lim_mag`, as in `plt.show(block=blocking_plots)`. That would stop the test until you close the window. 

I would also say that regardless of blocking or not (i.e., regardless of interactive usage) you should always want to save this plot as an artefact of running this test. It doesn't hurt.

guynir42 (commented):
I have a few ideas on improvements but overall it looks good.

models/source_list.py
dryczanowski: I think this is just what we decided to do, based on the fact that PSF photometry in the pipeline is currently unreliable.

guy-starkware: I think we should still allow limiting magnitude based on it, if someone wants it, just not have it as a default.

PR #328, opened 2024-07-05
add HDF5 compression for background objects

guynir42:
This should reduce the file sizes of the background images by about x2.


guynir42:
Writing compressed data with chunck size of `128x128` with minimal gzip compression factor yields a write time (for two full-sized arrays, one for counts and one for variance) that takes up 45MB instead of 70MB with a write time of 2s instead of 0.5s.

rknop (approved):

PR #329, opened 2024-07-05
Subtraction improvements

guynir42:
This PR attempts to make some improvements to the subtraction algorithm:
- [ ] Make sure images going into subtraction have their backgrounds removed first. 
- [ ] Try to chop up the images into smaller sections before subtraction. 

Maybe I'll think of more things as I go.


rknop:
Now that this PR is more than a year old, I'm going to close it.

PR #336, opened 2024-07-24
RA spanning 0°

rknop:


models/base.py
```diff
@@ -1622,24 +1650,88 @@ def find_containing( cls, siobj, session=None ):
 
         """
 
-        # Overabundance of caution to avoid SQL injection
+        # Overabundance of caution to avoid Bobby Tables.
+        # (Because pyhon is not strongly typed, siobj.ra and
```
whohensee, line 1654:
```suggestion
        # (Because python is not strongly typed, siobj.ra and
```

models/base.py
```diff
@@ -1649,6 +1741,115 @@ def find_containing( cls, siobj, session=None ):
...
+        Parameters
+        ----------
+          fcobj : FourCorners
+
+          session : Session
+             required here; otherwise, the temp table wouldn't be useful
+
+          prov_id : str, default None
+             The id of the provenance of objects to look for; defaults to
+             not filtering on provenance (which is almost enver what you want).
```
whohensee, line 1766:
```suggestion
             not filtering on provenance (which is almost never what you want).
```

models/base.py
```diff
@@ -1649,6 +1741,115 @@ def find_containing( cls, siobj, session=None ):
...
+        going to kill us here, so this may get wonky with big Δra/Δdec
+        or right near the poles.)  This box is defined by the least/greatest
+        RA/dec of all four corners.  (Below: the actual image is tilted
+        rectangle (modulo your font aspect ratio), the bounding box is the
+        one square to the screen.)
+
+                  __
+                 │╱╲│
+                 │╲╱│
+                  ‾‾
```
whohensee, line 1827: wow we get artwork in our docstrings!

tests/models/test_image_querying.py
```diff
@@ -121,121 +125,224 @@ def test_image_cone_search( provenance_base ):
...
             soughtids = set( [ s.id for s in sought ] )
-            assert len( { image1.id, image2.id, image3.id, imagepoint.id, imagefar.id } & soughtids ) == 0
+            assert { image1.id, image4.id, image5.id }.issubset( soughtids )
+            assert len( { image2.id, image3.id } & soughtids ) == 0
 
         finally:
-            session.rollback()
+            # When a test fails, this doesn't seem to actually
+            #  clean up the database.  Is this pytest subverting
+            #  the finally block?  Or is it some sqlqlchemy mysteriousness?
```
whohensee, line 340: could perhaps just try checking if it is in the session rather than persistent? Other than that, no clue...

whohensee (approved):
Not many comments on this one, I think it looks good as long as the proper values have the tests passing.

PR #355, opened 2024-08-29
Conductor password file & through_step

rknop:
A couple of small modifications:

* Added the ability to read the conductor password from a file so it doesn't have to sit in the clear in a config file, but can go to a secrets directory
* Added the ability to `top_level.py` and `pipeline_exposure_launcher.py` to only process up through a specified step.  (I added this because I wanted to abuse `pipeline_exposure_launcher.py` as a bulk downloader, but it's a general good thing to have, at least for testing.)
* Couple of small fixes to `invent_filepath` in `image` and `exposure`, including in documentation.

### DIFF
--- a/default_config.yaml
+++ b/default_config.yaml
@@ -23,7 +23,7 @@ storage:
     # should Image object save the weights/flags/etc in a single file with the image data?
     single_file: false
     # The convention for building filenames for images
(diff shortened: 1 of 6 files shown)


whohensee (approved):
No major comments, if all tests are passing it looks good to me. I will perhaps need to familiarize myself more with how the conductor operates in general, since I am not too familiar with anything higher-level than the pipeline itself.

PR #356, opened 2024-09-05
Download and create images from NOIRLab-reduced (not just raw) DECam exposures

rknop:


models/exposure.py
```diff
@@ -300,6 +300,17 @@ def filter_short(self):
             'Bad exposures are each bad in their own way (i.e., have different bits set). '
     )
 
+    preproc_bitflag = sa.Column(
+        sa.SMALLINT,
+        nullable=False,
+        server_default=sa.sql.elements.TextClause( '0' ),
```
whohensee, line 306: Why is this better than simply `server_default=0`?

models/exposure.py
```diff
@@ -612,33 +632,62 @@ def invent_filepath( self ):
...
+            Path or str pointing to where the file currently exists on
+            disk.
 
-        Keyword parmeters are passed on to FileOnDiskMixin.save().
+          - No positional parameters.  In this case, it will assume that
+            file (or files, if filepath_etensions is not None) are
+            already in the right place on disk, and it will use
+            self.get_fullpath(nofile=True) to find them.
+
+        Keyword parmeters are passed on to FileOnDiskMixin.save().  Do *not*
```
whohensee, line 661:
```suggestion
        Keyword parameters are passed on to FileOnDiskMixin.save().  Do *not*
```

tests/models/test_exposure.py
```diff
@@ -229,5 +229,15 @@ def test_exposure_comes_loaded_with_instrument_from_db(sim_exposure1):
...
 
 
+def test_save_preprocessed_exposure( decam_reduced_origin_exposure_loaded_in_db ):
+    exp = decam_reduced_origin_exposure_loaded_in_db
+    assert len( exp.get_fullpath(nofile=True) ) == 3
+    for extpath in exp.get_fullpath( nofile=True ):
+        assert os.path.isfile( extpath )
+    # Do more?  Really just testing that the fixture did what it was supposed to
+    # May be able to omit this test and just rely on
+    # test_image_from_reduced_exposure.
```
whohensee, line 239: I think its an okay check just in case somebody breaks something in the downloading function down the line, though I suppose that would also probably show up in the other tests too.

whohensee (approved):
Assuming tests pass, this looks good to me.

models/exposure.py
rknop: ....I don't remember exactly why I fell back on that.  But, when I was doing the database refactor, I figured out that just using a 0 did something strange database side that was messy there-- it had casts, or some such, that were gratuitous.  (Surprise: sqlalchemy does things that end up being clumsier database side!)  All of this is just a way of saying, "no, really, I just want to tell the database that the default is 0".)

PR #363, opened 2024-10-21
Rb

rknop:


extern/RBbot_inference
whohensee, line 1: I cannot see any details regarding this, so I am taking it as true that everything is good here.

pipeline/scoring.py
```diff
@@ -36,12 +48,69 @@ def get_process_name(self):
...
+             != (41,41) ):
+            raise ValueError( f"RBbot currently requires cutouts to be 41×41" )
+        data = np.empty( ( len(ds.measurements), 3, 41, 41 ) )
+        for i, meas in enumerate( ds.measurements ):
+            cdex = f'source_index_{meas.index_in_sources}'
+            data[ i, 0, :, : ] = cutouts.co_dict[cdex]['new_data']
+            data[ i, 1, :, : ] = cutouts.co_dict[cdex]['ref_data']
+            data[ i, 2, :, : ] = cutouts.co_dict[cdex]['sub_data']
+
+        # TODO : zero out masked pixels?
```
whohensee, line 84: TODO for this PR or future?

I'm not sure how much impact this would have on the scores if done vs not.

requirements.txt
```diff
@@ -34,8 +34,13 @@ SQLAlchemy==2.0.23
 SQLAlchemy-Utils==0.41.1
 tables==3.9.1
 urllib3==2.1.0
+uv==0.4.23
 wget==3.2
 
+# NOTE : pytorch is also required, but I haven't figured out how to get
+# a CPU-only version of pytorch from PyPi.  (The GPU version adds 5GB to
+# the Dockerfile...!)   So, it's manually installed in the Dockerfile.
+
```
whohensee, line 43: This is 5GB of storage on the machine running the pipeline as opposed to 5GB ram usage per running instance of the pipeline, right?

share/RBbot_models/RBbot-feasible-disco-117-config.json
whohensee, line 1: all of these link to an external git module so I can't look at them, but I will assume they are configured correctly unless you want them reviewed too.

tests/fixtures/decam.py
whohensee, line 1: Very nice improvements/re-factor in this file!

whohensee (approved):
Looks good, nice to have a real RB algorithm functioning. I wonder if it does better than my 'random'.... (just kidding). 

I didn't look at the stuff in the external RBbot module that was added. If you need me to do that, let me know.

extern/RBbot_inference
rknop: The file is really named `requirements.txt`, and I only just now noticed the docs were wrong.  (I noticed because I went to look at them given that I had to change the docker image, and I didn't remember what to do for github with that.)

pipeline/scoring.py
rknop: I think I need to do this.  Checking with Nabeel (who build the actual RB model we're using).

requirements.txt
rknop: ...you are pushing the limits of my docker knowledge.

It's 5GB of storage on the machine plus 5GB of RAM on the running machine, but I don't think it's 5GB for running container, because I believe multiple containers can share the memory used to store the image.

In any event, it's ridiculous, since the docker image is 3GB without the GPU stuff, and we aren't using the GPU at all.

share/RBbot_models/RBbot-feasible-disco-117-config.json
rknop: Naah.  This is in Nabeel's code.  They'd be opaque in any event.

PR #370, opened 2024-11-06
Alard lupton

rknop:


improc/zogy.py
```diff
@@ -336,35 +337,66 @@ def zogy_add_weights_flags(ref_weight, new_weight, ref_flags, new_flags, ref_psf
...
+    # For two fluxes to have the same magnitude:
+    #   m = -2.5*log(f1) + zp1 = -2.5*log(f2) + zp2
+    #   -2.5*log(f1/f2) = zp2 - zp1
+    #   f1/f2 = 10**(-0.4*(zp2-zp1))
+    #   σ1²/σ2² = 10**(-0.8*(zp2-zp1))
+    w1 *= 10 ** ( -0.8 * ( ref_zp - sub_zp ) )
+    w2 *= 10 ** ( -0.8 * ( new_zp - sub_zp ) )
+
+    # combine the weights
+    mask = (ref_weight == 0) | (new_weight == 0) | ( new_flags !=0 ) | ( ref_flags != 0 )
```
whohensee, line 391: If I understand right, the `|` statements are applying bitwise OR, applied individually across each element of the array? Like the output of `mask` is an array of same size with `1` (or `True`?) for any element where that location was equal to 0 in one of these 4 arrays?

Thats a pretty neat and compact way of doing this with numpy arrays.

models/instrument.py
```diff
@@ -1022,14 +1022,22 @@ def get_gain_at_pixel( self, image, x, y, section_id=None ):
...
-        else:
-            return 1.
+        raise NotImplementedError( f"{self.__class__.__name___} needs to implement get_gain_at_pixel" )
+
+        # Leaving the previous implementation here commented out
+        #   in case for some reason we need to fall back to it.
+        # For now, all of DemoInstrument, PTF, and DECam have
+        #   their own implementations.  We either need to solve
+        #   Issue #369, or insist that every Instrument subclass
+        #   implement get_gain_at_pixel (which is the current behavior).
```
whohensee, line 1032: I'll mark down to read through 369 - at first glance it doesn't seem too burdensome to have each Instrument implement its own.

pipeline/coaddition.py
```diff
@@ -232,16 +232,37 @@ def _zogy_core(self, datacube, psfcube, sigmas, flux_zps):
...
+            normalized.)  (...in fact, I think it doesn't even do quite
+            the right coadd (up to a normalization factor) for something
+            other than an image, unless all of the flux_zps are the same
+            as each other.  So, don't use this for anything other than
+            images.)  Must already be aligned.
+
         psfcube: ndarray
-            The PSF cube to use for coaddition.
+            The PSF cube to use for coaddition.  3D numpy array, each layer must
+            have a normalized (i.e. sum=1) PSF image at the scame pixel scale
```
whohensee, line 249:
```suggestion
            have a normalized (i.e. sum=1) PSF image at the same pixel scale
```

pipeline/coaddition.py
```diff
@@ -416,6 +439,11 @@ def _coadd_zogy(
         if bkg_means is None or bkg_sigmas is None:
             raise ValueError('Background must be given if images are not Image objects. ')
 
+        # NOTE -- this will use a LOT of memory.
+        # All of the numpy arrays will now be stored twice -- once in
+        #   the list, and once in the datacube created.  Suggest
+        #   refactoring so that the for loop above just builds the
+        #   datacubes directly instead of going through the lists.
```
whohensee, line 446: Have we checked lately how much memory usage we are up to? Is that still a concern limiting our throughput?

pipeline/subtraction.py
```diff
@@ -1,12 +1,22 @@
...
+import subprocess
 
+import numpy as np
+import pandas
 import sqlalchemy as sa
 
+from astropy.io import fits
+import astropy.coordinates
+import astropy.units as units
+from astropy.coordinates import SkyCoord
```
whohensee, line 14:
```suggestion
```

whohensee: Appears to be unused

pipeline/subtraction.py
```diff
@@ -225,49 +236,335 @@ def _subtract_zogy(self,
...
-            The Image containing the reference data, including the data array, weight, and flags
-            Image must also have the PSF and ZeroPoint objects loaded.
+            The Image containing the reference data, including the data
+            array, weight, and flags.  ref_image and new_image must be
+            aligned to each other.
+
+        ref_sources: SourceList
+            Source List for ref_image.  Must have the is_star parameter
+            (which, as of this writing, means that the source list must
+            be a sextrfits source list.)
```
whohensee, line 299: As a fleeting thought, do you think it would be better to say something like "as of Nov 2024" to add context the next time we come around to this? This might be be too minor to worry about.

pipeline/subtraction.py
```diff
@@ -225,49 +236,335 @@ def _subtract_zogy(self,
...
+            hotgaussfactors = [ 0.5, 1.0, 1.5 ]
+            if new_psf.fwhm_pixels < ref_psf.fwhm_pixels:
+                SCLogger.warning( f"New has better setting {new_psf.fwhm_pixels:.2f} "
+                                  f"than ref {ref_psf.fwhm_pixels:.2f}! "
+                                  f"Going to use higher orders so there can be sharpneing scariness (brrr....)" )
+                hotgaussorders = [ 6, 4, 2 ]
+                hotgaussfactors = [ 0.5, 1., 2. ]
+                nominalfwhm = trtlt * pixscale
+            else:
+                noimnalfwhm = np.sqrt( new_psf.fwhm_pixels**2 - ref_psf.fwhm_pixels**2 ) * pixscale
```
whohensee, line 371:
```suggestion
                nominalfwhm = np.sqrt( new_psf.fwhm_pixels**2 - ref_psf.fwhm_pixels**2 ) * pixscale
```

pipeline/subtraction.py
```diff
@@ -225,49 +236,335 @@ def _subtract_zogy(self,
...
+                SCLogger.error( f"_subtract_hotpants: {len(newstars)} stars for PSF matching isn't enough" )
+                raise RuntimeError( "Not enough stars for _subtract_hotpants PSF matching" )
+            if len(newstars) < nstars_ideal // 4:
+                SCLogger.warning( f"_subtract_hotpants: {len(newstars)} stars for PSF matching is low, "
+                                  f"reducing kernel orders" )
+                hotgaussorders = [ 1, 0 ]
+                hotgaussfactors = [ 0.6, 1.2 ]
+            if len(newstars) > nstars_ideal * 2:
+                # embarssment of riches, crop the list to make hotpants run faster
+                sortdex = numpy.argsort( newstars.flux )
```
whohensee, line 490:
```suggestion
                sortdex = np.argsort( newstars.flux )
```

whohensee: Unless you want to change the import statement instead

pipeline/subtraction.py
```diff
@@ -225,49 +236,335 @@ def _subtract_zogy(self,
...
+
+            SCLogger.debug( f"_subtract_hotpants returning." )
+            return retval
+
+        # except Exception as ex:
+        #     # This is just here for debugging so I can see
+        #     #   what the exception was before it jumps to
+        #     #   the finally block.
+        #     import pdb; pdb.set_trace()
+        #     raise
```
whohensee, line 561: double checking if you wanted this comment block to stay around for future debugging or not

whohensee (approved):
Just a few misnamed variables that need fixing, structurally it all looks good to me.

pipeline/coaddition.py
rknop: I'm still worried.  I'd like it if the subtraction pipeline only used 2GB, but that's probably hopeless at this point.  But I'd still like to get it less than 4GB or 5GB.  I haven't looked at it closely yet, but I've idly looked at it and we still go above 5GB, at least in the tests.  (The tests may be more memory-intensive than the actual pipeline, as we keep some extra data around for testing purposes.)

For the coadd pipeline, the memory usage always will be more, because we might be reading in lots of images all at once.  (Turns out swarp can handle not doing that, but also that can make swarp slower as it has more disk i/o.)  So, we won't be able to run as many reference-building pipeline instances on one node at once.  But, that's less of a big deal, because we expect many (many) subtraction runs for each reference built, so it's OK if the reference building is more resource intenstive.  But, still : if we're building a reference from 20 images, having double copies of arrays for all of those 20 images is probably a bit much!  I'm not sure which reference pipeline we're going to use for LS4, the zogy or swarp one.  For decat-wfd, we need to use swarp, because it handles combining together lots of misaligned images better.

At some point in the not-too-distant future, I hope to be really running this on the decat-wfd data, at which point I can look at the memory usage for reference building and subtraction, and start to worry about that for real.

pipeline/subtraction.py
rknop: Yeah, maybe.  I often put that in as a disclaimer because *all the time* things change somewhere else in the code, and we don't think to update the documentation in other places.  (At least one change in this PR was me fixing documentation for something that changed several PRs ago.)

pipeline/subtraction.py
rknop: I do want to, but also don't think we should.

(I much prefer using `numpy.` everywhere-- I have been bitten by namespace pollution before, and on the balance between "distinct and explicit" and "too much to type", I usually lean towards the former, *especially* when we're just talking 3 extra letters.  But the standard that everybody else uses is `np.`, and since this pipeline isn't my private playground, we should stick with what most people do.)

pipeline/subtraction.py
rknop: Yeah, let's leave it here.  Future code archeologists decades or centuries will gain insight into our thought processes, and we don't want to deprive them of that.

PR #372, opened 2024-11-08
Reduce docker image size

rknop:
Add --no-install-recommends to apt-get install commands in application dockerfile.

### DIFF
--- a/docker/application/Dockerfile
+++ b/docker/application/Dockerfile
@@ -37,18 +37,18 @@
 #     test_bindmount
 #
 # As of 2024-06-10, the sizes of images from the various stages are as follows:
-#   base                1.18GB
-#   build               2.82GB
-#   included_code_build 2.82GB
-#   conductor_build     2.82GB
-#   build_selenium      2.89GB
-#   included_code       2.06GB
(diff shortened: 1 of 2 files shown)


docker/application/Dockerfile
```diff
@@ -36,19 +36,19 @@
...
+#   included_code_build 3.49GB
+#   conductor_build     3.49GB
+#   build_selenium      3.48GB
+#   included_code       2.55GB
+#   conductor           2.55GB
+#   bindmount_code      2.54GB
+#   conductor_bindmount 2.54GB
+#   base_selenium       1.22GB
+#   test_bindmount      2.92GB
+#   test_included       2.93GB
```
rknop, line 51: Note that this comment doesn't make it look like the docker image size was reduced....!  The reason is I hadn't been keeping this comment up to date.  Between 06-10 and yesterday, things had grown quite a bit as new things were included (most significantly, the stuff needed for R/B).  Before this PR, the test_bindmount image (which is the one we mainly use) was 3.51GB.

whohensee (approved):
Great!

PR #377, opened 2024-12-03
Misc. scaling edits

rknop:
This one is a hodgepodge.  It started when I tried to start working on running the Decam DDF data, which ran into various problems having to do with references and running lots of things at the same time.

* To models/base.py, add a Psycopg2Connection context manager to get a database connection directly via Postgres that completely eschews SQLAlchemy.  (Session() still exists, as lots of the code still uses the SQLAlchemy connection, but this allows using SQL without having to worry about what baggage may be coming with the SA wrapper.)

* Add GaiaDR3DownloadLock to models/catalog_exerpt.py to fight another race condition that happened when I ran a bunch of things at once.  Use this in pipeline/catalog_tools.py

* Webap: install the SeeChange code base into the webap docker image.  (When I first wrote the webap, we didn't have "make install" yet, and I didn't want to bloat the dockerfile with a full git checkout; that's why I didn't originally include the SeeChange code in the webap.  Well, that, and the fact that we were more tied to SQLAlchemy and relations back then, and I didn't want to expand the nightmare.)

* Refactor import_decam_reference a bit to use a Pipeline object instead of lots of individual worker objects( Detector, Backgrounder, etc.)

* Move provenance tagging from pipeline/top_level.py to models/provenance.py::ProvenanceTag.  (Useful at least for webap.)

* Add -PIXEL_SCALE to sextractor call.  This helps the star/galaxy identification a lot in test DECam images... but we probably ought to do better, which might require yet another iterative call to sextractor.  See comment ca. line 655 in pipeline/detection.py

* Put backgrounding, astrocal, and photocal back in as steps in report.  (They're all done in the "extraction" stage, but can receive separate reports of completion.)  This had a bunch of fallout that eventually was mostly solved by removing underscores from astro_cal and photo_cal in a few places.

* Get reports into the datastore_factory test fixture, because we need it for new tests for the webap.  (Previously, datastore_factory didn't cache reports.)

* More webap functionality, lots of webap refactoring, more webap tests.  More webap tests probably still need to be written, and OMG are they painful to write.

* Update the rkwebutil submodule (included underneath webap, should probably move to extern), use rkAuthClient from rkwebutil/rkauth_client.py in place of custom code in (at least) ConductorConnectror

* Misc
   - conductor flask secret key handling
   - Dockerfile tweaks.  (No point for conductor_bindmount, webap built by applciation Dockerfile.)
   - Makefile.am updates to make them current with all stuff needed


tests/fixtures/webap.py
```diff
@@ -0,0 +1,74 @@
...
+    buttons = browser.find_elements( By.TAG_NAME, 'button' )
+    button = None
+    for possible_button in buttons:
+        if possible_button.get_attribute( "innerHTML" ) == "Log In":
+            button = possible_button
+            break
+    assert button is not None
+    button.click()
+
+    def check_logged_in( d ):
```
whohensee, line 57: Note sure if the variable `d` is used here, but I may be missing the reason is it still worth having passed to the function

tests/webap/test_webap.py
```diff
@@ -62,86 +221,116 @@ def test_webap( browser, webap_url, decam_datastore ):
...
-        # check stuff about the rows?
+        # Now the tab content div should have information about the sources
+        sourcesdiv = subcontentdiv.find_element( By.XPATH, "./div" )
+        assert sourcesdiv.get_attribute('id') == "exposurecutoutsdiv"
+        sourcestable = sourcesdiv.find_element( By.TAG_NAME, 'table' )
+        rows = sourcestable.find_elements( By.TAG_NAME, 'tr' )
+        assert len(rows) == 8
 
-        # There is probably more we should be testing here.  Definitely.
+        # OMG writing these tests is exhausting.  There is still lots more to do:
```
whohensee, line 328: yeah this was a lot...

webap/local_augments.yaml
whohensee, line 1: empty file, but I assume for configuration in the future?

webap/local_overrides.yaml
whohensee, line 1: also empty file, but I assume for configuration in the future?

whohensee (approved):
I didn't find any major issues, looks good! I didn't look as closely at the major webap files and some files in the hacks directories, but I did at least look them over lightly, and nothing jumped out.

improc/__init__.py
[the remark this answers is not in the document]
rknop: Yeah.  The very existence of this file turns it into a module.  (cf: https://docs.python.org/3/tutorial/modules.html )

pipeline/detection.py
```diff
@@ -654,6 +654,15 @@ def _run_sextractor_once(self, image, apers=[5, ], psffile=None, psfnorm=3.0, wc
...
 
+            # NOTE -- looking at the SExtractor documentation, it needs
+            # SEEING_FWHM to be ±20% right for CLASS_STAR to be reasonable for
+            # bright sources, ±5% for dim.  We have a hardcore chicken-and-egg
+            # problem here.  The default SEEING_FWHM is 1.2; try just going with that,
+            # and give it PIXEL_SCALE that's right.  We may need to move to
+            # doing this iteratively (i.e. go from two to three runs of SExtractor;
+            # first time around, do whatever, but look at the distribution of FWHMs
+            # in an attempt to figure out what the actual seeing FWHM is.)
+
```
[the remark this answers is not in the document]
rknop: Made Issue #381

requirements.txt
```diff
@@ -2,7 +2,7 @@
 alembic==1.12.1
 astro-datalab==2.20.1
 astropy==5.3.4
-astroquery==0.4.6
+# astroquery==0.4.6
```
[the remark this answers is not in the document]
rknop: Turns out that astroquery=0.4.3 is already on the old packages list....  I'll just delete this line.

tests/fixtures/webap.py
rknop: It's a callback thing in `WebDriverWait.until`.  Even though I don't use it, the function passed to `until` will be getting an argument when `until` calls it.

webap/local_augments.yaml
rknop: Both this and `local_overrides.yaml` are empty to make sure that the test config works right.  If you are futzing around in your own checkout and create these files, `git status` will show you that you did that, and that might help you figure out why the test no longer pass.

(So, we want to make sure both of these files stay empty in the repo.)

PR #379, opened 2024-12-05
Ruff

rknop:
Add the ruff linter.   Put it in the github workflows before the tests already run.

You can run the ruff linter in your own checkout by just running `ruff check` in the top directory.  If you don't have ruff installed on your system, just go into the shell of the test environment and run it there.


util/util.py
```diff
@@ -425,7 +426,7 @@ def get_inheritors(klass):
 
 
 def as_UUID( val, canbenone=True ):
-    """Convert a string or None to a uuid.UUID
+    r"""Convert a string or None to a uuid.UUID
```
whohensee, line 429: do we need this raw string tag?

util/util.py
```diff
@@ -456,7 +457,7 @@ def as_UUID( val, canbenone=True ):
 
 
 def as_datetime( string ):
-    """Convert a string to datetime.date with some error checking, allowing a null op.
+    r"""Convert a string to datetime.date with some error checking, allowing a null op.
```
whohensee, line 460: same question on if the r-string is needed?

whohensee (commented):
Looks good. I don't remember what my earlier comments were, but there was at least one error so keep an eye out for that.

I like the style that ruff is enforcing, so this should be good to have!

conductor/updater.py
```diff
@@ -4,23 +4,21 @@
...
 import logging
 import json
-import multiprocessing
 
 # Have to manually import any instrument modules
 #  we want to be able to find.  (Otherwise, they
 #  won't be found when models.instrument  is
 #  initialized.)
-import models.decam
+import models.decam   # noqa: F401
```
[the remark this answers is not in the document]
whohensee: I have since read the sections on adding ruff so of course this makes sense now

util/util.py
rknop: This one *is* needed because of the backslashes further down in the docstring.

whohensee (approved):

Issue #373, opened 2024-11-11
Write backgrounds as float32, think about compression


rknop:
Right now, background maps are written to HDF5 files as float64.  This is false precision, and a waste of disk space.  Convert to float32, or (better) to something like a bscaled int16.

(Any realistic variation in background across the image is not going to require more than 16 bits of precision if the mapping is scaled between the minimum and maximum pixel values.)

rknop:
We should also think about compression.  The *easiest* thing to do would be to move to storing hdf5 files as FITS images and using fpack....

rknop:
https://www.astro.louisville.edu/software/xmccd/archive/xmccd-4.1/xmccd-4.1e/libcfitsio/libcfitsio/fpackguide.pdf

fpack's default compression algorithm is called Rice.  But, it also does "quantization"; figure out if that's part of rice, or if it's something fpack does before running the compression algorithm.

rknop:
Looks like we're already using hdf5 gzip compression.

Think about quantization (manual bscaling?).

PR #387, opened 2024-12-13
Background quantization

rknop:
We were saving backgrounds as 32-bit floats, which have 24 bits of precision.  This was overkill; backgrounds, and the noise in backgrounds, don't require that much precision.

In the saving/loading routine, quantize the background to 16-bit or 8-bit integers before saving.  Often, an 8-bit integer is sufficient.  This leads to more than a 4x savings in disk space, however, because once quantized (at least to 8-bit, if not 16-bit), the gzip algorithm will end up compressing a lot better.

For convenience in quantization, I changed what's stored in and saved from background objects.  Used to store background and variance in background, now stores background and rms residual (which is just sqrt(variance)).

Issue #373.

### DIFF
--- a/improc/tools.py
+++ b/improc/tools.py
@@ -84,7 +84,63 @@ def sigma_clipping(values, nsigma=3.0, iterations=5, axis=None, median=False):
     return mean, rms
 
 
+def find_bscale( arr, rescut ):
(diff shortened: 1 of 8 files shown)


whohensee (approved):
Looks good! The quantization to smaller integers is clever, nice catch!

PR #390, opened 2024-12-17
Short filter in database

whohensee:
Store only the short filter names in database for exposures and ensure that instrument objects have the necessary functions and dictionaries to convert to full filter names when needed. 

This issue will be more complete once we implement a system for "allowed instruments", after which we can be more controlled about what can enter the database by requiring any exposure to have an allowed instrument, and ensuring that the filter value is legal for that instrument. (If I haven't made sure there is an issue for this and added it here then I should be reminded)

Focused on Exposures, although Image also has a database column for filter. I think this is generally acquired directly from Exposure, however, so the changes that ensure Exposures use short filter names should trickle down to Image as well, but its an option to try to be more explicit.


tests/models/test_image_querying.py
```diff
@@ -507,8 +507,8 @@ def test_find_images(ptf_reference_image_datastores, ptf_ref,
     assert all(im.instrument == 'PTF' for im in found6)
     assert set( f.id for f in found6 ) == set( f.id for f in found1 )
 
-    found7 = Image.find_images(filter='r DECam SDSS c0002 6415.0 1480.0')
-    assert all(im.filter == 'r DECam SDSS c0002 6415.0 1480.0' for im in found7)
+    found7 = Image.find_images(filter='r')
+    assert all(im.filter == 'r' for im in found7)
```
whohensee, line 511: This works fine for this test, for now, however since we are now working with short filter names, would it be better to make this query search for "filter r AND instrument DECam"?

models/exposure.py
```diff
@@ -250,21 +250,42 @@ def format(self, value):
 
     exp_time = sa.Column(sa.REAL, nullable=False, index=True, doc="Exposure time in seconds. ")
 
-    filter = sa.Column(sa.Text, nullable=True, index=True, doc="Name of the filter used to make this exposure. ")
+    _filter = sa.Column(sa.Text,
+                        nullable=True, #nullable because a constraint checks _filter or filter_array
+                        index=True,
+                        doc=("Name of the filter used to make this exposure. "
+                             "This is generally the short filter name - conversion "
```
rknop, line 257:
```suggestion
                             "This is the short filter name - conversion "
```

models/exposure.py
```diff
@@ -250,21 +250,42 @@ def format(self, value):
...
-        return self.instrument_object.get_short_filter_name(self.filter)
+        else:
+            return self._filter
+
+    @filter.setter
+    def filter( self, val ):
+        # Once instrument can only be a valid instrument, should add
+        # a check here which says "IF there is an instrument, only
+        # allow a valid filter for that instrument". Leaving the
+        # ability to set any filter at the moment for tests.
```
rknop, line 274: Make an issue to go along with this comment.

models/exposure.py
```diff
@@ -366,6 +387,10 @@ def __init__(self, current_file=None, invent_filepath=True, **kwargs):
         # We will run this exact code again later so that the keywords
         # can override what's detected from the header.
         self.set_attributes_from_dict( kwargs )
+        # NOTE: since above reads directly from header, it will often set
+        # a full filter name without an instrument, which must be detected
+        # and fixed below before making anything permanent (filepaths,
+        # provenances etc).
```
rknop, line 393: I wonder if it's worth putting in some explicit code right here to look at the _filter attribute.  If it's not None, and the isntrument isn't None, call the instrument's `get_short_filter_name` method.

rknop: We'll probably need the same code in Image.

models/instrument.py
```diff
@@ -1202,6 +1202,16 @@ def get_short_filter_name(cls, filter):
 
         return filter
 
+    @classmethod
+    def get_full_filter_name(cls, shortfilter):
```
rknop, line 1206: In Exposure there's a method `filter_full` that returns the long filter name.  Here something "full" returns the short filter name.  We should have consistent naming to avoid confusion.  Perhaps this function should be `get_canonical_filter_name` or something like that?  I would stick with using "full" to refer only to long filter names.

models/instrument.py
```diff
@@ -1569,7 +1579,7 @@ def preprocessing_calibrator_files( self, calibset, flattype, section, filter, m
                     # we had to do it before searching for the file so that we don't have a race
                     # condition for multiple processes all downloading the file at once.
                     calib = self._get_default_calibrator( mjd, section, calibtype=calibtype,
-                                                          filter=self.get_short_filter_name( filter ),
+                                                          filter=filter,
```
rknop, line 1582: Add in the docstring for this function that `filter` needs to be the short filter name, not the long filter name.

tests/models/test_decam.py
```diff
@@ -32,7 +32,7 @@ def test_decam_exposure(decam_exposure):
     assert e.dec == -43.0096
     assert e.exp_time == 86.0
     assert e.filepath == 'c4d_230702_080904_ori.fits.fz'
-    assert e.filter == 'r DECam SDSS c0002 6415.0 1480.0'
+    assert e.instrument_object.get_full_filter_name( e.filter ) == 'r DECam SDSS c0002 6415.0 1480.0'
```
rknop, line 35:
```suggestion
    assert e.filter == 'r'
    assert e.instrument_object.get_full_filter_name( e.filter ) == 'r DECam SDSS c0002 6415.0 1480.0'
```

tests/models/test_decam.py
```diff
@@ -362,7 +362,7 @@ def test_decam_download_and_commit_exposure(
                 match = re.search( r'^c4d_(?P<yymmdd>\d{6})_(?P<hhmmss>\d{6})_ori.fits', fname )
                 assert match is not None
                 # Todo : add the subdirectory to dbfname once that is implemented
-                dbfname = ( f'c4d_20{match.group("yymmdd")}_{match.group("hhmmss")}_{exposure.filter[0]}_'
+                dbfname = ( f'c4d_20{match.group("yymmdd")}_{match.group("hhmmss")}_{exposure.filter}_'
```
rknop, line 365: This right here is one of the reasons why I really like going to short filter names as the standard.  Probably, usually, in the old world, the short filter name would have been the first character of the long filter name... but could we coun't on that?  We had that baked in, but, now, don't have to worry about it.

tests/models/test_image_querying.py
rknop: For purposes of this test, I think this is fine.  The database should be clean and only have what we expect in it.

For real production use, yeah, you want to filter on both instrument and filter.

rknop (changes requested):
Few small suggested changes.  The most annoying one will be a method rename I've requested.

I'm going to pull down this PR and see if I can add a few lines to the migration script to do the filter translation on a live database.

models/exposure.py
rknop: (Just make an issue for this, don't write it in now.)

models/exposure.py
whohensee: #391

models/instrument.py
whohensee: We discussed this in slack - the name is good, the behavior in `instrument.py` is confusing because by default it spits back out the short filter until it is overridden by an instrument. I'll add a comment to make this less confusing.

tests/models/test_decam.py
whohensee: Yes, I noticed that here too! Much more flexible this way.

rknop (approved):

Issue #396, opened 2024-12-20
Think about how we do photometry.


rknop:
Right now we have our own python code to do photometry in `pipeline/measuring.py`, `models/measurements.py`, and `improc/photometry.py`.

There are some issues with it.  In `pipeline/measuring.py`, it (as of this writing, but I'm changing it) defaults to using a local background annulus for aperture photometry.  However, it also defaults to a fixed-pixel-size annulus, which is not good when the seeing gets big.  I was seeing it behave poorly in a case like that (e.g. an obvious isolated, though blobby from bad seeing, spot on the sub image ended up with the photometry code saying the centroid was 6 pixels away, which was visibly wrong.)

But, when it does psf photometry, it calls a routine in `models/measurements.py` that has no concept of the background annulus.  At the very least, we need to be self-consistent.

Since we're talking about subtraction images here, we don't really need to be doing background subtraction.

We should also think about removing all of our custom code and using a standard package like `photutils`.  I trust `photutils`, and they have probably through through edge cases better than we have.  In that case, it might make more sense to make the measurements on the sub image rather than the cutouts.  Yes, it's nice to make measurements on cutouts, because you can easily redo them.  But, it will probably be more efficient to pass a list of coordinates on the whole sub image to a photutils routine than to iterate over the cutouts and do the calculations on each cutout.

rknop:
When addressing this issue, search for comment "This is a little bit kludgy" in measuring.py.  That is one (of many) places where things will need to be edited.

Issue #389, opened 2024-12-16
report chicken-and-egg


rknop:

rknop:
Gah.  The ability to have different referencing provenances in the same refset breaks the provenance model.

We could just remove the referencing (or, equivalent, preprocessing/extraction upstreams of the referencing) provenance from the subtraction provenance, on the basis that `refset` is already an argument of subtraction, and so the refset is baked in.

But, that breaks the provenance model, because the refset is a user string, and the pointers between the refset and the actual provenances of the thing that made the provenances could be chaged at any time.  We're no longer tracking the provenance of what went into the subtraction.

We could relax the "only one subtraction provenance for a given provenacne tag" restriction, which woudl allow subtractions that are otherwise the same but that use different reference provenances... but that also breaks the provenance model, because we could then end up with two subtractions of the same field with different provenances but the same provenance tag, so we don't know which one is the "right" one to look at.

The only way to keep all of this consistent with the provenance model is demand that a refset have a *single* set of upstreams.  That is, remove the refset_provenance_association table, and make provenance_id a field of refset (call it referencing_provenance_id or some such).

I predict, however, that that cuases pain for people who actually want to use the pipeline, because they will end up with different referencing provenances that they want to all use as part of the same refset batch.  You *don't* want to do this when you're doing a steady-state survey, because you really want to use consistent references.  But in the first year of a survey when thigns are chaotic....  yikers.

rknop:
I think the thing to do is restrict refsets to a single referencing provenance, and just require the user to deal with the fallout.

The user being, of course, us :(

Issue #374, opened 2024-11-13
Put in refset/provenance association when manually loading decam pre-built references


rknop:

rknop:
...I think this is done, check that.

rknop:
It's not done.

Issue #371, opened 2024-11-07
Run DECAT-DDF data with hotpants & zogy subtractions, make sure we catch known supernovae


rknop:

rknop:
Progress with PR #401 ; have moved to using a test DECam image (from the previous PR) that has a nice SN from DECAT-DDF (DC21cyddn).  With PR #401, that one is found and is passing our analytic cuts.

Still have to do lots.

PR #397, opened 2025-01-06
Latest decam ddf run attempt changes

rknop:
All things that came up when I did my latest run of trying to run at (small) scale on NERSC.

Addresses issues #371, #374, #170, #277 (I think?), #340, #395,

Addresses issue #389 by making the data model more restrictive.  A refset can now only have a single reference provenance assocated with it.  (See discussion in Issue #389.)  This cascaded into simplifying references and upstreams and so forth, which was all too complicated and tangled.  This in turn cascaded into replacing image_upstreams_association with image_coadd_components.  We had the notion of general image upstreams, but in practice it was used in two cases: coadds and subtractions.  We *already* had special case handling for subtractions, and making the image_upstreams_association work for both was more complicated than just for coadds, and wasn't worth it given the already-existing subtraction special handling.  But, of course, that munged up all kinds of places images were used and tested.

Struggles with Issue #394.  Doesn't really solve it, but adds in some stuff to fix problems assocated with Issue #394.  I think these address Issue #277.

Struggles with, but does not resolve, Issue #396.

Splits pipeline_exposure_launcher.py into that plus exposure_processor.py.  The idea is (later) that the latter will be runnable from the command line to process a single manually-specified exposure.  (In contrast, pipeline_exposure_launcher sits there and asks the conductor what to do.)  Some thought may still be required to think about whether exposure_processor.py should tell the conductor things or not.  (Right now, only pipeline_exposure_launcher.py communicates with the conductor.)

Removes the exception catching that DataStore did.  I think the idea behind this was that we would be able to pass exceptions on to reports in a coordinated manner, but in practice it made finding the reasons why tests failed a whole lot harder, and wasn't worth it.

I also futzed with the analytical cuts as I looked at things that were getting passed and missed.  (Most of this may have come out of the photometry futzing.)  Ultimately, this lead me to post Issue #398.

I changed one of the test exposures and one of the test reference images used in the tests.   Painful.  But, I realized it made sense to have an image in there that had a nice SN detection in it (DC21cyddn -- see https://decat-webap.lbl.gov/decatview.py/cand/DC21cyddn?rbtype=2).


pipeline/top_level.py
```diff
@@ -285,31 +302,44 @@ def setup_datastore(self, *args, no_provtag=False, ok_no_ref_provs=False, **kwar
...
+            ds.report = None
 
         return ds, session
 
+
+    def _get_stepstodo( self ):
+        stepstodo = self.ALL_STEPS
+        if self.pars.through_step is not None:
+            if self.pars.through_step not in stepstodo:
+                raise ValueError( f"Unknown through_step: \"{self.parse.through_step}\"" )
```
whohensee, line 338:
```suggestion
                raise ValueError( f"Unknown through_step: \"{self.pars.through_step}\"" )
```

tests/pipeline/test_alerting.py
```diff
@@ -61,7 +61,8 @@ def test_build_avro_alert_structures( test_config, decam_datastore_through_scori
     assert all( len(a['prvDiaNonDetectionLimits']) == 0 for a in alerts )
 
 
-def test_send_alerts( test_config, decam_datastore_through_scoring ):
+# ROB! Remove the "user" fixture, you put it here for interactive debugging
```
whohensee, line 64: rob! How could you forget!

tests/conftest.py
```diff
@@ -39,8 +39,8 @@
 #   at the end of tests.  In general, we want this to be True, so we can make sure
 #   that our tests are properly cleaning up after themselves.  However, the errors
 #   from this can hide other errors and failures, so when debugging, set it to False.
-verify_archive_database_empty = True
-# verify_archive_database_empty = False
+# verify_archive_database_empty = True
+verify_archive_database_empty = False
```
whohensee, line 43: Flip this back, right?

util/logger.py
```diff
@@ -3,8 +3,8 @@
 import multiprocessing
 import logging
 
-_default_log_level = logging.INFO
-# _default_log_level = logging.DEBUG
+# _default_log_level = logging.INFO
+_default_log_level = logging.DEBUG
```
whohensee, line 7: Flip this back too

whohensee (approved):
Caught a few errors, and a few config things that have to be flipped back once testing is done, but otherwise it looks good to me as long as all tests are back to passing.

pipeline/measuring.py
```diff
@@ -61,9 +84,48 @@ def __init__(self, **kwargs):
...
+        )
+
+        self.negatives_radius_units = self.add_par(
+            'negatives_radius_units',
+            'fwhm',
+            str,
+            'Units of negatives_radius (pixels or fwhm)'
         )
 
+        # ROB READ THIS -- think about augmenting the negatives cuts
```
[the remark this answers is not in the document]
rknop: Yesish, but since I'm going to revisit all the measurements later, I will read it then.

tests/conftest.py
rknop: Yeah; I usually flip this during testing.  With this on, if there are any errors that short-circuit a test and its cleanup, it shows up as an error in failing to clean up the database, and I don't see where the error originally happened.

Issue #398, opened 2025-01-07
Re-evaluate analytic cuts


rknop:
They aren't working all that well right now.

Perhaps just import the ones from lensgrinder, which are many-years tried and tested cuts.  However, those use a lot of information that comes out of sextractor, which we don't run on detections right now (at least for zogy), so we probably shouldn't work too hard on this until after we do Issue #396.

rknop:
Among many other things, look at test_measurements.py:test_filtering_measurements

whohensee:
Also be sure to rewrite test_measuring.py::test_measuring. It is currently marked skip until this issue is completed.

Issue #345, opened 2024-08-13
Don't mark objects bad when created


rknop:

rknop:
Right now, it looks like associate_object will only associate bad objects with bad measurements, and good objects with good measurements, so the failure case I described above won't happen.

Still, when pulling out all measurements of an object, you often want to pull the good and the bad, so it still makes  sense to get rid of the is_bad field of Object.

Issue #228, opened 2024-03-28
Better circle edges in photometry


rknop:
Right now (ass of PR 213), the aperture phtoometry class makes a circle by including all pxiels whose center is within the circle.  We need to weight the pixels around the edge based on the intersection of the pixel square and the circle.

One possibility is the shapely library; I haven't used this before, so I don't know it's the right answer, but it looks promising:
```

>>> circle = shapely.Point( [ 10.2, 10.5 ] ).buffer( 3.2 )
>>> square = shapely.Polygon( [ (11.5, 7.5), (12.5, 7.5), (12.5, 8.5), (11.5, 8.5) ] )
>>> circle.intersection(square).area
0.6191305416936494
>>> square.area
1.0
```

rknop:
This issue will become obsolete once the code that uses photutils for photometry is merged, because photutils has its own circular aperture handling.

Issue #210, opened 2024-03-02
Configurable paths to astromatic configurations


guynir42:
For example, PSFex uses a file that is saved by default to `/usr/shared/psfex/psfex.xsl` and that is hard coded in the `Detector` class. Maybe we should allow that to be a config parameter?

PR #401, opened 2025-01-24
Photometry & Measurement Changes

rknop:
Addresses issues #210, #396, #398, #228, #345

The biggest changes are in `improc/photometry.py`, `models/measurements.py`, and `pipeline/measuring.py`.  For `improc/photometry.py`, I completely rewrote the whole file.  The diffs that you'll see try to interleave the changes if it detects some lines that are the same, but it will probably be easier to read if you just look at new file by itself (and maybe look at the old version if you want to see what got deleted).

The good news is that (a) we have fewer crappy things that pass the diagnostic tests, and (b) the nice SN that we have now passes the diagnostic tests.  There is still some crap that passes the diagnostic tests, but hopefully r/b will catch them once we have a properly trained model.

I needed a newer version of photutils than we had, which led me to bump the version of everything in `requirements.txt`.  Unsurprisingly, that led to some other issues.  `sep` has fallen into neglect, so we had to move to `sep_pjw`.  (Hopefully it will be some time before *that* falls into neglect.)  This required a few small edits in the code.  Another thing changed in handling indexing of FITS tables that I don't understand, but that is almost certainly the result of the verison bumps of the libraries; this shows up in `models/decam.py`.  More irritatingly, we no longer seem to be able to update the headers of compressed FITS images without uncompressing and recompressing the data; see Issue #402.

The database migration won't work on a live database... measurements changed too much for that to be worth doing (if even possible at all).  I do have one live database, but it's not very full, and really it's all part of testing, so it's not a big deal to discard it.


tests/models/test_decam.py
```diff
@@ -6,6 +6,7 @@
 import pytest
 
 import numpy as np
+# from astropy.io import fits
```
whohensee, line 9:
```suggestion
```

whohensee: Assuming this is no longer needed.

tests/models/test_measurements.py
```diff
@@ -9,10 +9,10 @@
 from models.provenance import Provenance
 from models.image import Image  # noqa: F401
 from models.measurements import Measurements
-from pipeline.data_store import DataStore
 
 
-def test_measurements_attributes(measurer, ptf_datastore, test_config):
+# ROB!  You put the "user" prefix infor testing, remember to remove it before merging to main
```
whohensee, line 14: rob!

tests/pipeline/test_pipeline.py
```diff
@@ -345,6 +349,10 @@ def test_bitflag_propagation(decam_exposure, decam_reference, decam_default_cali
         assert ds.cutouts._upstream_bitflag == desired_bitflag
         for m in ds.measurements:
             assert m._upstream_bitflag == desired_bitflag
+        # ...why is ds.scores None?  Shouldn't the run
+        #   of the pipeline have run deepscoring?
+        # for s in ds.scores:
+        #     assert s._upstream_bitflag == desired_bitflag
```
whohensee, line 355: This is interesting - if you haven't figured it out I will mark it down to investigate.

whohensee (approved):
Sorry this took a while, I stopped to investigate and understand some of the changes better. Nothing major in here I think, looks good generally. I will also review a few of the more concerning issues that popped up while you were working on this PR.

tests/models/test_decam.py
rknop: I have this here so that I can uncomment it along with the `fits.writeto` statements below that I uncomment to look at images to see what's going on.  (Which, as you can tell since I added this line, happened this time around.)  (Of course, the first time I tried to run it, I got an error because `fits` was unknown.)

tests/models/test_measurements.py
rknop: Did I do that again?

tests/pipeline/test_pipeline.py
rknop: I haven't figured it out, and am still worried about it.

PR #406, opened 2025-01-27
Mem

rknop:
You will recognize some of this... this does (basically) the memory cleanup in zogy_subtract that you did before.  (Slightly rearranged and ocmmented.  Also, all the `=None` wasn't necessary before the del.)  I also added some other places that reduce a bit of memory usage.

The "verify database empty" thing is currently False, but I will set that back before the final merge.

The other changes are things I did so I could keep runnign while NERSC is down.  These are temporary and will need to get changed back in a later PR.  I put in a deliberate ruff failure in conftest.py so that the automated tests on github would all fail fast before they started downloading from my own transfer-limited server.

There were also some small changes to results pipeline numbers.  Not sure why.  Am worried about it, but sticking my head in the sand.  (They aren't significant changes, really.)  This reminds me of something I saw before where github actions and locally I got different numbers, then that went away.  Of course, right now we can't check the github actions results.  So, I'm figuring we'll just go with it until next week and revisit it if the numbers change again and it passes our worry threshold (which it may well not).


rknop:
> Okay, looks good. Nice to finally have that memory reduction implemented. Also hopeful the changed tests are nothing major...

I will revert them next week.

There are a handful of other tests that failed because while I did copy over the SeeChange testing data, I failed to copy over the full set of DECam calibrator files.  Suggestion: do *not* delete your `tests/temp_data` directory until next week, as there is not an easy way to get those calibrator files back now.  (There is the hard way of reassembling them again, but that would be annoying.)

Issue #388, opened 2024-12-16
top_level.py::Pipeline.make_provenance_tree should honor self.pars.through_step


rknop:

rknop:
This interacts with Issue #389 

It may well be that solving #389 will make this into a non-issue.

PR #407, opened 2025-01-28
DataStore provenance tree refactor

rknop:
Adresses issues #405, #388

This one ended up being pretty giant again because it's a refactoring.

This is in preparation for injecting fakes.  Move most of `make_provenance_tree` from Pipeline to DataStore (called `make_prov_tree`).  Remove a hardcoded global list of upstreams and put that inside the function, so that logic can be applied to process upstreams (specifically, whether or not a fake injection step is included) based on parameters and/or additional function arguments in future code.

Along the way, I also simplified DataStore's provenance handling a bit. It can no longer make arbitrary provenances, and can only deal with what it has in its internal provenance tree.  (That tree can be replaced at any time, however.)

I also simplified the provenance tree.  I kept having to put in special-case code to deal with provenance "siblings", so I just decided to remove the whole concept of siblings, give backgrounds, wcses, and zeropoints their own provenances, and let the provenance tree just be a provenance tree.  (I also made 'referencing' provenances proper provenances that can be upstreams of other things.  Previously, subtractions had the upstreams of the referencing provenance as their upstream, which was a special-case exception that was annoyihng.  Now, the Reference object is the upstream of the subtraction, and the referencing provenance is the upstream of the subtraction provenance, which is much more straightforward, if less intuitive to somebody who's not familiar with the data structures.)

(One effect of this is that the upstream of a subtraction is now a zeropoint and a ref, not an image and sources and ref; and, the upstreams of a coadd are not images, but zeropoints.  Guy and I went back and forth on this a lot.  I always thought the upstream of subtraction should just be a zeropoint, and the upstreams of coadds should be a list of zeropoints, because it works best within the logic of the database organization.  You need those zeropoints to do the subtraction or coadd.  From the zeropoint, you can find the wcs, from which you can find the source list, from which you can find the image.  It sounds weird, of course, because it's the image that's subtracted, or the images that are coadded, not the zeropoints.  Because of that weirdness, we ended up with subtraction having an image and a source list as upstreams (it couldn't just be the image, because the process of running the alignment for the subtraction depended on the source list), and that led us to making siblings of the extraction so that we'd have a well known wcs and bg and zeropoint. (Other things went into this discussion, like the fact that we were probably unlikely to ever redo a zeropoint without redoing the extraction.)  But, this didn't make full sense with data structure organization, because a source list implies an image, so having both be an upstream of a subtraction meant duplicated (or, worse, inconsistent) data linkages.  Worse, because the upstreams of coadds were images, not zeropoints, finding the upstreams of a coadd required searching through provenance upstreams to find the right source list, wcs, or zeropoint to of the component images of the coadd.  Now that I get to decide how things go all by myself, I finally decided to change this to what always made the most sense to me.  While it's a lot of changes, and in a few places there are more table joins, overall the structure is simpler now.  By cleaning up the upstream/downstream relationships so that they made sense from a database point of view (if not from an intuitive "what was coadded" point of view), I was able to simplify a bunch of the get_usptreams and get_downstreams code.)

This cascaded to having to then remove/fix a whole bunch of special-case code elsewhere to deal with the fact that zeropoints, backgrounds and worldcoordinates now needed a provenance, and the upstreams of things changed.

I removed the vestigal code having to do with a session stored in a DataStore.  (There was already an exception saying "don't do that" if you tried.  Now I just ripped it all out.)

 I rebooted the alembic migrations again because structure was not getting fully propagated with what alembic generated.  Eventually, we'll have to do a lot of work to make sure we can migrate living databases, but for now, we can still reboot.


pipeline/top_level.py
```diff
@@ -322,24 +306,135 @@ def setup_datastore(self, *args, no_provtag=False, ok_no_ref_provs=False, **kwar
                 if report.exposure_id is None:
                     raise RuntimeError('Report did not get a valid exposure_id!')
             except Exception as e:
+                import pdb; pdb.set_trace()
```
whohensee, line 309:
```suggestion
```

tests/pipeline/test_data_store.py
```diff
@@ -58,158 +195,43 @@ def test_set_prov_tree():
...
+            else:
+                assert last_provs[newproc].id != ds.prov_tree[newproc].id
 
-    ds = DataStore()
-    ds.set_prov_tree( provs )
-    subprov = ds.get_provenance( 'subtraction', {} )
-    assert set( subprov.upstreams ) == { refimgprov, refsrcprov, provs['preprocessing'], provs['extraction'] }
+    # TODO:
+    #   Passing a Provenance to edit_prov_tree
+    #   Verify that provenance tag gets set?
```
whohensee, line 230: Anything here you want to do before merging?

tests/conftest.py
```diff
@@ -48,8 +46,8 @@
 #   at the end of tests.  In general, we want this to be True, so we can make sure
 #   that our tests are properly cleaning up after themselves.  However, the errors
 #   from this can hide other errors and failures, so when debugging, set it to False.
-# verify_archive_database_empty = True
-verify_archive_database_empty = False
+verify_archive_database_empty = True
+# verify_archive_database_empty = False
```
whohensee, line 50: I think this is correct, we want it set to True. I guess False just slipped through the cracks somewhere?

whohensee (approved):
This one was pretty big and took quite a bit of bouncing around to wrap my head around, but once I had that mostly accomplished, I found very few issues. Looks good to me.

tests/pipeline/test_data_store.py
rknop: ...this is a question on many different levels.

For now, I want to leave this as a TODO to go together with a whole bunch of other aspirational TODOs we have in tests about things we probably also ought to test.

tests/conftest.py
rknop: Yeah, I failed to reset it in the previous PR.  (Things were scary during nersc-down week.)

rknop:
Closing this PR.  Something hung on github, and it was stuck in "checking for ability to merge automatically".  I merged manually on my machine and pushed.

PR #411, opened 2025-02-05
Create FakeSet and FakeInjector

rknop:
The first step of injecting fakes: injecting the fakes.

This PR isn't so bad, because it's just added stuff.  Some additions in `default_config.yaml`, as well as `conftest.py` and a couple of other fixtures files to support the new tests.

Next PR will be the abillity to inject fakes next to host galaxies.  (Right now they're just spread uniformly randomly throughout the image.)

PR after that will be integrating the fake injection into the pipeline, which will probably also mean the creation of a FakeAnalysis object that stores info on which fakes were found and how they were measured.  But, that's two PRs down the road, so don't think about it yet.


tests/pipeline/test_fakeinjection.py
```diff
@@ -0,0 +1,78 @@
...
+    ds = fakeinjector.run( ds )
+    assert np.all( ds.fakes.fake_mag >= 23. )
+    assert np.all( ds.fakes.fake_mag <= 25. )
+
+    # Random random seed
+    fakeinjector.pars.random_seed = 0
+    ds = fakeinjector.run( ds )
+    seed1 = ds.fakes.random_seed
+    ds = fakeinjector.run( ds )
+    # Technically, this is flaky, but it will only fail something like 1/2³¹ of the time, so whatevs.
```
whohensee, line 77: Seems inconvenient. I hate it when my tests fail `1/2^31` of the time...

whohensee (approved):
Assuming tests pass, looks good to me. I'm going to have to play around with this a bit, but its a super neat feature to have implemented.

Issue #381, opened 2024-12-06
Iterate sextractor one more time for star/galaxy identification


rknop:
sextractor's star/galaxy identifier depends on decent estimates for `SEEING_FWHM` and `PIXEL_SCALE`.

This is a chicken-and-egg problem, because we don't know the seeing fwhm until we run sextractor and psfex.

Right now, we run sextractor twice, once before, and once after psfex, so that we have an estimate for the seeing.  If we want our star/galaxy separation to be better, we may need to run it three times.  First, to get what's there.  Then, do some sort of basic statistics to come up with a ±20% estimate of the seeing FWHM.  Run a second time to get better star/galaxy separation.  Run psfex. Run a third time to have a good FWHM in our photometry.

sextractor is a nontrivial step (typically takes tens of seconds), so there is a cost to this.

Thought required.

Se comments in `pipeline/detection.py`.

rknop:
Related to Issue #154

rknop:
...we don't have to iterate another time!  PSFex does its own identification of PSFs (if the manual can be trusted), so we can just do what we do now, but for the second time around use the FWHM estimate we got from PSFex.

Issue #341, opened 2024-07-29
What is our convention for storing x and y?


rknop:

rknop:
See `tests/pipeline/test_extraction.py::test_compare_sep_sextractor`

rknop:
Since `zogy.py` uses all `numpy` stuff, I'm pretty confident it's doing the same stuff we do elsewhere.

PR #414, opened 2025-02-06
Sextractor parmaeter tweak, (x,y) documentation

rknop:
Addresses #413, #381, #341


util/logger.py
```diff
@@ -3,12 +3,12 @@
 import multiprocessing
 import logging
 
-_default_log_level = logging.INFO
-# _default_log_level = logging.DEBUG
+# _default_log_level = logging.INFO
+_default_log_level = logging.DEBUG
```
whohensee, line 7: Probably leave this on info.

whohensee (approved):
Certainly an interesting issue. and a good one to have fixed. Changes looked relatively straightforward to me, so long as the new test values pass. Hopefully our ancient FORTRAN-speaking predecessors will keep life easy for us in the future.

PR #415, opened 2025-02-07
Fakes on hosts

rknop:
Issue #410 

Relatively small change; just add a function to `fakeinjector.py` that will select hosts for fakes and position the fakes near them.

Next fakes PR will be running fake images through subtractions and seeing which fakes were found.


whohensee (approved):
Looks good to me, excited to see the fakes coming along.

Issue #419, opened 2025-02-13
some sort of cache problem


rknop:

rknop:
...now I'm having trouble reproducing this...

Issue #417, opened 2025-02-11
Refactor memory tracing


rknop:
With the fakes analysis PR, we have issues with memory tracing.  Reason: some of the processes are run more than once in a single pipeline run.  Subtraction and everything after will be run once for the real image, and once for the fake injected image.  Right now, the various pipeline objects trace memory by saving to DataStore.runtimes with a hardcoded dictionary key (the name of the process).  This means that if we're memory tracing, the values stored the first time through will get rewritten the second time through.

Think of a way to do memory tracing that can dodge this.  Probably a single memory trace around the whole fake injection block is what we'd want.  We need a way of, at runtime, turning off memory tracing.  Perhaps making it an argument to all the pipeline objects' run methods.  Or, perhaps, a global variable somewhere that we can switch on and off, instead of looking at the global SEECHANGE_TRACEMALLOC env variable everywhere.  (Then just look at the env var in top_level.py.)

rknop:
Eh... I'm just doing this in the fakes analysis PR.  Shouldn't have made an issue.  Oh well.

Issue #198, opened 2024-02-05
Injection simulation design


guynir42:
We have yet to give much thought to how the injection simulations will be integrated into our pipeline. 
These are useful as they test the recovery efficiency of the pipeline to transients with different brightness, and are essential in setting limits or rates based on (non-)detections. 

Here are a few things to think about: 
- Where exactly (at what stage) do we insert the simulated sources? Is it at the new image? the subtraction? 
- What do we use to build up the injections? We will likely want the PSF and ZP of the image, but maybe other things? 
- Do we plan to inject the sources into random spots on the image? Is there a plan to avoid bright stars or to put all sources close to galaxies? 
- Is there a plan to inject sources into the same sky position over different images, to build up fake objects? 
- How do we store the simulation parameters that go with the data products, and make sure we don't get confused about what is real? 
- Do we intend to blind the human "scanners" to which events are simulated and which are real? 
- Are the simulation parameters going to affect the provenance of the data products? 
- How many injections do we need to get good statistics on the different parameters of injected sources? What is the parameter space (brightness, time scale, etc.)? 
- Is there an efficient way to inject many simulations into a single image and detect all of them (e.g., by injecting into the subtraction image, or by simply making lots of injections on one image, or by doing a separate subtraction just for the simulated sources and adding it later to the image)? Do we even need something like that?

PR #420, opened 2025-02-13
Fake injection in the pipeline

rknop:
Addresses Issues #198, #417, and, maybe, #419.

This PR ended up more extensive than I wanted it to be.  The main thing was to integrate the fake injection into the pipeline and add some fake analysis.

However, the FakeAnalysis object needed to have the FakeSet and the non-faked DeepScore objects as upstreams.  Reason: FakeSet should be obvious.  It needs the non-faked DeepScore because the fake analysis in the pipeline reruns the steps from subtraction through deepscore.  This must be done with the same parameters that were done for the subtraction, so by making both the FakeSet and the DeepScore an upstream, you've uniquely identified the FakeSet as well as all the parameters used to detect and measure the fakes (plus others).

But... there are lots of DeepScores for one subtraction.  It would be prohibitively cumbersome (and, frankly, absurd) to make _all_ of them upstreams of the FakeAnalysis.  We do need that linkage, though, so that the FakeAnalysis can be found working down.

To solve this, I introduced the DeepScoreSet object, and, by analogy, the MeasurementSet object.  There is a single *Set produced by a run through the pipeline.  That's the thing that is in the upstream/downstream tree, and is the thing that holds the provenance_id.   Measurements are now leaves of MeasurementSet.  They don't point up to cutouts (MeasurementSet does), and they don't have provenance; they only point up to the MeasurementSet object.  They also don't have downstreams of their own; it's the MeasurementSet that has the downstream of DeepScoreSet.  Similar for DeepScoreSet and DeepScore.

But, of course, that was a refactor, so it ended up affecting code all over the place.


whohensee (approved):
A little all-over-the-place, but this one looks good to me. I'll play myself with some of the stuff added once it is merged to main, especially the new measurement and deepscore sets, but I didn't catch anything bad and I'm excited to get the fakes going.

Issue #425, opened 2025-02-26
Unify webap and conductor


rknop:
I think we don't need two web services.  From a user's point of view, it would be nice to have a unified interface.

Should probably implement user groups, so that only users in a proper group can access the conductor functionality.

Issue #423, opened 2025-02-25
Sources getting left behind after exposure deletion


rknop:
I'm playing with DECAM-DDF.  Run three exposures.   Then delete them with `delete_from_disk_and_database`.  However, there are some `sources...fits` files left over after this.  Look into it.

rknop:
The sources files are not in the database, just in the local file store and in the archive.

rknop:
...but, no, that can't be right, the final source list isn't saved until the psf is known.

Still, I wouldn't be surprised if this were related to Issue #421

rknop:
I think this is the culprit; in `detection.py`, function `extract_sources_extractor`, near the beginning there is:

```
        delfiles = [ sourcepath ]
```

And then later, if a psf is found, there is:

```
                    delfiles = [ pathlib.Path(i) for i in psfpaths ]
```

I think I want to make that `delfiles.extend(...)`

Issue #422, opened 2025-02-25
Add a "gaussian fit failed" flag in diagnostics in photometry_and_diagnostics.py


rknop:
Search for this issue number there.

This would then need to be detected in `pipeline/measuring.py`

Right now in `diagnostics` we just set terrible values for the gaussian fit in hopes that they will fail thresholds.

Issue #421, opened 2025-02-25
cannot convert flloat infinity to integer


rknop:
[2025-02-25 19:48:10 -  59 - ERROR] - Exception processing chip S30: cannot convert float infinity to integer
Traceback (most recent call last):
  File "/seechange/pipeline/exposure_processor.py", line 107, in processchip
    ds = pipeline.run( self.exposure, chip )
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/seechange/pipeline/top_level.py", line 509, in run
    ds = self.extractor.run(ds)
         ^^^^^^^^^^^^^^^^^^^^^^
  File "/seechange/pipeline/detection.py", line 308, in run
    sources, psf, _, _ = self.extract_sources( image, wcs=ds.wcs )
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/seechange/pipeline/detection.py", line 379, in extract_sources
    sources, psf, bkg, bkgsig = self.extract_sources_sextractor(image)
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/seechange/pipeline/detection.py", line 424, in extract_sources_sextractor
    psf = self._run_psfex( tempnamebase, image, do_not_cleanup=True )
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/seechange/pipeline/detection.py", line 785, in _run_psfex
    psfdatasize = int( usepsfsize / psf_sampling + 0.5 )
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OverflowError: cannot convert float infinity to integer

rknop:
The issue seems to be where it runs `psfex` in `detection.py`.  It thinks it worked, because `psfex` returned a 0 exit code.  However, when reading the xml with information, the `psf_sampling` was coming out to 0, so really something failed.

To fix: put checking that table inside the thing that checks to see if `psfex` succeeded.

rknop:
...so, yeah, did that, and of course the thing is still failing.

Todo : look at image, decide if it's ok that psfex is failing.  (I.e. is this just a bad chip?)  If not, figure out why.  It may have something to do with the number of identified stars or some such.

[...]

PR #426, opened 2025-02-27
Decat ddf

rknop:
Miscellaneous changes as I work with DECAT DDF.  Led to the creation of, and commenting on, but not full solution of issues #421, #422, #423, #422, and #425.

Nothing really groundbreaking here.  The biggest change is probably adding `main()` to `pipeline/exposure_processor.py` so we can manually run a single exposure without having to set up the whole system so that `pipeline_exposure_launcher` will run it.  Also added a `tools` directory, with a script to look for redundant references, and a couple of nascent scripts to allow submitting jobs as they come into perlmutter batch queues.  (I have only done a little bit of basic testing of those, and will be pushing them harder in coming weeks.)

### DIFF
--- a/improc/photometry.py
+++ b/improc/photometry.py
@@ -515,7 +515,12 @@ def diagnostics( measurements, cutouts, noise_cutouts, mask_cutouts, fwhm_pixels
         # wanted.  It can't handle any nans, and it wants a 1/σ image it
         # calls weight (yes, 1/σ, not 1/σ²).
         gcutout = np.copy( cutout ) - m.bkg_per_pix
-        bad = ( cutout_mask ) | ( np.isnan( gcutout ) ) | ( cutout_noise <= 0. ) | ( np.isnan(cutout_noise) )
+        bad = ( ( cutout_mask ) |
+                ( np.isnan( gcutout ) ) |
+                ( np.isinf( gcutout ) ) |
+                ( cutout_noise <= 0. ) |
+                ( np.isnan( cutout_noise ) ) |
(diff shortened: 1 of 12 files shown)


improc/photometry.py
```diff
@@ -527,28 +532,42 @@ def diagnostics( measurements, cutouts, noise_cutouts, mask_cutouts, fwhm_pixels
...
+            m.position_angle = theta
+        except Exception:
+            # Sometimes the gaussian fitter fails.  I'm getting
+            #   exceptions saying that there are nan or inf values, but
+            #   I'm not seeing it when I look at the actual contents of
+            #   the array, so I don't know what's going on.  We're just
+            #   going to assume that this is a pathological case where
+            #   things are bad anyway and just fill in some values that
+            #   probably fail any cuts. It would be better to set a flag
+            #   saying the gaussian fit failed.  See Issue #422
```
whohensee, line 565: That is interesting. A good issue, but yeah probably low priority.

pipeline/pipeline_exposure_launcher.py
```diff
@@ -60,6 +62,14 @@ class method.
           in which case all we do is download the exposure and load it into the
           database.
 
+        max_run_time : float, default None
+          Normall, when youc all an ExposureLauncher it will loop
```
whohensee, line 66:
```suggestion
          Normally, when you call an ExposureLauncher it will loop
```

whohensee (approved):
Looks good to me. I will review the issues you created as well, though I think I caught all of them in the review here.

PR #427, opened 2025-02-27
Merge webap and conductor

rknop:
The changes are not as big as they look, because of stuff being moved around.

The javascript files did not change that much.  But, lots of stuff was broken out from `webap/static/seechange.js` into four other `.js` files in that directory.

This will simplify deployment, as one fewer server needs to be deployed.


whohensee (approved):
This PR looks good to me. I am mostly leaving ensuring the conductor/webap works right to you, but perhaps once we get closer to being truly online I will have you give me an overview.

Issue #347, opened 2024-08-15
Fix object naming


rknop:
The database refactor removed numeric ids, which the naming scheme for Object depended on.

Right now, the pipeline assigns a placeholder name that is the last 12 characters of the string representation of its id.

Reimplement something (perhaps using a database sequence) that makes more user-friendly names, or decide that we don't care.

PR #428, opened 2025-03-05
Better object names

rknop:
Addresses Issue #347


models/object.py
```diff
@@ -295,142 +303,163 @@ def get_mean_coordinates(self, sigma=3.0, iterations=3, measurement_list_kwargs=
...
-        ----------
-        convention: str
-            The naming convention that will be used to generate the name.
-            Example: SomeText_<instrument>_<yyyy><mm>_<number>
-        present_time: datetime.datetime, optional
-            The time to use as the present time. Defaults to now.
-        session: sqlalchemy.orm.session.Session, optional
-            The session to use for the query. If not given, will open a new session
-            that will be automatically closed at the end of this function.
+        if ( not no_new ) and ( year <= 0 ):
```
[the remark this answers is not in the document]
rknop: I want the default value to thrown an error; the point is that the year needs to be specified.  Probably I should default year to None and check that, as it would make it more obvious.

models/object.py
whohensee, line 1: I'm glad we caught this. I think it is a good lesson to learn for my future PRs, to really look if certain updates are possible/necessary/redundant.

whohensee (approved):

models/object.py
rknop: It's impossible to catch everything... it's all so complicated.

PR #443, opened 2025-03-14
Iterative Sextractor / Background

rknop:
Added the Bijaoui sky subtraction background, decided it was a disaster.  Left the code that led to that decision in (`tests/improc/test_bijaouisk.py`).  It does so much better than other things on really crowded fields that it's sad that it fails so badly on non-crowded fields.  (At least, it does better on the one crowded field I tried.  But, I added this to the lensgrinder pipeline specifically for really crowded fields, and I remember it working well there.)  Feel free to more or less ignore the entire `bijaouisky.py` file, because it's not something we'll probably use.  (There are notes in the docstrings saying not to use those functions.)

Added another sky background method that iterates several times between finding objects, and then doing a sky background masking out pixels that those objects have.  Each iteration uses the sky from the previous iteration, so it should find more and more object pixels as the sky level is less and less overestimatined.  Doesn't make any difference in non-crowded fields.  Makes things substantially better in really freaking crowded fields, but they're still not great.  But, it's probably what we'll live with in the near future.  As I did this, I realized that the whole "sequence" option that I'd added to extraction was ill-considered, so I just removed it.

In service of that, I removed the actual sextractor running code from `backgrounding.py` and put it in a new file (`sextractor.py` in `improc`; `backgrounding.py` now calls that.

Also cleaned up a couple of things.  `sep_pjw` has been deprecated as it has merged back into the now-revitalized `sep` package on PyPi, so I've updated the dockerfile accordingly.  Added some things to `Makefile.am` files that should have been added in previous PRs.  (One of these days I may write a test that runs on an installed SeeChange rather than out of the source tree.  For now, the times things will break when something is left out of a `Makefile.am` is if the webap needs it.)


whohensee (approved):
This looks good to me. I spend a while understanding `pepper_stars`, which was a cool little function. No changes.

Issue #451, opened 2025-03-20
Configchooser fallout


rknop:
This isn't really an issue with config chooser per se, but rather with the fact that we treat our configs as global variables.

Right now, when we read a config, it gets cached.  If it's the one that matches the SEECHANGE_CONFIG environment variable, it gets cached as the default config.

ConfigChooser then modifies the default config based on whether it's a galactic or extragalactic filed.  This modifies the global variable.

A second run of Pipeline within the same running python context then breaks, because it no longer has the base config that ConfigChooser assumed, but one of the ones that ConfigChooser supplied.  We didn't notice this in our tests because we don't run Pipeline all that often, and because the ConfigChooser tests themselves restore state.  As I was working on something else, I ran into the issue.

The right solution is to back down from global variables.  (As always.)  The config for a run of the pipeline should exist within the Pipeline object, not globally, and the Pipeline object should pass it to subtasks.

rknop:
The assumption of a correct global config is so baked into our code in so many places that I don't think we can get away from it.  (I mean, we could, but that would be a big and unpleasant refactor (...which I got a fair way into before deciding I hated it...), and would include passing Config objects around between lots of functions, which becomes ugly.

Alternate solution:  put a warning in the ConfigChooser comments that you're making a mess when you use it.  The one place we use it, in `pipeline/exposure_processor.py`, make sure to clean up the mess after using ConfigChooser.

rknop:
Also, make it so that `Config.set_value` won't work on any of the configs that are cached in the `Config` class, to discourage real-time updating of the global configs, and make you think about what you really want to be doing with config if you're trying to do that.

rknop:
....in fact, ConfigChooser has been deprecated and ripped out.

Each process that might have multiple configs now has it baked into the config for that process, so a single provenance can handle both galactic and extragalactic configurations.

PR #452, opened 2025-03-20
Config redux

rknop:
See Issue #451


whohensee (approved):
Yes, this seems like the best solution for now. A little ugly in those few tests, but it looks to me like it will work how we need it to.

PR #453, opened 2025-03-27
Conductor and Web Updates

rknop:
Implement some not-yet-implemented buttons in the conductor.

Add some fields to report and knowlexposure so the conductor can show when an image started and when an image finished.

Add a "through step" option to the conductor.

Add the ability to show image reports, and to plot results of fake analysis, for images and exposures.

Other misc changes mostly to the webap.

(Also, the process names for astrometic and photometric calibration had become confused. Sometimes 'astrocal' and 'photocal', sometimes 'wcs' and 'zp'. I tried to unify them all as 'astrocal' and 'photocal'. What a mess that was.)


tests/conftest.py
```diff
@@ -55,8 +55,8 @@
 #   at the end of tests.  In general, we want this to be True, so we can make sure
 #   that our tests are properly cleaning up after themselves.  However, the errors
 #   from this can hide other errors and failures, so when debugging, set it to False.
-verify_archive_database_empty = True
-# verify_archive_database_empty = False
+# verify_archive_database_empty = True
+verify_archive_database_empty = False
```
whohensee, line 59: might want to swap this back

whohensee (approved):
Looks good. I found a few issues, but nothing major. They do need to be corrected though. I also did not hold the `.js` files to a high level of scrutiny, so it is possible stuff may have scraped by in there onto the webap.

Issue #457, opened 2025-04-14
Add alerts to pipeline!


rknop:
We have an alert sending object, and we test it sending to a Kafka server... but right now the main pipeline doesn't use it!

Add that!

PR #460, opened 2025-04-15
Alerts in Pipeline, Hopskotch / SCiMMA

rknop:
Some edits around alerts.  Realized I needed to refactor the config section for alerts a little bit so we could do things like hide passwords in secrets files.  Added alerting to `top_level.py` (Issue #457), and refactored that one a little bit so that we can save after the pipeline is done but before sending alerts.   (That way, unless we configure it otherwise, we never send out alerts on something we don't have successfully saved to our database.)  Fixed a few stupid bugs and typos that may or may not have been related.

Testing not using our formal tests, but using the decat-ddf data, I have successfully sent some alerts to SCiMMA, which is how we're going to distribute public alerts for LS4.


pipeline/exposure_processor.py
```diff
@@ -114,9 +114,9 @@ def start_work( self ):
             if len(rows) > 1:
                 raise RuntimeError( f"Error, multiple known exposures with instrument {self.instrument} "
                                     f"and identifier {self. identifier}; this should not happen" )
-            if rows[0][0] != self.cluster_id:
-                raise ValueError ( f"Error, known exposure with instrument {self.instrument} and identifier "
-                                   f"{self.identifier} is claimed by cluster {rows[0][0]}, not {self.cluster_id}" )
+            # if rows[0][0] != self.cluster_id:
+            #     raise ValueError ( f"Error, known exposure with instrument {self.instrument} and identifier "
+            #                        f"{self.identifier} is claimed by cluster {rows[0][0]}, not {self.cluster_id}" )
```
whohensee, line 119: Looks like these can probably be killed.

whohensee (approved):
Looks straightforward enough. We'll have to remember to do the proper testing for hopskotch since its not in the automated suite, but that shouldn't be too bad. (Better not be me-from-the-future reading this and laughing at this naivete!)

PR #463, opened 2025-04-16
End-to-end pipeline tests with regression tests

rknop:
Added two tests that run the full pipline and then check to make sure that the detected sources positions, fluxes, and other measurements match what's expected.  These can serve as a regression test.

One test for zogy, the other for alard/lupton w/hotpants.  The latter pointed out some problems in detection that made it not work after an alard/lupton subtraction, so those were fixed.

### DIFF
--- a/models/psf.py
+++ b/models/psf.py
@@ -198,7 +198,7 @@ class PSF(Base, UUIDMixin, FileOnDiskMixin, HasBitFlagBadness):
              it from the database.  Use this for efficiency, or if you
              know the image isn't yet in the database.
 
-          filenamer_is_absolute : bool, default False
+          filename_is_absolute : bool, default False
              If False (default), then filename is relative to the local
              store root.  If True, then filename is an absolute path
              (and must be specified).  In this case, the psf object's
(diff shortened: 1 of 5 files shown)


models/psf.py
```diff
@@ -195,6 +198,14 @@ def save( self, filename=None, image=None, sources=None, **kwargs ):
              it from the database.  Use this for efficiency, or if you
              know the image isn't yet in the database.
 
+          filenamer_is_absolute : bool, default False
```
whohensee, line 201:
```suggestion
          filename_is_absolute : bool, default False
```

tests/seechange_config_test.yaml
```diff
@@ -96,3 +96,4 @@ fakeinjection:
   host_maxmag: 0.5
   host_distscale: 1.
   random_seed: 42
+
```
whohensee, line 99:
```suggestion
```

whohensee (approved):
Looks good. I like having some single-full-run tests rather than the very heavy `test_data_flow`.

PR #464, opened 2025-04-18
PSF refactor, fix zogy shifts

rknop:
Fixed the problem where zogy subtractions was shifting the flux around.  This is verified with a new test `tests/improc/test_zogy.py::test_zogy_positioning`.

The issue had to do with where zogy expected the PSF to be centered (and we weren't quite meeting that expectation).  It was a little bit mysterious, and probably has to do with numpy's `fftshift` and which pixel it expects the 0 frequency to show up on.  (See comment in `improc/zogy.py::zogy_subtraction`.)

In order to fix this, I removed zogy's custom padding routine, and refactored zogy to take a PSF object instead of an image of the PSF.  I then added a method to PSF that gives you a centered image-size PSF image... with an optional offset so that zogy could put it where it wants it.  I also had to clean up the psfex rendering a bit so that it was consistently placing things offset by half a pixel.  (Because of rounding conventions and even vs. odd numbers, it was sometimes going one way, sometimes the other...!)  To test all this, I needed to add a delta function and gaussian PSF, and then because of this refactor, I needed to add an image PSF to use with the simulator and zogy tests.  Rather than have lots of if statements in lots of functions (which is what we usually do when there are different formats or algorithms for some kind of object), I refactored PSF to have subclasses, which meant learning about SQLAlchemy polymorphism (which, for this use case, turns out to be pretty straightforward).

As a side effect, I also modified the image simulator so that you could pass in a random seed, allowing it to be repeatable.  This is a gratuitous addition to this PR.  Originally, I thought I was going to use it in my new tests, but the simulator is way too heavyweight for what I needed.  But, when I thought I was going to do that, I really didn't want to write more flaky tests.  So, as a nice side effect, we no longer have to mark a bunch of zogy tests as flaky that we used to have to mark as such.

As a side effect to the side effect, I started wanting to see the output of some simulator tests, and i so doing I refactored a bit what `INTERACTIVE` means when running tests.  (Guy didn't run in a docker environment, so could show plots interactively.  I wanted to make plots without showing the interactively, so `INTERACTIVE` wasn't enough.  I split it into `MAKE_PLOTS` and `INTERACTIVE`.)


models/psf.py
```diff
@@ -569,3 +790,388 @@ def get_downstreams( self, session=None ):
...
+        else:
+            psfpath = pathlib.Path( self.local_path ) / f'{filename}.fits'
+            psfxmlpath = pathlib.Path( self.local_path ) / f'{filename}.xml'
+
+        # header0 = fits.Header( [ fits.Card( 'SIMPLE', 'T', 'This is a FITS file' ),
+        #                          fits.Card( 'BITPIX', 8 ),
+        #                          fits.Card( 'NAXIS', 0 ),
+        #                          fits.Card( 'EXTEND', 'T', 'This file may contain FITS extensions' ),
+        #                         ] )
+        # hdu0 = fits.PrimaryHDU( header=header0 )
```
whohensee, line 874: Do we want this comment block here?

models/psf.py
```diff
@@ -569,3 +790,388 @@ def get_downstreams( self, session=None ):
...
+
+        if psfxmlpath is not None:
+            raise ValueError( "psfxmlpath is meaningless for ImagePSF" )
+
+        if psfpath is None:
+            psfpath = self.get_fullpath( download=download, always_verify_md5=always_verify_md5, nofile=False )
+
+        with h5py.File( psfpath, 'r' ) as h5f:
+            if 'psf' not in h5f:
+                raise ValueError( "No psf gropu found in the file" )
```
whohensee, line 1171:
```suggestion
                raise ValueError( "No psf group found in the file" )
```

tests/improc/test_zogy.py
```diff
@@ -37,6 +105,7 @@ def test_subtraction_no_stars():
         pixel_qe_std=0,  # keep pixel QE constant between images
         gain_std=0,  # keep pixel gain constant between images
         star_number=0,
+        random_seed=42
```
whohensee, line 108: seems like a good seed to me :)

tests/improc/test_zogy.py
```diff
@@ -424,8 +491,8 @@ def test_subtraction_seeing_background():
                         expected = f * np.sqrt(np.sum(P ** 2) / (B + f * np.sum(P ** 2)))
                         measured = np.nanmax(c)
                         # SCLogger.debug((expected - measured) / expected)
-                        # TODO: figure out why we still have cases where the S/N is reduced by 35%
-                        if abs((expected - measured) / expected) > 0.35:
+                        # TODO: figure out why we still have cases where the S/N is reduced by 37%
```
whohensee, line 494: Is this either something that should be done for this PR or made into an issue?

tests/models/test_psf.py
```diff
@@ -276,3 +492,269 @@ def test_psfex_rendering( psf_palette ): # round_psf_palette ):
...
+    # So, to generalize, for even-length sides, the center is at (n//2 -
+    # 0.5), and for odd-length sides, the center is at (n//2) (both for
+    # 0-offset arrays).
+
+    pixes = [ 511, 512, 513, 514 ]
+    ctrs = [ 255., 255.5, 256., 256.5 ]
+    intparts = [ 0, 1, -1, 2, -2 ]
+    fracparts = [ 0, 0.1, -0.1, 0.5, -0.5, 0.7, -0.7 ]
+
+    # 6 nested for loops FTW!  N⁶ baby!
```
whohensee, line 694: Thats gotta be a record

whohensee (approved):
This was a pretty big one, and I will go through some of the math again to fully understand. It looks good to me and I will submit my approving review, but if I catch anything else when I go through again then I'll mention that.

tests/models/test_psf.py
rknop: Guy had a comment somewhere in the  tests that read "beware the 4-loop".  So  I know I've beaten that one at least.

PR #469, opened 2025-05-13
Different code versions for different parts of code

whohensee:
Removes git hash codeversion functionality and replaces them with codeversions that have a `process`, `version_major`, `version_minor`, and `version_patch` column, following the concepts of semantic versioning.

Codeversions now are a `UUIDMixin`, and provenance generates the codeversion portion of its hash now by using the `version_major` and `version_minor` columns. `version_patch` is intended for changes that do not affect the data products of the pipeline.

Now when coders are working on the pipeline, they are expected to increment the version numbers in the dictionary located in the `CodeVersion` definition as above.


models/provenance.py
```diff
@@ -344,43 +372,31 @@ def update_id(self):
         if self.process is None or self.parameters is None or self.code_version_id is None:
             raise ValueError('Provenance must have process, code_version_id, and parameters defined. ')
 
+        # for hash get the static versions from codeversion rather than UUID which changes each run of tests
```
rknop, line 375: Hmm... this didn't occur to me when we were talking about this earlier, but it does now.  Right now, at least formally, the provenance is tied to a specific code version (which means a specific process,major,minor,patch).  What you here makes the provenance ID only depend on the major and minor parts of the patch.

That leads to a bit of an inconsistent database design, as the provenance is linked both to a specific maj.min.pat and just a general maj.min.  I'm not sure this is any more than an aesthetic issue, as in practice when you make provenances, the ID (which is the thing that uniquely identifies the provenance) is based only on maj.min.

Where it might be a problem is if the `code_version_id` field of a `Provenance` object is ever interpreted anywhere to get a code version.  You'll end up with a mix of different code versions of different patches based on how each one was generated.

For now, I think this is OK, but a future project might be to make the `code_version_id` of `Provenance` private... well, you can't do that in Python the way you can in Java, but you can tell people you're trying by naming it `_code_version_id`.  Then, write either a `@property code_version_id` or a method to get the actual `code_version_id`, and have it get the higest patch that matches (process, major, minor) of the provenance.

Hmm.

I hope I didn't make more of a mess with my last minute suggestion to link provenance to only major, minor.  Since all the tests pass,  I think we should leave this as in the PR, but it's something maybe worth more thought in the future.

models/provenance.py
```diff
@@ -394,20 +410,38 @@ def get_code_version(cls, session=None):
...
+            #                                        .order_by(CodeVersion.version_major.desc())
+            #                                        .order_by(CodeVersion.version_minor.desc())
+            #                                        .order_by(CodeVersion.version_patch.desc())).first()
+
+            # ISSUE consider raising exception if there exists a more up-to-date version than the hardcoded
+
+            codebase_semver = CodeVersion.CODE_VERSION_DICT[process]  # (major, minor, patch) eg. (2,0,1)
+            with SmartSession() as sess:
+                code_version = sess.scalars(sa.select(CodeVersion)
+                                               .where( CodeVersion.version_major == codebase_semver[0])
```
rknop, line 429: I think you want to add a `.where( Codversion.process==process )` here.

models/provenance.py
rknop: Issue #477.

tests/conftest.py
```diff
@@ -216,7 +216,14 @@ def pytest_sessionfinish(session, exitstatus):
 
         # delete the CodeVersion object (this should remove all provenances as well,
         # and that should cascade to almost everything else)
-        dbsession.execute(sa.delete(CodeVersion).where(CodeVersion._id == 'test_v1.0.0'))
+        # dbsession.execute(sa.delete(CodeVersion).where(CodeVersion._id == 'test_v1.0.0'))
```
rknop, line 219: I think in the interest of cleaning up, and making sure all Provenances and so forth get deleted (which this use to do), we should add whatever the SQLA idiom for `TRUNCATE TABLE code_versions CASCADE` is here.  (Or, just write the SQL.)

rknop: ...Note that if we follow one suggestion I made in Issue #477, that will no longer work.  I noted that in Issue #477.  For now, though, we should put this here to maintain the cleanup function.

tests/conftest.py
```diff
@@ -363,34 +370,10 @@ def test_config():
...
-        assert len(them) == 0
-
-
 @pytest.fixture
-def provenance_base(code_version):
+def provenance_base():
     p = Provenance(
-        process="test_base_process",
-        code_version_id=code_version.id,
+        process="test_process",
```
rknop, line 376: See Issue #478.  (No changes needed here, just something potentially painful I'm thinking about for the future.)

tests/fixtures/decam.py
```diff
@@ -377,7 +376,7 @@ def decam_datastore(
         decam_exposure,
         'S2',
         cache_dir=decam_cache_dir,
-        cache_base_name='007/c4d_20211025_044847_S2_r_Sci_IDDLGQ',
+        cache_base_name='007/c4d_20211025_044847_S2_r_Sci_CYNFGI',
```
rknop, line 379: ...I wonder if the barf in the filename here is related to Issue #474?

tests/models/test_image_propagation.py
```diff
@@ -285,7 +285,7 @@ def make_associates( image, extra="" ):
         assert ref.bitflag == sim_image2.bitflag
         sim_image4 = Image.from_new_and_ref( zp3, ref )
         improvs = Provenance.get_batch( [ refprov.id, sim_image3.provenance_id ] )
-        prov4 = Provenance( process='testsub',
+        prov4 = Provenance( process='test_process',
```
rknop, line 288: Yeah, this is the right thing to do for now.

Not sure what I'd do with these if I ever do Issue #478

rknop (changes requested):
This looks good.  This is one of those really kind of nasty changes that send tendrils everywhere, and is harder to do than the amount of code changed would suggest.

A number of comments, many of which are musings that led me to create two new issues.

There are two small changes I think we need, which I comment on.  One is in `models/provenance.py` aroudn line 428 (just a missing where in a query, I think), and the other is in `test/conftest.py` line 219 (you commented out something where we were depending on the side effects to empty out database tables).

models/provenance.py
whohensee: Good catch!

models/provenance.py
whohensee: I think this is a valid concern as an otherwise identical provenance with two different codeversions (`1.1.1` and `1.1.2`) would end up with the same hash.

tests/fixtures/decam.py
whohensee: intriguingly the errors related to this cache path don't even pop up in `test_decam.py`, I had to run some other tests to find errors related to this and change it.

rknop (approved):

PR #472, opened 2025-05-15
Add code to calculate an object position based on a weighted mean of source detections

rknop:
This is not integrated into the pipeline; the idea is that this is something that would be run alongside, either (say) daily to update future alerts, or as part of a "data release" or similar.


whohensee (approved):
Looks good after talking about it on zoom

PR #480, opened 2025-05-15
Fiddle with logger, fix a horrible bug, add tests for logger

rknop:
I imported this logger over to the Roman SN PIT.  Made some changes there and, crucially, wrote tests.

So, I figured it was only fair to import all that back.


whohensee (approved):
I didn't catch anything, looks good and I won't tell anyone we didn't already have these if you won't. Good if tests pass.

PR #488, opened 2025-05-21
Change the locking mechanism for calibrator files so that dead process / stale locks don't stop things from working

rknop:


models/calibratorfile.py
```diff
@@ -237,12 +239,99 @@ def __repr__( self ) :
...
+        constantly.  If you go to get a lock, see that somebody else has
+        it, but the promise hasn't been renewed recently enough, you can
+        grab it.
+
+        (This DOES assume that every computer running the pipeline, and
+        the database server, are all time-synced.  Please use ntp!  If
+        the clocks are off by seconds, then the timeouts may be too fast
+        or too slow.)
+
+        What a mess.
```
whohensee, line 334: Yikes, that is indeed a mess.

whohensee (approved):
Okay, looks good. Just a few of those tests regarding multiprocessing that weren't added to make sure we aren't leaving processes around sound worth maybe turning into a longer-term issue? Or at least keeping in mind if such an issue pops up in the future.

PR #499, opened 2025-06-24
Fix pipeline exposure launcher hang

rknop:
The only real substantive change in this PR is the single line added to `scoring.py`.  It was causing tests to hang because we did stuff with pytorch in one process, and then tried to do more stuff with pytorch in a different process.  In fact, we also do that when running pipeline_exposure_launcher in production, but there the first process dies before the second process starts.  pytorch must be doing something strange with threading and so forth given the nature of the fix.

Some other changes here were put in for debugging, and left in because they will be useful for future debugging.  (In particular, all the changes to `updater.py` come from debugging problems that turned out not to be code problems, but the noirlab archive being offline.)

### DIFF
--- a/docker/application/Dockerfile
+++ b/docker/application/Dockerfile
@@ -232,7 +232,7 @@ RUN source /venv/bin/activate \
 
 # Pytorch
 RUN source /venv/bin/activate \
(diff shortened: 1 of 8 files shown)


whohensee (approved):
Looks fairly straightforward, approved!

PR #500, opened 2025-06-25
Implement "skip_existing" in decam.py downloading exposures

rknop:
Addresses Issue #362


whohensee (approved):
As long as this new test passes and we are happy enough with the 1 second test, looks good and straightforward to me.

Issue #495, opened 2025-06-23
Conductor: Add sorting by field and filter on the exposure list


whohensee:

rknop:
Added sorting for most columns for both the Conductor known exposures list and the Exposure list.

Issue #494, opened 2025-06-23
Conductor Bug: Cannot show known exposures by date window


whohensee:
This one is pretty big - Currently when trying to display known exposures over some date window the page just hangs and never displays any. Thus the only viable option is to show all known exposures, and that list grows quite large... 

Additionally, it would be nice to show additionally the number of displayed exposures.

Issue #493, opened 2025-06-23
Conductor: when showing provenance tag, show code version details better now that it is semantic


whohensee:
Currently we display the code version `_id`, which is quite non-human-readable. Since we now have semver (eg 1.0.2), review better options to display in this menu.

Issue #491, opened 2025-06-23
Conductor UI has 2 different "release" terms (confusing)


whohensee:
Note that perhaps the "release" button next to "hold" should be changed to something like "unhold", since there is also a release_time, which means something different (when the worker finished with it, not when it was unheld by the webap).

rknop:
I made the button "Clear Hold".

PR #501, opened 2025-06-27
Conductor Updates

rknop:
Issue #494 : fix by putting in date validation, which will add the HH:MM:SS field automatically.

Issue #491 : replace one of the "Release" button texts with "Clear Hold"

Issue #495 : Lots of sorting for both known exposure in the Conductor, and the exposure list in Exposure Search, is now available.  Also added more filtering on known exposures.

Issue #493 : Provenances now show semantic versions.

Also, incidentally, cleaned up the .gitignore file, removing a whole bunch of boilerplate that was there.  (As a rule, I don't like starting with a "here's everything everybody ever wanted to ignore" in .gitignore, but much prefer adding stuff only as we really know we want to ignore it for a given project.)

Lots of yaml files in spin/rknop-dev edited.  This is just for my own personal dev install on spin, so can be ignored.


.gitignore
```diff
@@ -13,118 +14,53 @@ data/DECam_default_calibrators
...
-
-# Environments
+### Stuff created by autoconf and ./configure
+# THOUGHT REQUIRED : do we want to add the various
+#   Makefile.in files from autoconf to the archive?
+#   That's often what people do.  Right now,
+#   installation requires rerunning autoconf,
+#   which can be inconvenient.  If we ever
+#   want to drop that requirement, then we
+#   need to remove some of these following lines
```
whohensee, line 41: Do you think this is worthy of an issue?

whohensee (approved):
Looks good. I don't see any issues in the code, and I'll keep an eye out while using the new conductor.

Issue #471, opened 2025-05-15
Clean up Process Names


whohensee:
As of writing, the codeversion dictionary is filled with process names that are either duplicates or poorly thought-out in a system where each process requires a codeversion that we expect to maintain and keep version numbers of. We should think of a proper naming scheme for the processes and objects that go with them, and track down where some of these gratuitous process names are coming from and prune them.

Here is the dictionary at time of writing.
```
    CODE_VERSION_DICT = {
        'preprocessing': (0,1,0),
        'extraction': (0,1,0),
        'astrocal' : (0,1,0),
        'photocal' : (0,1,0),
        'subtraction': (0,1,0),
        'detection': (0,1,0),
        'cutting': (0,1,0),
        'measuring': (0,1,0),
        'scoring': (0,1,0),
        'bg': (0,1,0),
        'wcs': (0,1,0),
        'zp': (0,1,0),
        'test_process' : (0,1,0),
        'referencing' : (0,1,0),
        'download': (0,1,0),
        'DECam Default Calibrator' : (0,1,0),
        'import_external_reference' : (0,1,0),
        'no_process' : (0,1,0),
        'alignment' : (0,1,0),
        'coaddition' : (0,1,0),
        'manual_reference' : (0,1,0),
        'gratuitous image' : (0,1,0),
        'gratuitous sources' : (0,1,0),
        "acquired" : (0,1,0),
        'fakeinjection' : (0,1,0),
        'exposure' : (0,1,0),
    }
```

rknop:
Looking at some of the code where the new code versions are used -- e.g. see `alignment.py`, search for ` code_version = Provenance.get_code_version( process='alignment' )` -- it may be that having a process name field in provenance is now redundant, because the code version has a process name, and the provenance has a code version id foreign key.

We wouldn't want to try to fix that without first cleaning up code versions in this issue, so I'm adding this as a comment here rather than making it its own issue.

rknop:
Looking at `instrument.py`, `get_exposure_provenance` , that's a place where we let *anything* be passed in as the process, so that's something that would need to get cleaned up as part of this.

Perhaps we could make the process always be just 'exposure', but then a parameter would be "acquisition_method", and that could be set to "download" by default.  (That function does set parameters of the Provenance it generates ad hoc.  Most processes set the parameters based on a Parmaeters object.)

rknop:
[...]

Issue #365, opened 2024-11-05
Update tests for alerts with previous detections


rknop:
In `tests/pipeline/test_alerting.py` , the test `test_alerts_with_previous` is currently marked to be skipped.  Reason: I found images in the DECat pipeline that used the reference were already using in tests (so I didn't have to rewrite a whole bunch of other tests if we changed our test DECam images) that had detections.  However, SeeChange is not finding them.  It's possible that the detections were not real (though I'm skeptical; they look like legitimate quasars) and the DECat pipeline detected them where it shouldn't, and zogy is properly missing it.  It's also possible that zogy is missing real things.

Once more investigation into what we're catching in subtraction is done, come back and revisit this test.

rknop:
Perhaps use the sim_lightcurve fixtures I added in PR #510

rknop:
PR #510 adds tests for alerts with previous detections.

Created issue #517 to look into what was up with those DECam exposures.

Issue #346, opened 2024-08-15
Update measurement searching in Object


rknop:
The measurement searching in Object needs to be redone after the database refactor.

I want to rethink it a bit as well.  The old way depended on just python-filtering the measurements field of the Object object.  (Which was a little bit horrifying; having an eager-loaded measurements field turned one of the lightest-weight database objects (Object) into one of the heaviest-weight (Object plus an index scan of our biggest table).  This is one of many reasons to ditch sqla relationships.)

Redoing it will require putting in some database queries.   This cascades to several other functions.  I want to think about primary use cases; primarily, we're going to want to search based on a ProvenanceTag (which isn't in there right now), and probably use a default one.

rknop:
When `pipeline/alerting.py` was written, the functon `build_avro_alert_structures` put in the database queries to get previous measurements of an object given the current tree of provenances in the data store passed to the `Alerting` object.

This code could (should?) be moved out to Object so it can be used outside of the context of generating alerts.

rknop:
Measurement searching is basically gone.  It's not something we want to routinely support.  One can, of course, always construct custom database queries.  However, given that when we have a new detection, we associate it with an object, then measurement searching should just be measurement filtering; the searching is just a mater of matching the object id and the provenances, which is straightforward.

Filtering is supported by `Object.get_measurements_et_al`

Issue #331, opened 2024-07-15
De-flakify tests


rknop:
The flaky tests are occasionally annoying, as sometimes they fail.  They also can get in the way of trying to track down where something goes wrong in a full test run with `pytest -v --pdb`, becasue you'll get an interruption on the first run of a flaky test even if it's going to fail later.

Make all the flaky tests non-flaky by using predefined random seeds.

There's an argument to be made that we're testing the code statistically by using these tests with different random seeds.  However, because what we always end up doing is just rerunning the tests when flaky tests fail (rather than investigating why the failed), this isn't really working.  What's more, a flaky of 3 or 5 isn't a lot of statistical samples to *really* test it; the test fail often enough that we would waste a lot of time doing the investigation of why the flaky tests sometimes failed if we really did do that investigation.  As a result, they're an annoyance that aren't really serving the purposse that they nominally serve.

It would be worth putting in statistical tests that do *real* statistical tests (so, enough trials that your statistics are more robust) to make sure variances are coming out more or less where expected, but then perhaps gate those behind an environment variable so that they aren't run as part of github actions.

rknop:
At least one flaky test was replaced with a non-flaky test when measurements was overhauled (PR #401, Issues #396 and #398).

rknop:
Some progress made with the PSF / Zogy shift PR (PR #464 ) -- the simulator can now take a random seed, and the various zogy tests that use it pass a random seed (usually 42).  There are now only one or two remaining flaky tests.

PR #510, opened 2025-07-30
Add matches to Legacy Survey objects to alerts

rknop:
This PR ended up pulling me down into a rabbit hole, and I ended up doing more than was needed for what's described in the title.

Because I was touching object/measurement association, I ended up adding some tests for that.  (Done to replace tests that have been disabled, for code that I think we'll just deprecate.  There is code that tries to find the measurements of objects by searching on coordinates, but measurements all have an object_id, so the more simple search is how we're just going to do it.  We can throw out that code rather than fixing it to work again, and throw out the corresponding tests.)

To do the object/measurement association test, I needed a lot of data structures in there.  Each measurement needed everything back to an image to do all of this searching and association.  Instead of figuring all of that out, I just built up a new set of simulated subtractions and lightcurves that start from a 256×256 image created by improc/simulator.py, so that the pipeline would just fill in all the stuff we need to test.  Developing that took a while (basically a solid day plus).

In doing all that, I got annoyed by a whole bunch of other things, and ended up fixing them.  Some were minor move about / clean up stuff.  One was something I realized I could do now that I made my new fixtures.  At least two had an issues (#423, #365).

Added random seeds to lots of tests so we could finally get rid of all flaky tests.  (Well, except for the memory free ones, but those are also marked skip because the python garbage collector is not something one can understand.)  Issue #331.

Ended up refactoring data store's provenance tree builder a bit.  This was motivated by the need to handle an image as a starting point, rather than an exposure, in the case where that image is already preprocessed so there shouldn't be a preprocessing provenance (and then extraction shouldn't have a preprocessing provenance as an upstream).  This also required some small changes in `top_level.py` and `preprocessing.py`.

Made it so that you can pass a psf to Detector even if it's not a subtraction image, so in my contrived simulated case I didn't have to run psfex over and over again knowing that the psf was going to be the same for all of the images.

Changed the default zeropoint in alerts from 27.5 (the (soon-to-be-deprecated) SNANA value) to 31.4 (for nJy, which is what LSST users).

Got sidetracked and cleaned up process names in Provenance.py and elsewhere.

Overall, this PR addresses at least Issues #346, #511, #515, #471, #423,  #331, #365, #518, #512


models/object.py
```diff
@@ -68,163 +71,117 @@ def __init__(self, **kwargs):
...
+        min_deepscore : float, default None
+          If given, minimum deepscore of measurements to return
 
-            local_thresh.update(new_thresh)  # override the Measurements object's thresholds with the new ones
+        Returns
+        -------
+          dict of lists, all lists having the same length
+          First four keys are always there; last two only if deepscore_prov_id is not None
+           { 'measurements': list of Measurements,
+             'meaurementsets': list of MeasurementSet,
```
whohensee:
```suggestion
             'measurementsets': list of MeasurementSet,
```

models/object.py
```diff
@@ -68,163 +71,117 @@ def __init__(self, **kwargs):
...
+             'images': list of Image (the difference images, *not* the original science image!),
+             'zeropoints': list of ZeroPoint,
+             'deepscores': list of DeepScore,
+             'deepspscoresets': list of DeepScoreSet
```
whohensee:
```suggestion
             'deepscoresets': list of DeepScoreSet
```

pipeline/detection.py
```diff
@@ -355,23 +376,23 @@ def run(self, *args, **kwargs):
...
 
                 if sources is None or psf is None:
-                    # TODO: when only one of these is not found (which is a strange situation)
-                    #  we may end up with a new version of the existing object
-                    #  (if sources is missing, we will end up with one sources and two psfs).
-                    #  This could get us in trouble when saving (the object will have the same provenance)
-                    #  Right now this is taken care of using "safe_merge" but I don't know if that's the right thing.
+                    # TODO: think about what it means when only one of these is set.
+                    #   That probably shouldn't happen, since both are determined
+                    #   at the same time!  It probably means database corruption.
```
whohensee, line 381: Should we make this an issue?

tests/fixtures/simulated.py
whohensee, line 1: The tests in this file must have been a whole journey to write... Nice work! They will be very useful to have.

tests/models/test_image_querying.py
```diff
@@ -62,25 +62,25 @@ def test_image_cone_search( provenance_base ):
...
         image1.end_mjd = image1.mjd + 0.007
-        _1 = ImageCleanup.save_image( image1 )
+        _1 = ImageCleanup.save_image( image1, seed=1397792661 )
         image1.insert()
 
         image2 = Image(ra=120.0002, dec=9.9998, provenance_id=provenance_base.id, **kwargs )
         image2.mjd = rng.uniform(0, 1) + 60000
         image2.end_mjd = image2.mjd + 0.007
-        _2 = ImageCleanup.save_image( image2 )
+        _2 = ImageCleanup.save_image( image2, seed=1693252949 )
```
whohensee, line 71: How did you come to these seed values? Just a random 10-digit integer? Is there any advantage to that over just using any other distinct integers, for example `1, 2, 3,...`?

tests/conftest.py
```diff
@@ -55,8 +51,8 @@
 #   at the end of tests.  In general, we want this to be True, so we can make sure
 #   that our tests are properly cleaning up after themselves.  However, the errors
 #   from this can hide other errors and failures, so when debugging, set it to False.
-verify_archive_database_empty = True
-# verify_archive_database_empty = False
+# verify_archive_database_empty = True
+verify_archive_database_empty = False
```
whohensee:
```suggestion
verify_archive_database_empty = True
# verify_archive_database_empty = False
```

util/logger.py
```diff
@@ -3,8 +3,8 @@
 import multiprocessing
 import logging
 
-_default_log_level = logging.INFO
-# _default_log_level = logging.DEBUG
+# _default_log_level = logging.INFO
+_default_log_level = logging.DEBUG
```
whohensee:
```suggestion
_default_log_level = logging.INFO
# _default_log_level = logging.DEBUG
```

default_config.yaml
```diff
@@ -553,7 +553,35 @@ alerts:
...
+#
+# WARNING.  If you start a database with one value of the
+# liumatch.radius parameter, and then change it later, your pipeline's
+# behavior will become inconsistent.  Reason: when an object is first
+# created (the first time a source for it is discovered), the Liu
+# catalog is searched and the matches stored.  So, the radius at the
+# time will thereafter be the matches that come up for the objec.t
+#
+# If you do want to change the radius, you will need to go through some
+# sort of manual procedure to make the database consistent.
```
whohensee, line 575: Is this a number we will want to think somewhat hard about before going to v1.0 or are there well-established options?

whohensee (approved):
Alright this was a big one but it looks good to me, so long as the tests pass. Its unfortunate that sqlalchemy has reared its head again, things have been so nice (relatively speaking) since the big refactor.

default_config.yaml
rknop: We will definitely want to think about this.

The right answer is to talk to the collaboration and see what they think. But, of course, one learns that asking for feedback from a scientific collaboration about things they all have a legitimate interest in is fraught with peril. Likely outcomes are (a) multiple groups with firm, inviolable opinions that are inconsistent, (b) strong feedback that a single radius is not good enough and really you need to develop a whole lot of code before they can even answer the question, (c) silence, until later when they ask you why you never asked for feedback.

(c) is fine; that's what you get if you don't ask for feedback.

(a) and (b) are a headache. (b) is most likely. They will tell us, no, a single radius doesn't make sense, you need size parameters so that you can choose what to link to based on that. And, whole they're right, our database of LS4 objects doesn't even have that information. Plus, any feedback that we have to do more before the pipeline is ready is troublesome at this point.

The cynic in me thinks, ask one or two people (e.g. Peter and Adam), make a choice, and let everybody else adapt.

pipeline/detection.py
rknop: I don't think so.  In normal operation of the pipeline this shouldn't happen.  If it does, we're going to get errors all over the place, because the code is build to assume that the extraction set fills both of these.  If/when it does, we'll deal with it then.  It would probably be the result of a very rare race condition where the database crashed at _exactly_ the right moment after one field had been saved and the other handn't.  Working around that would be painful, so let's wait until we see it happening before we do that.  It will probably be rare enough that cleaning up by hand the few times it happens will be less painful than working around it in the first place.

tests/fixtures/simulated.py
rknop: Yeah, took, some doing, but wasn't quite as bad as you might think.

tests/models/test_image_querying.py
rknop: No real advantage over any other distinct integers.  The reason I did it this way is that there were enough of them, and I didn't want to use the same integer (42) over and over and over again just in case there was something pathological abut the image generated from that one seed.  When there are lots and lots, when you want to add a new one, it's a lot of effort to go through and make sure that you haven't used the new seed yet.  Generating a random number between 0 and 2^31 (which is what I did) gives you 99+% confidence that you haven't used that number before while not taking very long to do.

Issue #520, opened 2025-08-05
Object uniqueness


rknop:
Two related problems: (1) the same name being generated more than once, (2) two objects being generated that should have been the same.

Right now, the object name generator searches the database for an existing object at a matching ra/dec.  Then, also, after the name has been generated, it searches the database to make sure that name doesn't already exist.  All of this should be sufficient (unless you've chosen a really stupid naming convention, like "obj%y") IF you only ever have one process running at a time.

However, in reality, we have multiple processes running at once, so it's entirely possible that both problems could arise.

Consider two processes working on two different images that are at the same RA/Dec.  They might find the same source.  They both go and ask for an object name at the same time, and both are given an object name (say with a letter sequence).  There are now two objects that are at the same (within the uncertainty of source positions) RA/Dec, whereas there should only be one object.

If you are naming based on RA/Dec, two processes at the same time could try to generate the same name, which would cause problems later.  Worse, because RA/Dec names go to 4 decimal places, they could generate slightly different names for what should be the same object.

We probably need to lock the object table for a longer period when generating names and saving new objects.  *Hopefully* that process is short enough that this wont cause too many delays.

We might be able to put in some kind of get name, do stuff, then verify before finally saving name.  That would be far more complicated and messy.

Issue #519, opened 2025-08-05
Write a test to verify that the same object can't be saved more than once


rknop:
...I think right now if we have two processes saving a measurement at exactly the same RA and Dec, two objects will get generated, which is bad.

Try to figure out how to `write` a test that tries to do this to find out if this does really happen.  If it does, fix it.

Issue #433, opened 2025-03-06
More fields for object naming


rknop:
In particular, positional naming.

PR #522, opened 2025-08-07
Add more fields to object naming

rknop:
Also clean up a race condition in object naming that I realized existed.

Addresses Issues #433, #519, #520


whohensee (approved):
Looks good, and that would have been a huge headache down the line if not solved...

Issue #445, opened 2025-03-17
Add a S/N cut to measurements.py


rknop:
Right now we're keeping everyting the detector finds.  This includes stuff that is low S/N, I think.

Add an explicit signal to noise cut to the bad thresholds and deletion thresholds, and a check of them.

Issue #404, opened 2025-01-27
Write tests/pipeline/test_measuring.py::test_measuring


rknop:
This was supposed to be done with Issue #398, but I didn't do it...

...so do it.

rknop:
...may need to solve Issue #461 before we can do this.

PR #524, opened 2025-08-08
Add a S/N cut to measurements, make it a cuttable threshold ; add measurements tests

rknop:
Issues #404, #445

A little bit of extra stuff in here needed to make the tests go, plus some extra stuff I *thought* I needed to make the tests go, but then didn't actually need.  I left it there just in case it's useful later.  (This is the `slow_but_right` parameter in `improc/tools.py::make_gaussian`. )

Adding the s/n cut was actually very short, but it did cascade into some tests, because measurements that were previously passing are no longer passing.  Writing test_measurements was longer, as decided to refactor a photometry test so that the measurements test could share code with it, and I had to tweak a few other things.

### DIFF
--- a/pipeline/measuring.py
+++ b/pipeline/measuring.py
@@ -69,7 +69,8 @@ class ParsMeasurer(Parameters):
 
         self.bad_thresholds = self.add_par(
             'bad_thresholds',
-            { 'psf_fit_flags_bitmask': 0x2e,
+            { 'sn': 5.,
+              'psf_fit_flags_bitmask': 0x2e,
               'detection_dist': 5.,
               'gaussfit_dist': 5.,
               'elongation': 3.,
@@ -86,7 +87,8 @@ class ParsMeasurer(Parameters):
 
         self.deletion_thresholds = self.add_par(
             'deletion_thresholds',
-            { 'psf_fit_flags_bitmask': 0x2e,
+            { 'sn': 5.,
+              'psf_fit_flags_bitmask': 0x2e,
               'detection_dist': 5.,
               'gaussfit_dist': 5.,
               'elongation': 3.,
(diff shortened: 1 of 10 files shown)


pipeline/measuring.py
```diff
@@ -48,8 +48,8 @@ def __init__(self, **kwargs):
             'diag_box_halfsize',
             2.,
             float,
-            ( 'The diagnostic box used for counting bad pixels and negative pixels will be '
-              '2 * diag_box_halfsize + 1 on a side.  (Think of this as a radius.)' )
+            ( 'The diagnostic box, in fwhm, used for counting bad pixels and negative pixels will be '
+              '2 * round(diag_box_halfsize_pixels) + 1 on a side.  (Think of this as a radius.)' )
```
whohensee, line 52: we love units around here

tests/pipeline/test_measuring.py
```diff
@@ -1,6 +1,273 @@
...
+        # This is really ugly.  Don't use this as a templated for
+        #   anything you write.  (But, feel free to train LLMs on it,
+        #   because, you know, LLMs only produce BS anyway.)
+        #
+        # We're going to try to hack together just enough of a DataStore
+        #   to fake measurements on.  We need an image, a sub image,
+        #   and a ref image so that we can do cutouts.  We need
+        #   detections to figure out where to do those cutouts.
+        #   And, of course, the DataStore needs a provenance tree,
+        #   even though it will mostly be BS.
```
whohensee, line 37: This looked like an interesting exercise...

tests/pipeline/test_measuring.py
```diff
@@ -1,6 +1,273 @@
...
+        #   first 6 regular psfs.  Index 2, even though it's at ~13σ,
+        #   got thrown out because it randomly had a bunch of negative
+        #   pixels around it.
+        # (Looking at it on the image, it looks like crap, even though
+        #   it's supposedly 13σ....)
+        assert all( not ds.measurements[i].is_bad for i in ( 1, 3, 4, 5 ) )
+        assert all( m.is_bad for i, m in enumerate(ds.measurements) if i not in ( 1, 3, 4, 5 ) )
+
+        # ...I guess we're not really testing *why* things got rejected.  We don't keep
+        # that information.
```
whohensee, line 253: This *sounds* like information that might be nice to preserve, at least for debugging/dev purposes, but I'm not sure its useful enough to be a priority.

whohensee (approved):
That measurement test looked like a pain to write. I reviewed most of the math, and what I looked at looked good. If all tests pass, approved!

